PR #5, opened 2014-07-07
Add base files for the branch (.travis.yml, .coveragerc, .gitignore, README.md, LICENSE)

guewen:


guewen:
For the travis config file, I could find where are some dependencies, such as `base_external_referentials`.

bwrsandman:
@guewen found it here http://bazaar.launchpad.net/~openerp-commiter/openobject-addons/extra-6.0/files/head:/base_external_referentials/

guewen:
@bwrsandman thanks. I don't think that we should accept modules having dependencies that are not in OCA. On the other hand, the modules are already there...

bwrsandman:
We could adopt the dependancy module, I guess.

guewen:
I'm not sure I want to.

.coveragerc
```diff
@@ -0,0 +1,15 @@
+[report]
```
bwrsandman: With the current travis setup, we don't need to add coveragerc

.travis.yml
```diff
@@ -0,0 +1,20 @@
+language: python
```
bwrsandman, line 1: This needs to updated with the new quality tools samples

pedrobaeza:
@guewen, please fix commented issues to go on with this PR.

LeartS:
What's the status on this? Is it still needed? I see those files are present now..

guewen:
@LeartS it was for 6.1. Since the files have changed a lot I close this PR. Anyway the builds can't be green because most of the modules depends on not-oca addons.

PR #13, opened 2014-10-14
[ADD] add the module sale_automatic_workflow_cancel this module give the...

sebastienbeau:
... posibility to cancel automatically unconfirm order after some delay and depending of the payment method

### DIFF
--- a/sale_automatic_workflow_cancel/__openerp__.py
+++ b/sale_automatic_workflow_cancel/__openerp__.py
@@ -26,10 +26,10 @@
  'website': 'www.akretion.com',
  'license': 'AGPL-3',
  'category': 'Generic Modules',
- 'description': """Depending of the configuration of the payment
+ 'description': """Depending on the configuration of the payment
  method sale order will be automatically cancelled after some delay.
- For example you can configure that every sale order paid buy check will
- be cancelled after 90 days.
+ For example you can configure that every sale order paid by check will
+ be cancelled after 90 days if no payment is have been registry.
  """,
  'depends': [
      'sale_automatic_workflow',
--- a/sale_automatic_workflow_cancel/__openerp__.py
+++ b/sale_automatic_workflow_cancel/__openerp__.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+#   Module for OpenERP
+#   Copyright (C) 2014 Akretion (http://www.akretion.com).
+#   @author Sébastien BEAU <sebastien.beau@akretion.com>
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU Affero General Public License as
+#   published by the Free Software Foundation, either version 3 of the
+#   License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU Affero General Public License for more details.
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+{'name': 'Sale Automatic Workflow Cancel',
+ 'version': '0.0.1',
+ 'author': 'Akretion',
+ 'website': 'www.akretion.com',
+ 'license': 'AGPL-3',
+ 'category': 'Generic Modules',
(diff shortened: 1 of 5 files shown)


sale_automatic_workflow_cancel/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+#
+###############################################################################
+
+{'name': 'Sale Automatic Workflow Cancel',
+ 'version': '0.0.1',
+ 'author': 'Akretion',
+ 'website': 'www.akretion.com',
+ 'license': 'AGPL-3',
+ 'category': 'Generic Modules',
+ 'description': """Depending of the configuration of the payment
```
pedrobaeza: Depending s/of/on

sale_automatic_workflow_cancel/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+ method sale order will be automatically cancelled after some delay.
+ For example you can configure that every sale order paid buy check will
```
pedrobaeza: s/buy/by

sale_automatic_workflow_cancel/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+ be cancelled after 90 days.
```
pedrobaeza: You should complete it with: `...if no payment is done`

sale_automatic_workflow_cancel/payment.py
```diff
@@ -0,0 +1,77 @@
...
+        ids = self.search(cr, uid, domain, context=context)
+        for method in self.browse(cr, uid, ids, context=context):
+            sale_domain = self._get_sale_domain(
+                cr, uid, method, context=context)
+            order_ids = order_obj.search(
+                cr, uid, sale_domain, context=context)
+            _logger.debug('Cancel %s Sale Order for method %s'
+                          % (len(order_ids), method.name))
+            for order_id in order_ids:
+                with commit(cr):
```
pedrobaeza, line 73: Why do you need commit? Better to embrace in a `try:` for each order so that one errored order doesn't affect the others

guewen, line 73: I think it's fine to commit in this context:
- we are in a cron
- it will lock only 1 sale order at a time and for a short duration, not locking a long range and blocking users
- an ordered error is rollbacked and the process continues with the next
- if the cron is interrupted, all the already cancelled orders will already be committed.

sale_automatic_workflow_cancel/payment.py
```diff
@@ -0,0 +1,77 @@
...
+
+class PaymentMethod(orm.Model):
+    _inherit = 'payment.method'
+
+    _columns = {
+        'days_before_cancel': fields.integer(
+            'Days before cancel',
+            help="After 'n' days, if the 'Import Rule' is not fulfilled, the "
+                 "the sale order (or the import of the sale order) will "
+                 "be canceled."),
```
guewen, line 40: The help message seems unrelated to the actual implementation.

sale_automatic_workflow_cancel/payment.py
```diff
@@ -0,0 +1,77 @@
...
+        order_obj = self.pool['sale.order']
+        domain = self._get_method_domain(cr, uid, context=context)
+        ids = self.search(cr, uid, domain, context=context)
+        for method in self.browse(cr, uid, ids, context=context):
+            sale_domain = self._get_sale_domain(
+                cr, uid, method, context=context)
+            order_ids = order_obj.search(
+                cr, uid, sale_domain, context=context)
+            _logger.debug('Cancel %s Sale Order for method %s'
+                          % (len(order_ids), method.name))
```
guewen, line 71: Do not use string substitution in loggers.
The correct form is:

```
_logger.debug('Cancel %s Sale Order for method %s', len(order_ids), method.name)
```

sale_automatic_workflow_cancel/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+#   GNU Affero General Public License for more details.
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+{'name': 'Sale Automatic Workflow Cancel',
+ 'version': '0.0.1',
+ 'author': 'Akretion',
```
gurneyalex, line 25: can you add "Odoo Community Association (OCA)" ?

pedrobaeza:
@sebastienbeau, are you going to continue this work?

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

PR #14, opened 2014-10-14
7.0 sale automatic workflow reserve

sebastienbeau:


sebastienbeau:
Hi, as for now I do not need this feature anymore, I will not work more on it for now. If you need it you contribution is welcome

gurneyalex:
@legalsylvain are you planning to work on this? If not maybe we should close this PR.

PR #15, opened 2014-10-17
7.0 add warning field

rdeheele:
if the user passes a sale order from the manual to the automatic workflow
process, it can be useful to inform him that a cron could automatically confirm
the order in a close future.
it warns the user via an onchange pop-up box.


tafaRU:
:+1:

yvaucher:
:+1:

joaoalf:
:+1: 
lgtm

PR #17, opened 2014-10-22
[IMP] Add functionality to create picking invoices as per the sale workflow

craiggowing:
Allow picking invoices to be created when set in the workflow.
This functionality is missing when the invoice is from picking, so would have to be done manually in each case.


pedrobaeza:
I close this, as it seems not to be attended.

PR #12, opened 2014-09-15
Check if the lines can be reconciled before to reconcile them

guewen:
Prevents a lot of errors in the logs with reconciles that cannot be done.

Fixes #11

PR #18, opened 2014-11-10
6.1-Check account before reconciliation

florian-dacosta:
When we calculate the amount of the payments and of the invoice, we do not check the account, just the credit or debit amount.

I just check if the account of the invoice, match the account of the account.move.line before suming the amount.
This fix a bug to reconcile invoices containing discount/refund lines.
Example : I have an sale order with 2 lines : 
1) Computer => amount = 500
2) Refund/discount => amount = -100

The problem of this second line is that the invoice will generates an account move line with a debit = 100.

So when we sum the debit of the account move line of the invoice, we will have 400 +100 = 500
The payment linked to the sale order has an amount of 400.
It won't try to reconcile even if it should.

Maybe instead of checking if the invoice account match the account move line Account, I should just check if the account move line account is payable/receivable.

I think the bug is also present in v7.0, if confirmed I can propose a patch for the version 7

### DIFF
--- a/sale_automatic_workflow/sale.py
+++ b/sale_automatic_workflow/sale.py
@@ -100,15 +100,18 @@ class account_invoice(Model):
             payment_amount = 0
             invoice_amount = 0
             account_id = invoice.account_id.id
+            partner_id = invoice.partner_id.id
             if invoice.sale_ids and invoice.sale_ids[0].payment_id and invoice.move_id:
                 for move in invoice.sale_ids[0].payment_id.move_ids:
                     if move.credit > 0 and not move.reconcile_id \
-                            and move.account_id.id == account_id:
+                            and move.account_id.id == account_id \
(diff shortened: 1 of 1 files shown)


guewen:
Hi @florian-dacosta 
I already proposed this fix for 7.0: https://github.com/OCA/e-commerce/pull/12, it still awaits for reviews. 

You should also check if the partner is  the same.

guewen:
In fact, your fix is not exactly the same, you filter on the lines (I have been confused by the description "Check account before reconciliation") when I only validate that all the lines have the same account. Maybe both are complementary.

florian-dacosta:
You are right, I can also check is the partner is the same, I have updated the PR

I saw your PR, and indeed it seems it is complementary.

pedrobaeza:
:+1:

gurneyalex:
:+1:

PR #22, opened 2014-11-20
Remove product_links_sync

guewen:
The module has not been ported to 7.0 and above and depends on modules
that do no longer exist in recent versions.


pedrobaeza:
:+1:

sebastienbeau:
:+1:

PR #24, opened 2014-11-20
Migration of sale_automatic_workflow to 8.0

guewen:
# Migration of sale_automatic_workflow to 8.0
- Added some tests
- Removed the dependency on `sale_payment_method`: we should be able to use automatic workflows even if we don't use the sale payment methods. We will need to reimplement the few features I removed in a third "link" module between `sale_automatic_workflow` and `sale_payment_method`.
- Use new API whenever possible
- Updated `pot`


sale_automatic_workflow/sale_workflow_process.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    validate_invoice = fields.Boolean(string='Validate Invoice')
+    validate_picking = fields.Boolean(string='Confirm and Close Picking')
+    invoice_date_is_order_date = fields.Boolean(
+        string='Force Invoice Date',
+        help="When checked, the invoice date will be "
+             "the same than the order's date"
+    )
+    warning = fields.Text('Warning Message', translate=True,
+                          help='if set, display the message when a '
```
yvaucher: s/if/If

sale_automatic_workflow/stock_picking.py
```diff
@@ -0,0 +1,45 @@
...
+from openerp import models, fields, api
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    workflow_process_id = fields.Many2one(comodel_name='sale.workflow.process',
+                                          string='Sale Workflow Process')
+
+    def _create_invoice_from_picking(self, cr, uid, picking, vals,
```
yvaucher, line 31: Is it the only method not ported to api 8.0? Any reason for that?
An `@api.model` should work here to make the module fully ported.

sale_automatic_workflow/i18n/fr.po
```diff
@@ -1,80 +1,76 @@
...
+msgid "Deliver all products at once"
+msgstr "Livrer tous les articles en une fois"
 
 #. module: sale_automatic_workflow
-#: selection:sale.workflow.process,invoice_quantity:0
-msgid "Ordered Quantities"
-msgstr "Quantités commandées"
+#: selection:sale.workflow.process,picking_policy:0
+msgid "Deliver each product when available"
+msgstr "Livrer chaque article dès disponibilité"
```
yvaucher, line 73: s/disponibilité/qu'il est disponible

sale_automatic_workflow/stock_picking.py
guewen, line 31: Not sure if it is good or not to change it.
This method should become a `@api.multi` in the base module the day it is migrated because `picking` should be `self`:

```
@api.multi
def _create_invoice_from_picking(self, vals):
```

I am a bit scared to use a new-api signature that will anyway not be compatible with the base new api.

pedrobaeza, line 31: You can use @api.v7 and @api.v8 trick. There are some samples in core code.

guewen, line 31: > You can use @api.v7 and @api.v8 trick. There are some samples in core code.

I know @api.v7 and @api.v8, but IMO they should be used at the first declaration of the method, not in an `_inherit`. What if another module declare another signature?

pedrobaeza, line 31: You're right about inheritance. There will be two behaviours depending the path. For now, better to let it as is.

pedrobaeza:
:+1:

yvaucher:
:+1:

PR #30, opened 2015-01-19
[FIX] Only allow adding receivable entries to sale order payment

dudunato:
It should only be possible to link a payment/credit from a receivable account
to a sale order, so filter out all other account types

The accounting behavior doesn't make sense when it is showing all kind of entries. It should only show receivable entries.

To reproduce the bug:

1 - Go to "Quotation and Sales"
2 - Choose an order
3 - Go to "Automation Information" tab 
4 - Click the Edit button
5 - Click the Add payments button

After these steps there will be receivable entries in that table, but also entries from the other side of the payment transactions, for example, the bank entry

### DIFF
--- a/sale_payment_method/sale.py
+++ b/sale_payment_method/sale.py
@@ -61,11 +61,14 @@ class sale_order(orm.Model):
         return res
 
     _columns = {
-        'payment_ids': fields.many2many('account.move.line',
-                                        string='Payments Entries'),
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Seems reasonable :+1:

guewen:
:+1:

challs:
:+1: from me too.

I have been working with Dudu on this problem together and have verified I think it is correct from an accounting perspective.

PR #32, opened 2015-02-11
Do not propagate sale sale automatic workflow when several orders are grouped in one picking

guewen:
If we have several sales orders in one procurement group,
we group them in one picking. The possibilities when we have
several orders for one picking are:
1. take the workflow of the first sale order
2. do not propagate the workflow at all
3. propagate the workflow if all the workflows are the same
The solution 1. is bad, the 3. could be hard to understand and
unpredictable. The solution 2. takes no assumption on what is
expected.


pedrobaeza:
:+1: 

Equitative solution

guewen:
Can I have another review please?

OSguard:
:+1:

PR #34, opened 2015-02-13
Add sale_payment_method_automatic_workflow

guewen:
Previously, the module sale_automatic_workflow had a dependency on
sale_payment_method.  The dependency was there to automatically reconcile
the payments made by sale_payment_method.

Depends on
- [x] https://github.com/OCA/e-commerce/pull/35

PR #33, opened 2015-02-12
[WIP] migration of module sale_payment_method

fevxie:
migration sale_payment_method port 7.0 to 8.0

PR #35, opened 2015-02-13
migration of sale_payment_method

guewen:
Supersedes #33

Continues migration of `sale_payment_method`.
The things that were common to `sale_payment_method` and `sale_automatic_workflow` are now in https://github.com/OCA/e-commerce/pull/34


sale_payment_method/payment_method.py
```diff
@@ -0,0 +1,52 @@
...
+    def _default_company_id(self):
+        company_model = self.env['res.company']
+        return company_model._company_default_get('payment.method')
+
+    name = fields.Char(required=True,
+                       help="The name of the method on the backend")
+    journal_id = fields.Many2one(
+        comodel_name='account.journal',
+        string='Journal',
+        help="If a journal a selected, when a payment is recorded "
```
pedrobaeza: s/a selected/is selected/

sale_payment_method/sale.py
```diff
@@ -0,0 +1,246 @@
...
+    @api.depends('amount_total', 'payment_ids.credit', 'payment_ids.debit')
+    def _compute_amount(self):
+        paid_amount = sum(line.credit - line.debit
+                          for line in self.payment_ids)
+        self.amount_paid = paid_amount
+        self.residual = self.amount_total - paid_amount
+
+    @api.one
+    @api.depends('payment_ids')
+    def _compute_payment_exists(self):
```
pedrobaeza: This computed field can be omitted. It's direct to check for this condition in all places (code, views, domains...)

sale_payment_method/sale.py
```diff
@@ -0,0 +1,246 @@
...
+    def _prepare_payment_move(self, move_name, journal, period, date):
+        return {'name': move_name,
+                'journal_id': journal.id,
+                'date': date,
+                'ref': self.name,
+                'period_id': period.id,
+                }
+
+    @api.multi
+    def _prepare_payment_move_line(self, move_name, journal, period,
```
pedrobaeza, line 166: As this returns two dicts, better to put `_prepare_payment_move_lines`

sale_payment_method/payment_method.py
guewen: ✓

sale_payment_method/sale.py
guewen: No, it doesn't work everywhere.

If in the view I use this:

```
                <button string="Create Invoice" position="after">
                    <field name="payment_ids" invisible="1"/>
                    <button name="action_view_payments"
                        string="View Automatic Payment"
                        type="object" class="oe_highlight"
                        attrs="{'invisible': [('payment_ids', '=', False)]}"
                        groups="base.group_user"/>
                </button>
```

Or this:

```
                <button string="Create Invoice" position="after">
                    <field name="payment_ids" invisible="1"/>
                    <button name="action_view_payments"
                        string="View Automatic Payment"
                        type="object" class="oe_highlight"
                        attrs="{'invisible': [('payment_ids', '=', [])]}"
                        groups="base.group_user"/>
                </button>
```

The button is always displayed. This has not changed since v7.

sale_payment_method/sale.py
guewen, line 166: ✓

sale_payment_method/sale.py
pedrobaeza: I'm using the expression here: 

https://github.com/odoomrp/odoomrp-wip/blob/8.0/mrp_operations_extension/views/mrp_routing_view.xml#L37

and it works for me, but I'm seeing that the field is not of the same type (many2many vs one2many).

Have you tried:

```
attrs="{'invisible': [('payment_ids', '=', [(6, 0, []])]}"
```

or

```
attrs="{'invisible': [('payment_ids[0][2]', '=', [])]}"
```

in case payment_ids is expressed in that form?

guewen: With

```
attrs="{'invisible': [('payment_ids', '=', [(6, 0, []])]}"
```

it always show the button

With 

```
attrs="{'invisible': [('payment_ids[0][2]', '=', [])]}"
```

I have this error:

```
Error: Unknown field payment_ids[0][2] in domain [["payment_ids[0][2]","=",[]]]
```

gurneyalex:
code looks ok. Having automated tests would be nice, but this will get exercised by the magento connector tests presumably. 

:+1:

sale_payment_method/sale.py
pedrobaeza: I think this is why @hbrunn creates this module: https://github.com/OCA/web/blob/7.0/web_compute_domain_x2many/__openerp__.py

guewen: Ok, thanks, that's good to know. I don't want to add a dependency (especially on another repo) for that since the function fields does it well.

pedrobaeza: Yeah, of course, it's too much to add a module for only this question.

hbrunn: Use `[[6, False, []]]` (it checks for equality in js code)

guewen: > Use [[6, False, []]](it checks for equality in js code)

Oh that's it! It works now. Thanks a lot @hbrunn

pedrobaeza:
@guewen, please rename the method mentioned above and I'll merge this to continue with the migration plan.

guewen:
> @guewen, please rename the method mentioned above and I'll merge this to continue with the migration plan.

That's done already: https://github.com/guewen/e-commerce/commit/dfcb129fcee0fbdef6ae40e6065d1a717017977e

pedrobaeza:
I didn't see the deprecation trick.

I merge then.

PR #37, opened 2015-03-03
Add check on invoice account

mdietrichc2c:
This PR is to ensure only move lines with the same account as the invoice are considered for the automatic validation; otherwise, any discount line on the invoice will also be used, and lead to the invoice not using the sale order payment.


yvaucher:
:+1:

PR #39, opened 2015-03-14
[FIX] change called method to confirm sale orders

bguillot:
Hi, 

I am maybe missing something but I think we can change the method called to validate the sale orders in the automatic workflow job. 
Indeed in 8.0 there is now a new method action_button_confirm() called by the button, instead of the workflow action in previous version. 
This method makes easier to override the order confirmation so if someone uses it, instead of action_wait() (like it was usually the case in previous versions), the override is skipt by the automatic workflow job.

Maybe it has been made in purpose and my PR is useless.

Thanks


pedrobaeza:
I fully agree: it's more inheritable this way.

:+1:

rdeheele:
:+1:

guewen:
:+1:

bguillot:
Thanks !

PR #42, opened 2015-04-10
Specific context to get period (avoid special periods + correct company)

mdietrichc2c:
This PR is to add 2 context variables, 'company_id' and 'account_period_prefer_normal', to retrieve the period for the automatic payment.
Without those, a payment set to January 1st can end in the Opening Entries period instead of Jan.


pedrobaeza:
:+1:

yvaucher:
:+1:

vrenaville:
:+1:

PR #43, opened 2015-04-16
[ADD] website_sale_vat_required

eLBati:
This module extends checkout e-commerce form in order to force user to fill 'VAT number' field


pedrobaeza:
This should be on website project, doesn't it?

eLBati:
@pedrobaeza Isn't website project for CMS modules mainly?

pedrobaeza:
No, website it's for modules that modify front-end things, like this one. You have also selected _website_sale_ prefix for this reason. Please move it, and take the ocassion to select a more specific icon than the cart icon (putting an overlayed VAT word for example).

eLBati:
Ok @pedrobaeza , done with https://github.com/OCA/website/pull/24

PR #45, opened 2015-06-01
8.0 website sale osc

Benniphx:
# odoo One Step Checkout

One Step Checkout combines all Odoo Checkout steps into one and removes all unnecessary fields and
questions. Never before has check- out been easier and faster!
Improving the checkout process results in more customers completing their sales, and this has an
immediate impact on your bottom line. It is the single most effective technical change you can make
to reduce shopping cart abandonment.


pedrobaeza:
I think this should be in OCA/website, not here.

Benniphx:
@pedrobaeza : @gurneyalex said it belonged here

pedrobaeza:
The limit is a bit undefined, but I prefer website because it modifies controllers and so on, and thus all these modules are in there, like website_sale_order_company, website_sale_unsaleable_options, and so on.

@gurneyalex, can you give your opinion with this new information?

jgrandguillaume:
Hi there,

Sorry for that, we've got a discussion over here and conclude that this PR deserve to be moved on this one: https://github.com/OCA/website

The reasons are:
- website repo already have website_\* module like website_sale_XY so it make more sense.

I've updated the description of each repo accoridngly. Please move it on the proper repo.

Thanks in advance and sorry for the wontrg info,

Joël

Benniphx:
https://github.com/OCA/website/pull/36

PR #46, opened 2015-06-03
Migration of sale_quick_payment to 8.0

leio:


sale_quick_payment/__openerp__.py
```diff
@@ -22,7 +22,7 @@
 
 {
     'name': 'Sale Quick Payment',
-    'version': '0.2',
+    'version': '0.3',
     'category': 'Generic Modules/Others',
     'license': 'AGPL-3',
     'description': """
```
pedrobaeza, line 28: Please extract description to a README.rst file, and use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

pedrobaeza: Please extract description to a README.rst file, and use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

pedrobaeza: Please extract description to a README.rst file, and use OCA's template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

sale_quick_payment/wizard/pay_sale_order.py
```diff
@@ -0,0 +1,74 @@
...
+    @api.multi
+    def pay_sale_order(self):
+        """ Pay the sale order """
+        self.ensure_one()
+        order = self.env['sale.order'].browse(self.env.context['active_id'])
+        order.add_payment(self.journal_id.id,
+                          self.amount,
+                          date=self.date,
+                          description=self.description)
+        return {'type': 'ir.actions.act_window_close'}
```
pedrobaeza: Now you don't need to return this, but only True to close the window.

sale_quick_payment/__openerp__.py
```diff
@@ -51,5 +51,5 @@
              'settings/sale.exception.csv',
              ],
     'demo': [],
```
lmignon: Empty entry can be removed

lepistone:
This has been resubmitted as #54, and I'll resubmit again in `OCA/sale-workflow`. Thanks!

PR #48, opened 2015-07-08
[ADD] website_sale_without_duplicated_partners

JavierIniesta:
This module was written to avoid duplicated partners when they buy online. Based on partner email.


JavierIniesta:
Addon is badly thought out.

PR #41, opened 2015-03-25
ShopByBrand Feature in Odoo!

JayVora-SerpentCS:
Added the ShopByBrand Feature in Odoo!

PR #51, opened 2015-08-21
New module: website_sale_cart_preview

LeartS:
Shows a preview of your cart on mouseover.
Extracted and adapted from the equivalent version 9 feature.

### DIFF
--- a/website_sale_cart_preview/__openerp__.py
+++ b/website_sale_cart_preview/__openerp__.py
@@ -26,7 +26,7 @@
     'website': "http://www.wearemonk.com",
     'category': 'Website',
     'version': '8.0.1.0.0',
-    'license': 'AGPL-3',
+    'license': 'GPL-3',
     'depends': ['website_sale'],
     'data': [
         'templates/website_sale_cart_preview.xml',
--- a/website_sale_cart_preview/__openerp__.py
+++ b/website_sale_cart_preview/__openerp__.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#     Copyright (c) Leonardo Donelli @ MONK Software http://www.wearemonk.com
+#
+#     This is free software: you can redistribute it and/or
+#     modify it under the terms of the GNU Affero General Public License
+#     as published by the Free Software Foundation, either version 3 of
+#     the License, or (at your option) any later version.
+#
+#     This is distributed in the hope that it will be useful,
+#     but WITHOUT ANY WARRANTY; without even the implied warranty of
+#     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#     GNU Affero General Public License for more details.
+#
+#     You should have received a copy of the
+#     GNU Affero General Public License
+#     along with web_expand_dialog.
+#     If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Cart Preview',
+    'summary': 'Preview your cart on mouseover',
+    'author': "MONK Software,Odoo Community Association (OCA)",
+    'website': "http://www.wearemonk.com",
+    'category': 'Website',
+    'version': '8.0.1.0.0',
+    'license': 'AGPL-3',
+    'depends': ['website_sale'],
+    'data': [
+        'templates/website_sale_cart_preview.xml',
+        'templates/website_sale.xml',
+        'templates/website.xml'
+    ],
+}
--- a/website_sale_cart_preview/controllers/main.py
+++ b/website_sale_cart_preview/controllers/main.py
@@ -6,35 +6,15 @@ from openerp.http import request
 import openerp.addons.website_sale.controllers.main
 
 
-class website_sale(openerp.addons.website_sale.controllers.main.website_sale):
+class WebsiteSale(openerp.addons.website_sale.controllers.main.website_sale):
 
     @http.route(['/shop/cart'], type='http', auth="public", website=True)
     def cart(self, **post):
-        cr, uid, context = request.cr, request.uid, request.context
-        pool = request.registry
-        order = request.website.sale_get_order()
-        if order:
-            from_currency = pool.get('product.price.type')._get_field_currency(
-                cr, uid, 'list_price', context)
-            to_currency = order.pricelist_id.currency_id
-            compute_currency = lambda price: pool['res.currency']._compute(
-                cr, uid, from_currency, to_currency, price, context=context)
-        else:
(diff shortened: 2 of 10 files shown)


pedrobaeza:
If this is from Odoo, you should extract commit history with git filter-branch and then apply your changes. They need also to be credited.

LeartS:
Didn't we already discuss this in https://github.com/OCA/website/pull/53 ?

pedrobaeza:
Sorry, I didn't remember the casuistic. It still lacks the credits to Odoo S.A.

website_sale_cart_preview/controllers/main.py
```diff
@@ -0,0 +1,40 @@
...
+            'compute_currency': compute_currency,
+            'suggested_products': [],
+        }
+        if order:
+            _order = order
+            if not context.get('pricelist'):
+                _order = order.with_context(pricelist=order.pricelist_id.id)
+            values['suggested_products'] = _order._cart_accessories()
+
+        if post.get('type') == 'popover':
```
oscarolar, line 14: You are only adding 3 lines of code and you are breaking inheritance mechanism for the method `cart` on website_sale, please use `super` to inherit correctly. 
a example has been posted on [this commit](https://github.com/JayVora-SerpentCS/e-commerce/commit/5067190ce99befcb0272c48cca1add3c945b1f3b), it was the same case, it can be solved the same way.

website_sale_cart_preview/static/src/css/website_sale_cart_preview.css
```diff
@@ -0,0 +1,26 @@
+.mycart-popover .cart_line {
```
oscarolar, line 1: why didn't you used .less instead of css?, this can break namespaces on general css.

website_sale_cart_preview/controllers/main.py
LeartS, line 14: The ironic thing is that [I was the one that showed the way to do it](https://github.com/OCA/e-commerce/pull/41#issuecomment-132981198) to the author on that PR! :laughing:
I don't remember why I didn't do the same thing here, if I forgot or there was a reason, I'll give it a look tomorrow.

website_sale_cart_preview/static/src/js/website_sale_cart_preview.js
```diff
@@ -0,0 +1,45 @@
+$(document).ready(function() {
```
oscarolar, line 1: using multiple $(document).ready() (Odoo already uses it in several places) it is not always the neatest thing to do.
maybe you can use animationRegistry to add your JS code limited only to the cart DOM selector, an example here https://github.com/odoo/odoo/blob/8.0/addons/mass_mailing/static/src/js/website_mass_mailing.js#L6

website_sale_cart_preview/controllers/main.py
oscarolar, line 14: Hahaha, shit happens. :smile:

LeartS, line 14: So, I looked into it: I couldn't use that method because I needed the `values` the original implementation creates, but I see that I can extract them from the rendered response the super method returns, so I did that. Unfortunately this means that when type = 'popover', 2 templates get rendered. But I don't think it's a big deal.

website_sale_cart_preview/static/src/css/website_sale_cart_preview.css
LeartS, line 1: Because I never wrote .less and I never saw an OCA module using it. But with an example I can do it.

website_sale_cart_preview/static/src/js/website_sale_cart_preview.js
LeartS, line 1: I couldn't find any documentation whatsoever on this stuff, but I can try. The source seems quite simple.

website_sale_cart_preview/static/src/css/website_sale_cart_preview.css
oscarolar, line 1: About this, I was looking at the OCA builds on runbot and you are right, it is necessary to install website_less module to compile .less files, and is not available, so nevermind.

pedrobaeza:
:+1:

website_sale_cart_preview/static/src/js/website_sale_cart_preview.js
LeartS, line 1: @oscarolar that seems to be specific for snippets, for example (excluding the source code file name) the `start` method accepts and `editable_mode` param. It probably works, but are we sure it's the correct way to do this? Do you know if there are any more general-purpose registries that could be used for this?

LeartS:
Rebased on current 8.0

rafaelbn:
Tested in runbot :+1: 
Super thanks!
@dgzurita you should take a look in this module

carlosdauden:
:+1:

pedrobaeza:
Please move it to e-commerce and I'll merge it at the moment.

LeartS:
@pedrobaeza what do you mean? We are in e-commerce repo.

pedrobaeza:
Sorry, I have mixed PRs, hehe. My bad.

Let's see if @oscarolar say something about your last comment, and I'll merge.

rafaelbn:
@oscarolar Please we are waiting for you comments to merge, PR has already 3 +1. Anything to say? Thanks

website_sale_cart_preview/static/src/js/website_sale_cart_preview.js
oscarolar, line 1: There is no doc to support this way of doing this feature, nevertheless its the most efficent way to load this kind of JS files is after browser finishes retrieveing all base JS  and not abuse over the `.ready();` method, this does not mean that you are wrong or I am right, just is not a good practice, its more about a shape rule, anyway thanks for the great contribution, lets merge it.

oscarolar:
you can merge this now :+1:

website_sale_cart_preview/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+#
+##############################################################################
+{
+    'name': 'Cart Preview',
+    'summary': 'Preview your cart on mouseover',
+    'author': "MONK Software,Odoo Community Association (OCA)",
+    'website': "http://www.wearemonk.com",
+    'category': 'Website',
+    'version': '8.0.1.0.0',
+    'license': 'AGPL-3',
```
pedrobaeza: Last comment. I think this GPL-3, not AGPL-3

LeartS:
@pedrobaeza changed the license, sorry for the elapsed time.

PR #52, opened 2015-08-21
[FIX] Validate pikcing was broken during the port to the new version

sebastienbeau:
Hi during the port to the new version the workflow method was not pass correctly to the picking.
Indeed even if a method was set on the sale order no method was set on picking.

### DIFF
--- a/sale_automatic_workflow/__init__.py
+++ b/sale_automatic_workflow/__init__.py
@@ -27,3 +27,4 @@ from . import automatic_workflow_job
 from . import invoice
 from . import stock_picking
 from . import stock_move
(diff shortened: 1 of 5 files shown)


pedrobaeza:
You don't need to propagate the data to the procurement group. You can get it this way:

```
move.procurement_id.sale_line_id.order_id.workflow_process_id
```

gurneyalex:
:+1:

pedrobaeza:
Please take into account my remark to avoid a new field.

sebastienbeau:
This have been fixed in sale_workflow

Issue #47, opened 2015-06-09
Move sale_automatic_workflow and sale_payment_method to sale-workflow project


bguillot:
Hello, 

I am here to discuss about moving the modules sale_automatic_workflow (and derived) and sale_payment_method (and derived) to the project sale-workflow. 

First of all those modules are very linked with other sale-workflow modules, it leads to loop dependencies for travis for instance (modules in e-commerce depends on modules in sale-workflow). 
That's why @gurneyalex and @sebastienbeau suggested to move at least sale_payment_method in my PR https://github.com/OCA/sale-workflow/pull/121

Moreover, thoses modules can be used even in a non e-commerce company (you may want to automate the validation of sale_orders or invoices without having an e-commerce shop).

So after discussing with @guewen it appears that maybe sale_automatic_workflow, sale_payment_method and sale_payment_method_automatic_workflow would be better in sale-workflow project (product_links is still relevant in e-commerce).

That makes also sense for the unpported modules : sale_quick_payment, sale_automatic_workflow_exceptions and sale_payment_method_transaction_id (thoses are glues modules).

Before extracting the branches I would like to have your opinions about that and what will be the best procedure to do it (indeed thoses modules are concerned by some PR's).

I can first do it with sale_payment_method (no PR in 7 or 8).

Thanks for your feedbacks

gurneyalex:
I'm :+1: on this in principle, but let's not rush things. I'll raise attention on the contributors mailing list.

max3903:
:+1:

florian-dacosta:
:+1:

bealdav:
:+1:

petrus-v:
:+1: 

I wonder if it's possible to let somehow a message to let users know where to find the new repo host

pedrobaeza:
I also think that e-commerce repository overlaps a lot of others repositories, and more over, e-commerce is only one channel more, not an isolated flow itself, so I prefer to remove this repo and split its contents in the corresponding areas repos.

nbessi:
I'm :+1:  on the concept but I do not want to break any existing setup. One of the role of OCA is to provide some stability. We were angry enought with Odoo coboy release let's try to do better.

jgrandguillaume:
:+1:  to move it

leio:
[...]

PR #54, opened 2015-09-03
8.0 sale quick payment

gurneyalex:
port sale_quick_payment to v8


gurneyalex:
updated version of #46

guewen:
Since there is a button to display the payments (https://github.com/gurneyalex/e-commerce/blob/8.0-sale_quick_payment/sale_payment_method/sale_view.xml#L19-L23) I think we could remove the `payment_ids` field from the sales orders' view.

bguillot:
Thanks this module is needed ! :+1:

guewen:
:+1:

rafaelbn:
Hi @gurneyalex:
- In order yo have same place for everything, don't you think is better yo have a tab in sales order for payments as in invoices?

After trying to test this module I have doubts sorry... 
- this is a module for e-commerce and this module install sale_payment_method
- In e-commerce usually payment are done throug Paypal, Ogone, Stripe, Credit Card, ...
- This module sale_quick_payment I guess helps to sale online and receive the payment in the sale order

Why is not related with how the customer paid (payment.acquirer)?

When customer buy online you would like to have their purchases "quick paid" when the payment is done through the payment.acquirer. How you make this with this module?

Thanks

guewen:
> this is a module for e-commerce and this module install sale_payment_method
>     In e-commerce usually payment are done throug Paypal, Ogone, Stripe, Credit Card, ...
>     This module sale_quick_payment I guess helps to sale online and receive the payment in the sale order

Hi @rafaelbn,

This set of modules (`sale_payment_method`, `sale_automatic_workflow`, ...) has never been thought in the context of the Odoo e-commerce modules (`website_sale`, ...).
These modules are used by the Magento, Prestashop (etc.) connectors when you don't necessary have the `website` modules installed.
Possibly an integration with the `payment.acquirer` would be possible, but I'm not sure about that. Anyway I don't think it could happen in 8.0 because the `sale_payment_method` module is already widely used.

guewen:
@rafaelbn I think this project brings confusion because it mixes addons related to odoo's e-commerce and to e-commerce addons unrelated to it and both not always share the same path.

pedrobaeza:
That's why we talked about moving this kind of modules to sale-workflow.

guewen:
> That's why we talked about moving this kind of modules to sale-workflow.

totally!

gurneyalex:
So this would be
- sale_quick_payment
- sale_sale_payment_method

Do we want to also move 
- sale_automatic_workflow
- sale_payment_method_automatic_workflow ?

pedrobaeza:
Yeah, I think so, please see https://github.com/OCA/e-commerce/issues/47

LeartS:
I agree. e-commerce is going to be a big part of Odoo with a lot of modules, we should try to keep it clean, at least for version 9.

rafaelbn:
IMO we must move this four to sale-workflow. I agree with moving:
- sale_quick_payment
- sale_sale_payment_method
- sale_automatic_workflow
- sale_payment_method_automatic_workflow

Nothing has to be with e-commerce (website_sale - that why the name was [discussed here](https://odoo-community.org/groups/oca-contributors-15/oca-contributors-24218)).

In fact is more related with connector than Odoo's e-commerce.

About this theme, we are working in several e-commerce with Odoo, and we need to have different workflows depending if the sale is from backend or from e-commerce (obviously) and this means that we will work in (near future):
- website_sale_quick_payment
- website_sale_sale_payment_method
- website_sale_automatic_workflow
- website_sale_payment_method_automatic_workflow

gurneyalex:
I'll change this over the week end.

rafaelbn:
Hello, please when possible change this PR to https://github.com/OCA/sale-workflow as discussed.
Also discussed here: https://github.com/OCA/e-commerce/issues/47
Thanks

gurneyalex:
Hi all,

I'm totally under the water, and need help on these PRs. I'd be very grateful if someone could pick them up and do the migration. @lepistone maybe you could lend me a hand here?

:heart_eyes_cat:

lepistone:
ok, I'm picking this up.

lepistone:
Superseded by #86 and OCA/sale-workflow#239.

PR #55, opened 2015-09-03
add autopay option on sale automatic workflow

gurneyalex:


sale_payment_method_automatic_workflow/model/automatic_workflow_job.py
```diff
@@ -37,6 +37,21 @@ def _reconcile_invoices(self):
...
+                ('workflow_process_id.autopay', '=', True)]
+
+    @api.model
+    def _autopay(self):
+        sale_obj = self.env['sale.order']
+        sales = sale_obj.search(self._get_domain_for_sale_autopayment())
+        _logger.debug('Sale Orders to automatically pay: %s', sales)
+        for sale in sales:
+            with commit(self.env.cr):
+                sale.automatic_payment()
```
pedrobaeza, line 61: Are you relaying in any other module for this function?

gurneyalex, line 61: it is defined in sale_payment_method

guewen:
Sounds good :+1:

gurneyalex:
added some tests too.

moylop260:
:+1:

bguillot:
Looks good :+1:

sale_payment_method_automatic_workflow/model/automatic_workflow_job.py
```diff
@@ -37,6 +41,21 @@ def _reconcile_invoices(self):
...
     @api.model
     def run(self):
+        self._autopay()
         res = super(AutomaticWorkflowJob, self).run()
         self._reconcile_invoices()
         return res
+
+    @api.model
+    def _get_domain_for_sale_autopayment(self):
+        return [('state', '=', 'draft'),
```
lepistone, line 51: It is intentional that this will not get `open` orders? Do you think this could interfere with the workflow that confirms orders in `sale_automatic_workflow`?

lepistone:
Neat! This is significant enough to bump the module version.

rafaelbn:
Hi @gurneyalex do you think you could make a PR moving this module to https://github.com/OCA/sale-workflow as discussed here https://github.com/OCA/e-commerce/issues/47
Thanks again

lepistone:
Ok, I moved this to OCA/sale-workflow#243. Please continue reviewing there. Thanks!

PR #56, opened 2015-09-09
Fix product_links when creating a product

florian-dacosta:
Hello,
If you create a product and at the same time you want to creates the product links (by the one2many field), it won't be possible.
The reason is that on the product.links form view, the field 'product_id' will be invisible, required and empty.
The product_id field should not be in the form view, even invisible when we  create it from the one2many fields on the product.

It is easy to reproduce. Install this module, create a product and try to add the links before saving your new product.

@fevxie @guewen  the changes are ok for you?

### DIFF
--- a/product_links/product_links.py
+++ b/product_links/product_links.py
@@ -36,8 +36,7 @@ class ProductLink(models.Model):
         comodel_name='product.product',
         string='Source Product',
         required=True,
-        ondelete='cascade',
-        default=lambda self: self.env.context.get('product_id', False))
+        ondelete='cascade')
     linked_product_id = fields.Many2one(
         comodel_name='product.product',
         string='Linked product',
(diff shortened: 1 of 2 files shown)


guewen:
When the view is opened from the menu, the field should be there, so I guess we should have 2 different views; one for the menu and one for the product's view.

florian-dacosta:
Yes, this is exactly what I did in this PR.
The view for the product is directly defined inside the "product_link_ids" field.

guewen:
hmm yes, sorry ;-)
:+1:

bguillot:
I prefer the way too :+1:

oscarolar:
:+1:

yvaucher:
Thanks

merged at https://github.com/OCA/e-commerce/commit/0cf23c0a08cd89ee5aea5e6c4dc107da0317c1c4

PR #65, opened 2015-10-18
add missing inheritation of payment.method

Karsten-77:


pedrobaeza:
As said in the issue, from which module do you get this inheritance?

pedrobaeza:
As stated on the issue, this doesn't apply, so I close.

PR #66, opened 2015-10-19
Add module website_sale_suggest_create_account.

yajo:
This module was written to extend the functionality of e-Commerce to suggest the buyer to create an account, but without forcing him to do it.

CC @rafaelbn.


website_sale_suggest_create_account/README.rst
```diff
@@ -0,0 +1,77 @@
...
+
+However, the *Checkout without sign in* button is not highlighted, to
+implicitly encourage users to create the account.
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+* *Settings > Configuration > General Settings* and enable *Allow external
```
oscarolar, line 23: Can you set this as a default value so no configuration is needed? (.yml or .xml files)

yajo, line 23: Allowing anybody to sign into your Odoo implies some security considerations, so I think this step should be made by hand. Anyway this parameter has no XMLID to inherit from, so does not seem possible to do it with data files.

carlosdauden:
:+1:

rafaelbn:
Tested in runbot :+1:

dgzurita:
Tested in runbot :+1:

PR #70, opened 2015-11-02
[ADD] new module website_multi_image 

oscarolar:
- This module adds multiple product images into a tab in product.template in a tab called Product Images, and allows an over-ride of product.image to allow for a multi-image carousel for the product view page.


pedrobaeza:
This mixes 2 modules (one for having multiple images per product and another for viewing in the website) in the same module. You should take https://github.com/OCA/product-attribute/pull/57 as the base for the first one (and the refactoring that is going to be done).

PR #80, opened 2015-11-24
sale_automatic_workflow: ship only available goods

lepistone:
This is a new option that will ship what is available, and create a
backorder or the rest, like when we check availability and transfer from
the picking. The default is still as before: all the quantity is shipped
even if it is not available.

### DIFF
--- a/sale_automatic_workflow/sale_workflow_process.py
+++ b/sale_automatic_workflow/sale_workflow_process.py
@@ -68,6 +68,11 @@ class SaleWorkflowProcess(models.Model):
     )
     validate_invoice = fields.Boolean(string='Validate Invoice')
     validate_picking = fields.Boolean(string='Confirm and Close Picking')
+    ship_only_available = fields.Boolean(
+        string='Ship only available goods',
+        help="When checked, the available quantity is shipped, and if a "
+        "backorder is created with the rest"
+    )
     invoice_date_is_order_date = fields.Boolean(
         string='Force Invoice Date',
         help="When checked, the invoice date will be "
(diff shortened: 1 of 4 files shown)


sale_automatic_workflow/sale_workflow_process.py
```diff
@@ -68,6 +68,11 @@ class SaleWorkflowProcess(models.Model):
     )
     validate_invoice = fields.Boolean(string='Validate Invoice')
     validate_picking = fields.Boolean(string='Confirm and Close Picking')
+    ship_only_available = fields.Boolean(
+        string='Ship only available goods',
+        help="When checked, the available quantity is shipped, and if a "
```
elicoidal, line 73: s/and if a /and if so, a

sale_automatic_workflow/sale_workflow_process.py
```diff
@@ -68,6 +68,11 @@ class SaleWorkflowProcess(models.Model):
...
+        "backorder is created with the rest"
```
elicoidal, line 74: period

sale_automatic_workflow/sale_workflow_process_view.xml
```diff
@@ -26,6 +26,7 @@
                     <group string="Workflow Options">
                         <field name="validate_order"/>
                         <field name="validate_picking"/>
+                        <field name="ship_only_available"/>
```
sebastienbeau, line 29: If I am right, we must have the option "validate_picking" in order to enjoy the option "ship_only_available", if it's the case, then may we can hide the option "ship_only_available" and show it only it "validate_picking" is tic, what do you think?

lepistone:
Superseded by OCA/sale-workflow#244. Thanks!

PR #91, opened 2015-12-30
sale_payment_method: set add_payment multicompany aware

laetitia-gangloff:
The use case is the following :

company B parent of company A

Create quotation in company A
Set user in company B
Add payment to the created quotation

result : there is the following error :
2015-12-30 16:20:54,469 24665 TEST del_test2 openerp.modules.module: ` openerp.osv.orm.except_orm: ('ValidateError', u'Error occurred while validating the field(s) company_id: Account and Period must belong to the same company.')

expected : no error !

### DIFF
--- a/sale_payment_method/sale.py
+++ b/sale_payment_method/sale.py
@@ -210,6 +210,7 @@ class sale_order(orm.Model):
                 'date': date,
                 'ref': sale.name,
                 'period_id': period.id,
+                'company_id': sale.company_id.id,
                 }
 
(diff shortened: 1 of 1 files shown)


sbidoul:
@laetitia-gangloff is it not possible to do the same thing by rebrowsing and accessing the property with force_company in the context?

laetitia-gangloff:
Oh right, the problem was only the company not correctly set on created move/lines.
Thank you !

PR #96, opened 2016-01-13
sale_payment_method : add force_cancel

laetitia-gangloff:
to be able to cancel order with payment.

My use case is the following :
- customer do an order an pay it
- later, customer retract
- I want to cancel order
- I don't want to do an invoice/refund
- I want to generate refund move when I cancel the order

So I wan to be able to force cancellation of an order, in this case.

### DIFF
--- a/sale_payment_method/sale.py
+++ b/sale_payment_method/sale.py
@@ -305,6 +305,11 @@ class sale_order(orm.Model):
         return action
 
     def action_cancel(self, cr, uid, ids, context=None):
+        if context is None:
+            context = {}
+        if context.get('force_cancel', False):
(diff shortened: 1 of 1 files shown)


sale_payment_method/sale.py
```diff
@@ -305,6 +305,11 @@ def action_view_payments(self, cr, uid, ids, context=None):
         return action
 
     def action_cancel(self, cr, uid, ids, context=None):
+        if context is None:
+            context = {}
+        if context.get('force_cancel', False):
```
guewen, line 310: `, False` is useless here as the return value of `{}.get()` is `None` when the key does not exist.

guewen:
:+1:

lmignon:
:+1:

PR #102, opened 2016-03-29
[8.0] website_sale_require_legal translation

paolovalier:
added pot translation file and italian translation


pedrobaeza:
The pot is no longer needed and should be removed, and the translations are done in Transifex (https://www.transifex.com/oca/OCA-website-8-0/dashboard/). Why don't you use it?

paolovalier:
@pedrobaeza well, simply I can't find it in Transifex so I thought it was the pot file missing...

pedrobaeza:
OK, I confused the project and point to website in Transifex. I have added the e-commerce project to Transifex, so please make the translation there: https://www.transifex.com/oca/OCA-e-commerce-8-0/

Issue #114, opened 2016-07-05
Transifex creates i18n folder for non-existing module in 8.0 branch


atchuthan:
OCA Transbot created i18n folder for product_multi_link module which is available for 9.0 version only.
https://github.com/OCA/e-commerce/tree/8.0/product_multi_link/i18n
This issue seems very weird to me.

pedrobaeza:
This is because in some moment in time, this module belongs to this repository, and the resource in Transifex was created. Check this issue with the explanation: https://github.com/OCA/maintainer-quality-tools/issues/265

PR #115, opened 2016-07-05
[REM] empty module "product_multi_link" with i18n folder

atchuthan:
Related issue: https://github.com/OCA/e-commerce/issues/114


pedrobaeza:
This is because there are translations in Transifex about this module. I have removed them, and I fast-track this one to fix the problem. Thanks for the catch-up.

PR #117, opened 2016-07-22
[IMP][website_sale_require_legal] Reuse template

yajo:
Nowadays, `website_legal_page` includes a reusable template that makes this
module more... modular.

You should notice no meaningful difference in UI.

@Tecnativa.


pedrobaeza:
:+1:

rafaelbn:
Thanks :+1:

PR #118, opened 2016-07-27
[ADD]Adding module website_sale_by_supplier

mikevhe18:
This module is an additional feature of the module "website supplier_list" (repository: website). This module has a function to display products related to suppliers in the page "Our Suppliers"


andhit-r:
I tried:
1. Login as admin
2. Switch to website
3. Click **Our Suppliers**
4. Click one of the supplier

Exception raised:
selection_1

andhit-r:
@mikevhe18 now it is 404 (Page not found)

mFlayyan:
tested on run bot , working

website_sale_by_supplier/controllers/main.py
```diff
@@ -0,0 +1,66 @@
...
+class WebsiteSale(website_sale):
+    @http.route(['/shop',
+                 '/shop/page/<int:page>',
+                 '/shop/category/<model("product.public.category"):category>',
+                 """/shop/category/<model("product.public.category"):category>
+                 /page/<int:page>""",
+                 '/shop/suppliers/<model("res.partner"):product_supplier>'],
+                type='http',
+                auth='public',
+                website=True)
```
mFlayyan: since type, auth and website values are still the same , you can just skip them

website_sale_by_supplier/controllers/main.py
```diff
@@ -0,0 +1,66 @@
...
+        product_supplier=None,
+        search='',
+        **post
+    ):
+        if product_supplier:
+            request.context.setdefault('supplier_id', int(product_supplier))
+        result = super(WebsiteSale, self).shop(
+            page=page,
+            category=category,
+            product_supplier=product_supplier,
```
mFlayyan: this parameter is not in the original method!

website_sale_by_supplier/views/website_sale_by_supplier.xml
```diff
@@ -0,0 +1,49 @@
...
+		            <h4>
+	                    <a>
+	                        <span t-field="partner.name"/>
+		                    <span class="small"> 
+		                        (Total Product : 0 product(s))
+		                    </span>
+		                </a>
+		            </h4>
+		        </div>
+	        </t>
```
mFlayyan: you can use the upper one on both cases , since if products = [] then len(products)=0

mFlayyan (approved):

website_sale_by_supplier/views/website_sale_by_supplier.xml
mikevhe18: If i use the upper one on both cases, it will showing error because if products=[] then len(products) = len([]) ("object of type 'NoneType' has no len()")

mFlayyan: $ python

> > > products = []
> > > len(products)
> > > 0

mikevhe18: Oh okaay, it's because in the controller my code is:

> ```
>     if product_ids:
>         response.qcontext['products'] = product_ids
> ```

So, if not products then null not []. That's why i've got error.
I already fix it. Thanks in advance

PR #89, opened 2015-12-04
[ADD] website_sale_require_login, website_sale_suggest_create_account for 9.0

iledarn:
The modules "website_sale_require_login" and "website_sale_suggest_create_account" are all right for 9.0 without any modifications

PR #124, opened 2016-09-17
[9.0][MIG] website_sale_require_login: Upgrade to v9

lasley:
This brings website_sale_require_login changes from #89, but with proper rebasing for author attribution. 

There were also some parts that weren't upgraded & some JS Lints, so I handled those as well.

Dependent on merge of:
- [x] #125

PR #125, opened 2016-09-17
[9.0][MIG] website_sale_suggest_create_account: Upgrade to v9 

lasley:
This is the other half of #89 - website_sale_suggest_create_account, which I submitted out of order 😆


pedrobaeza:
Same comments about the commit than in #124

website_sale_suggest_create_account/README.rst
```diff
@@ -0,0 +1,78 @@
...
+   :target: https://runbot.odoo-community.org/runbot/113/9.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/
+e-commerce/issues>`_. In case of trouble, please check there if your issue has
+already been reported. If you spotted it first, help us smashing it by
+providing a detailed and welcomed feedback `here <https://github.com/OCA/
+e-commerce/issues/new?body=module:%20
```
pedrobaeza: Update this to latest version

website_sale_suggest_create_account/__openerp__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# © 2015 Antiun Ingeniería, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Suggest to create user account when buying",
+    "summary": "Suggest users to create an account when buying in the website",
+    "version": "9.0.1.0.0",
+    "category": "Website",
+    "website": "http://www.antiun.com",
+    "author": "Antiun Ingeniería S.L., "
```
pedrobaeza: Please switch this to the new company `Tecnativa`

pedrobaeza (changes requested):

website_sale_suggest_create_account/__openerp__.py
lasley: Copyrights too, or just this author key?

I'll also try to keep this in mind and apply the change in other things going forward.

lasley:
@pedrobaeza - Updates made. Because I was rebasing and destroying our conversation anyways, I went ahead and just squashed the fix commit into my original.

I would love to hear if you have a strategy for removing that merge commit, but I can't figure out a way to combine the disparate parents without it.

Side note - now that I'm looking at Github, comments aren't being collapsed after being handled anymore. Double edged sword because our conversations aren't erroneously destroyed on rebasing, but now verifying what was done just became incredibly more difficult 😦

pedrobaeza:
Yeah, the comments not dissapearing seems a bit odd. Maybe they will dissapear when approving changes...

Please remove also the merge commit and you can also squash all the Yajo's commits in one.

lasley:
@pedrobaeza - How do I remove the merge commit? These were disparate parents because the module was not on 9.0. Is there some sort of wizardry that I don't know about here? It's part of the migration instructions as well, so we should update if you've got something fancy up your sleeve.

Will do on the squashing.

pedrobaeza:
Other times that I made the method where the module is not in branch, I didn't have a merge commit, so I'm not sure why it appears. Anyway, please try the new experimental mode, that also works when the module doesn't exist, and thus we can see if we switch the instructions to this one.

lasley:
@pedrobaeza - Done

That new process was really difficult, there were conflicts almost every step of the way. I feel like I was doing something wrong, but I repeated it a few times. Can you point me to where this development is happening? I don't see an issue in maintainer-tools?

pedrobaeza:
It happened in the mailing list (some prefer this channel because is more "general audience"): https://odoo-community.org/groups/contributors-15/contributors-44716?mode=thread&date_begin=&date_end=

lasley:
Thanks @pedrobaeza - I'll mess around locally a bit more then bring it to that thread. 

For the moment I reset then did some hilariously ghetto cherry-picks off of the backup from my initial branch which seem to have achieved the intended results.

pedrobaeza (approved):

pedrobaeza:
Merging this one to speed up PR treatment.

PR #129, opened 2016-09-29
[8.0] Recently viewed products: avoid server error on race condition

LeartS:
It's possible (although rare) that a view record is created by another
worker/thread/request just after we check for its existance but before
we create a new one; that would result in a internal server error
because of the unique (sessionid, product_id) constraint.
Execute the insert transaction in a `cr.savepoint()` context so that if it
fails we just log the error and prevent the entire request tranasction
from being aborted and resulting in an internal server error.

### DIFF
--- a/website_sale_recently_viewed_products/controllers/main.py
+++ b/website_sale_recently_viewed_products/controllers/main.py
@@ -1,8 +1,15 @@
 # -*- coding: utf-8 -*-
 
+import logging
+
+from psycopg2 import IntegrityError
+
 from openerp import fields, http
 from openerp.http import request
 from openerp.addons.website_sale.controllers.main import website_sale
+
+
+_logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


gurneyalex:
:+1:

pedrobaeza:
Should it be log as an error. If this happens only on duplicates, the product is added, so no need to log so critical, isn't it?

LeartS:
Yes, in theory a warning could be better. But for the moment I am logging it as an error because the unique constraint is on the (sessionid, product_id) pair, so in order to trigger the problem the same user should request the same product in the same istant. That seems impossible to me so I'm afraid there is some other strange condition triggering the same bug, and I wanted to log these as errors to have more visibility, to possibly help debug.

pedrobaeza (approved):
OK, let's see what happens then

LeartS:
@pedrobaeza can we merge?

PR #97, opened 2016-01-22
New ecommerce wishlist module.

LeartS:
Wishlist page, "Add to wishlist" button from product page, "Move to wishlist" button from cart page.

PR #122, opened 2016-08-26
[ADD][website_sale_wishlist] Wishlist for your online shop.

yajo:
Allow your visitors and users to wishlist your products and have a shortcut for seeing them.

Wishlisted products are stored in the session for public users, and linked to the user if s/he is logged in.

@Tecnativa

PR #130, opened 2016-09-29
[8.0] New ecommerce wishlist module

LeartS:
Same as https://github.com/OCA/e-commerce/pull/97, just rebased and from another repository.

### DIFF
--- a/website_sale_wishlist/README.rst
+++ b/website_sale_wishlist/README.rst
@@ -1,0 +1,49 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+========================
+Ecommerce Wishlist
+========================
+
+Let the users save the products they are interested in and may want to buy
+at a later time.
+
+Usage
+=====
+
+Open some products pages on the e-commerce, then hover or click on
+'Recent Products' in the top menu. You should see a page or a popover (click
+vs hover) with your last viewed products.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/113/50
+
+Known Issues / Roadmap
+======================
+
+* Multiple wishlist per user support
+* Tests
+
+Credits
+=======
+
+Contributors
+------------
+
+* Leonardo Donelli <donelli@webmonks.it>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/website_sale_wishlist/README.rst
+++ b/website_sale_wishlist/README.rst
@@ -5,8 +5,9 @@
 Ecommerce Wishlist
 ========================
 
-Let the users save the products they are interested in and may want to buy
-at a later time.
+This module extends the functionality of the ecommerce module and introduces a wishlist feature.
+The users are able to save the products they are interested in in a wishlist so that they can
+confirm the purchase later on.
 
 Usage
 =====
@@ -24,6 +25,17 @@ Known Issues / Roadmap
 
 * Multiple wishlist per user support
 * Tests
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/e-commerce/issues`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/e-commerce/issues/new?body=module:%20
+e-commerce%0Aversion:%20
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
 
 Credits
 =======
@@ -46,4 +58,4 @@ OCA, or the Odoo Community Association, is a nonprofit organization whose
 mission is to support the collaborative development of Odoo features and
 promote its widespread use.
 
-To contribute to this module, please visit http://odoo-community.org.
+To contribute to this module, please visit https://odoo-community.org.
--- a/website_sale_wishlist/models/wishlist.py
+++ b/website_sale_wishlist/models/wishlist.py
@@ -1,3 +1,7 @@
+# -*- coding: utf-8 -*-
+# © 2016 Leonardo Donelli
+# License AGPL-3 or later (http://www.gnu.org/licenses/agpl)
+
 from openerp import api, fields, models
 
 
--- a/website_sale_wishlist/models/wishlist.py
+++ b/website_sale_wishlist/models/wishlist.py
@@ -1,0 +1,36 @@
+from openerp import api, fields, models
+
+
+class ShopWishList(models.Model):
+    _name = 'wishlist'
+
+    user_id = fields.Many2one('res.users')
+    item_ids = fields.One2many('wishlist.item', 'wishlist_id', string='Items')
+
+    @api.multi
+    def add_product(self, product_id):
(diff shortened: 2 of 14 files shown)


website_sale_wishlist/README.rst
```diff
@@ -0,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+========================
```
lasley, line 4: `=` should only extend as far as the title:

```
=====
Title
=====
```

website_sale_wishlist/README.rst
```diff
@@ -0,0 +1,61 @@
...
+* Multiple wishlist per user support
+* Tests
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/e-commerce/issues`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
```
lasley, line 35: Please use new template text here https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst#bug-tracker

website_sale_wishlist/models/wishlist.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+# © 2016 Leonardo Donelli
+# License AGPL-3 or later (http://www.gnu.org/licenses/agpl)
+
+from openerp import api, fields, models
+
+
+class ShopWishList(models.Model):
+    _name = 'wishlist'
```
lasley, line 9: IMO this should be `product.wishlist` & `product.wishlist.item` - similar to `product.pricelist` and `product.pricelist.item`

website_sale_wishlist/models/wishlist.py
```diff
@@ -0,0 +1,40 @@
...
+
+    user_id = fields.Many2one('res.users')
+    item_ids = fields.One2many('wishlist.item', 'wishlist_id', string='Items')
+
+    @api.multi
+    def add_product(self, product_id):
```
lasley, line 15: It would be nice if this used recordset input instead - newer API model code tends to pass around Recordsets instead of IDs, so this is ambiguous IMO (even though your naming of the variable indicates that it is an ID, it's also customary to pass around Recordset singletons as `_id` variables 😑 )

Non blocker for sure, just worth noting.

website_sale_wishlist/models/wishlist.py
```diff
@@ -0,0 +1,40 @@
...
+        self.ensure_one()
+        if product_id in self.item_ids.mapped('product_id.id'):
+            return False  # already in wish list
+        self.item_ids.create(
+            {'wishlist_id': self.id, 'product_id': product_id})
+        return True
+
+    @api.multi
+    def remove_product(self, product_id):
```
lasley, line 24: Same comment re recordset vs integer IDs in models

website_sale_wishlist/static/src/css/website_sale_wishlist.css
```diff
@@ -0,0 +1,3 @@
+img.oe-image-wishlist {
```
lasley, line 1: `oe` prefix is deprecated, please switch to `o-`

lasley (changes requested):
Thanks for the contribution @LeartS - comments inline

pedrobaeza:
Maybe it's better to backport our version for 9.0: https://github.com/OCA/e-commerce/pull/122, that includes some interesting design decisions (like the absence of a specific controller for the wishlist).

lasley:
I just reviewed the other module, and I agree with @pedrobaeza - that implementation looks to be more thorough IMO

LeartS:
That is definitely a better module, but I fear the backport could take some time especially if there aren't people interested/available to do it.

pedrobaeza:
OK, let's see what people says

lasley:
I think the module should be fairly simple to backport. AFAIK nothing significant changed between 8 and 9 for e-commerce.

bealdav:
This module works well with customzied them too. Nice job @LeartS. I imagine it is not easy

It could be merged as it by alerting in Limitation "there will no upgrade path with the v9" one quoted above.

In this case, migration path is: Uninstall v8 version, Install v9 version.

I think it's acceptable for a very old Odoo version as V8

:+1:

website_sale_wishlist/static/src/css/website_sale_wishlist.css
bealdav: Hi @lasley `oe` make me think 'Odoo ERP', then no needs to deprecate, isn't it ? Joke, yes

lasley:
> In this case, migration path is: Uninstall v8 version, Install v9 version.

This would delete all the data though, I think? Wouldn't it instead just be handled in OpenUpgrade similar to a module that changed names & data schemas?

It seems that there really isn't much interest in the backport, so I don't see much reason to not push forward with this already working code. Sure, it'll create a fragmentation between an old version and a new version, but it'll add a feature to v8 without causing any harm to the instance. I think the latter wins out, mainly because we're basically dealing with legacy anyways - I definitely don't plan on making any changes to v8 modules that aren't bug fixes.

bealdav:
Yes, I plan when I'll migrate to define path upgrade with Openugrade, but I have no time for now.
I didn't understood if you're ok to merge or not :-)
Thanks

lasley:
> I didn't understood if you're ok to merge or not :-)

I'm for a merge, but it looks like a lot of my comments haven't been attended to

bealdav:
@lasley Thanks for your answer.
What is the situation ? At least 2 erp are in production with this module at this commit with existing data model.
For v9 migration it is required to move towards the new data model. We must provide the upgrade path in this case: no way to do otherwise
V8 version is somewhat a prehistorical version now, ;-) then not lot consequences to keep data model as it.

Then I ask you to authorize an exception here (according to these arguments) with a merge.
cc @LeartS

bealdav:
@lasley If I want to update translation it's easier with transiflex. Thanks

lasley:
> What is the situation ?

There are still blocking comments from my review, such as the ReadMe issues. Note that my nonblocking comment about the naming of the variable is also out of spec from our [guidelines on variable naming](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#variable-names), which I would really consider as blocking due to the existance of a rule (I didn't know this guideline last year)

bealdav:
@LeartS Do you think you could fix @lasley remarks ?

LeartS:
Sorry guys, I don't have much time to follow this; since I made this PR we migrated the project we were using this on to v10 and we are now using the other ecommerce module. I'll update it as soon as I have some extra time.
If anyone wants to make pull requests against the branch, as @bealdav did for the French translations, I'll make sure to merge them immediately.

P.S. Not to be critical, but I think  README formatting problems shouldn't be blocking, especially considering the several hundreds of open pull requests OCA has stuck in the review limbo.

bealdav:
Ok thanks @LeartS for this information. Then if I'm alone to use it, I'll fix it myself.

bealdav:
> we are now using the other ecommerce module

There is a wishlist module in v10 ?

If I need to modify something maybe I could try to move model data towards website_sale_wishlist v9 or further
@LeartS  Then could I go on the work on a new PR ?

lasley:
> P.S. Not to be critical, but I think README formatting problems shouldn't be blocking, especially considering the several hundreds of open pull requests OCA has stuck in the review limbo.

This statement is recommending that we circumvent the review process, which exists for a very good reason. I reviewed this PR one day after you submitted it, which I do not feel qualifies for being stuck in limbo. 

In this case, a temporary hold to see if anyone wanted to unify the modules & the PR was immediately picked back up once issue was raised.

A PR is not a fire and forget situation. The PRs that are truly stuck in limbo are abandoned by their author for the most part. Lowering our guidelines to allow authors to simply dump their code and for the maintainers to have to clean it up, is not a very good community spirit IMO.

Don't get me wrong, I'm not downplaying the usefulness of this (or your/any) contributions. I am trying to explain what it's like on the other side of the fence when we spend time reviewing author contributions, but the author doesn't fix or even respond to the change requests - even if in clear violation of written guidelines. You're not the only one with a lack of time here.

pedrobaeza:
As this seems to be stalled and it's for an old version (8.0), we can close it.

PR #131, opened 2016-09-30
[ADD] website_sale_qty: Website sale price tiers

obulkin:
- Add the calculated field price_quantity_tiers to product.template. It stores
  a dictionary with price tiers (quantity ranges with different unit pricing) by
  product.pricelist
- Update the website_sale.product template to show radio buttons corresponding
  to the price tiers for each product.template/product.pricelist combo
- Add JS that links the new price tier radio buttons to the existing quantity
  selection elements


lasley:
cc @laslabs

website_sale_qty/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+    'name': 'Website Sale - Price Tiers',
+    'summary': 'Add price tier radio buttons to product pages in website shop',
+    'version': '9.0.1.0.0',
+    'category': 'E-commerce',
+    'website': 'https://laslabs.com/',
+    'author': 'LasLabs',
+    'license': 'AGPL-3',
+    'application': False,
+    'installable': True,
+    'external_dependencies': {
```
lasley: remove empty keys

website_sale_qty/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Website Sale - Price Tiers',
+    'summary': 'Add price tier radio buttons to product pages in website shop',
+    'version': '9.0.1.0.0',
+    'category': 'E-commerce',
+    'website': 'https://laslabs.com/',
+    'author': 'LasLabs',
+    'license': 'AGPL-3',
```
lasley: Please switch all of this to LGPL-3. Get with me in chat when you have some time so we can discuss licensing

website_sale_qty/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+    'application': False,
+    'installable': True,
+    'external_dependencies': {
+        'python': [],
+        'bin': [],
+    },
+    'depends': [
+        'base',
```
lasley: base is inferred, remove

website_sale_qty/models/product_template.py
```diff
@@ -0,0 +1,54 @@
...
+
+        for record in self:
+            results = {}
+
+            for pricelist in pricelists:
+                pricelist_items = self.env['product.pricelist.item'].search([
+                    ('pricelist_id', '=', pricelist.id),
+                    ('product_tmpl_id', '=', record.id),
+                    '|',
+                    ('date_start', '<=', date.today()),
```
lasley: `fields.Date.today()`

website_sale_qty/models/product_template.py
```diff
@@ -0,0 +1,54 @@
...
+                    ('date_start', '=', False),
+                    '|',
+                    ('date_end', '>=', date.today()),
```
lasley: `fields.Date.today()`

website_sale_qty/models/product_template.py
```diff
@@ -0,0 +1,54 @@
...
+        pricelists = self.env['product.pricelist'].search([('id', '>', 0)])
+
+        for record in self:
+            results = {}
+
+            for pricelist in pricelists:
+                pricelist_items = self.env['product.pricelist.item'].search([
+                    ('pricelist_id', '=', pricelist.id),
+                    ('product_tmpl_id', '=', record.id),
+                    '|',
```
lasley, line 30: You can chain the ors to make this look better. `'|', '|',` Then remove the Ln31

website_sale_qty/models/product_template.py
```diff
@@ -0,0 +1,54 @@
...
+                    ('date_start', '=', False),
+                    '|',
+                    ('date_end', '>=', date.today()),
+                    ('date_end', '=', False),
+                ])
+
+                min_quantities = set([])
+                for price in pricelist_items:
+                    if price.min_quantity < 1:
+                        min_quantities.add(1)
```
lasley:
```
in_quantities.add(
    1 if price.min_quantity < 1 else price.min_quantity
)
```

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,47 @@
...
+
+odoo.define('website_sale_qty', function(require) {
+    'use strict';
+
+    var base = require('web_editor.base');
+    base.ready().done(function() {
+        var radios = $('.js-pqt-input');
+
+        if (radios.length > 0) {
+            var radio_map = {};
```
lasley: Use camelCase variables in JS where we would normally snake_case in Python/Ruby.

JS keeps with UpperCamelCase in the same spots we'd use UpperCamelCase in Python/Ruby.

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,47 @@
+/**
+*    Copyright 2016 LasLabs Inc.
+*    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+**/
+
+odoo.define('website_sale_qty', function(require) {
+    'use strict';
+
+    var base = require('web_editor.base');
```
lasley, line 9: `var _ = require('_');` - see my comment regarding underscore later for why

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,47 @@
...
+    base.ready().done(function() {
+        var radios = $('.js-pqt-input');
+
+        if (radios.length > 0) {
+            var radio_map = {};
+            radios.each(function(index, element) {
```
lasley: Use underscore for a loop without a frivolous var  - `_.each` is reversed (val, key) vs jQuery (key, val)

```
_.each(radios, function(element) {
```

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,47 @@
+/**
+*    Copyright 2016 LasLabs Inc.
+*    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+**/
+
+odoo.define('website_sale_qty', function(require) {
+    'use strict';
+
```
lasley, line 8: `var $ = require('$')` - even though it works, you haven't guaranteed that jQuery exists in the pipeline before using in L19&20

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,47 @@
...
+            var quantity_input = $('.css_quantity input');
+            var quantity_toggles = $('.css_quantity a');
+
+            // If present, this input should be selected by default
+            if (1 in radio_map) {
+                radio_map[1].prop('checked', true);
+            }
+
+            radios.click(function(event) {
+                quantity_input.val($(event.target).val());
```
lasley: `event.currentTarget`

lasley (changes requested):
Thanks @obulkin

lasley:
Seems we have a missing trigger
- Navigate to Ipad Mini
- Quantity to 5
- Price stays at $320
- Hit `-` button on qty field (to drop to qty 4)
- Hit `+` button on qty field (to go back to 5)
- Price is now $280 with $320 crossed out.

website_sale_qty/models/product_template.py
obulkin: I think that would be logically different. The current set of domains works out to:

```
pricelist_criterion and product_tmpl_criterion 
and (date_start_criterion_1 or date_start_criterion_2) 
and (date_end_criterion_1 or date_end_criterion_2)
```

But with the change, it would be:

```
pricelist_criterion and product_tmpl_criterion 
and ((date_start_criterion_1 or date_start_criterion_2) or date_end_criterion_1)
and date_end_criterion_2
```

My testing seems to confirm this.

website_sale_qty/models/product_template.py
obulkin: Is there any upside to calling this method instead?

obulkin: I suppose it'll eliminate the need for an extra import. Is that what you had in mind?

website_sale_qty/static/src/js/website_sale_qty.js
obulkin: So the presence of jQuery and Underscore globals isn't guaranteed as part of v9? I see a number of other v9 modules that aren't going through the `require` process for jQuery access.

website_sale_qty/static/src/js/website_sale_qty.js
obulkin: Good to know. If the goal is just to eliminate the unnecessary variable then I can also do the following, which seems like the best solution because it doesn't involve bringing in an extra library:

```
radios.each(function(index) {
    var $element = $(radios[index]);
    radio_map[$element.val()] = $element;
});
```

website_sale_qty/static/src/js/website_sale_qty.js
obulkin: Good catch. At least in Chrome, `event.target` works just fine here because of the way that a click on a `<label>` element produces a click on the associated `<input>`, but `event.currentTarget` would be more universal.

website_sale_qty/static/src/js/website_sale_qty.js
lasley: `_` is already included in Odoo, so you should use it. You will need to `var _ = require('_');` though

website_sale_qty/static/src/js/website_sale_qty.js
lasley: You cannot guarantee the presence of something you did not require. The other modules are wrong if they did not require things before using

website_sale_qty/models/product_template.py
lasley: You're putting an object in a domain. That's invalid. It should be a string in the proper format, which is what fields does.

website_sale_qty/models/product_template.py
lasley: Works for me

website_sale_qty/static/src/js/website_sale_qty.js
lasley: Chrome will also show issues with target vs curentTarget, but the issues are few and far between. Just a best practice thing really

website_sale_qty/static/src/js/website_sale_qty.js
obulkin: Fair enough. I'll do this moving forward.

website_sale_qty/models/product_template.py
obulkin: Ah. Got it. The relevant tests seem to pass either way, but I'm happy to change it.

website_sale_qty/static/src/js/website_sale_qty.js
obulkin: I understand that, but there must be some overhead associated with each call to `require`, and my suggested correction avoids that while also eliminating the unused variable. Is there a problem with it?

obulkin:
@lasley --- Good catch on the missing trigger. I've updated the code to fire a `change` event after the quantity input is updated via `val()`, which takes care of the issue.

website_sale_qty/static/src/js/website_sale_qty.js
lasley: The overhead is negligible. Using the underscore creates cleaner code. Please do it.

lasley:
Excellent thanks @obulkin 

I don't see new commits. Did you forget to push the code, perhaps?

obulkin:
@lasley --- I wanted to push everything up at once and was waiting for clarity around the inline comment about Underscore. The changes should be up soon.

lasley:
@obulkin - Please don't squash your commits until after reviews are over. This entire PR needs to be re-reviewed now, instead of just the diff.

lasley (approved):
code + test 👍

obulkin:
@lasley --- My bad. I'll hold off on squashing from now on.

lasley:
@obulkin - Request from customer, please make the price readout on the main product page say "Per `product.uom_id.name`":

image

website_sale_qty/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Website Sale - Price Tiers',
+    'summary': 'Add price tier radio buttons to product pages in website shop',
+    'version': '9.0.1.0.0',
+    'category': 'E-commerce',
+    'website': 'https://laslabs.com/',
```
oscarolar, line 10: Hello, can you change the website key according to this guideline https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L9

website_sale_qty/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    'author': 'LasLabs',
```
oscarolar: Hello, can you change the authors key according to this guideline:
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L10

website_sale_qty/__openerp__.py
pedrobaeza: Website key is allowed to be the company one

website_sale_qty/__openerp__.py
pedrobaeza: This comment is correct. It should include "Odoo Community Association (OCA)" for indexing purposes.

website_sale_qty/static/tests/js/website_sale_qty.tour.js
```diff
@@ -0,0 +1,105 @@
...
+            
+            {
+                title: 'Click button that decrements main quantity input',
+                element: '.css_quantity a:nth-of-type(1)',
+            },
+            {
+                title: 'Check that the radio corresponding to a quantity of 5 is now selected',
+                onload: function() {
+                    var radio = $('.js-pqt-input[value="5"]');
+                    if (!radio.is(':checked')) {
```
oscarolar, line 98: I don't think including `if` statements in tests is a good practice, can't this be replaced by?:

``` js
waitFor: '.js-pqt-input[value="5"]:checked', 
```

website_sale_qty/static/tests/js/website_sale_qty.tour.js
```diff
@@ -0,0 +1,105 @@
...
+
+            {
+                title: 'Click button that increments main quantity input',
+                element: '.css_quantity a:nth-of-type(2)',
+            },
+            {
+                title: 'Check that none of the radios are selected anymore',
+                onload: function() {
+                    var radios = $('.js-pqt-input');
+                    if (radios.is(':checked')) {
```
oscarolar, line 84: I don't think including `if` statements in tests is a good practice, can't this be replaced by?:

``` js
waitFor: '.js-pqt-input:not(:checked)', 
```

website_sale_qty/static/tests/js/website_sale_qty.tour.js
```diff
@@ -0,0 +1,105 @@
...
+                onload: function() {
+                    // Input event should be fired to properly replicate user action
+                    $('.css_quantity input').val(5).trigger('input');
+                },
+            },
+            {
+                title: 'Check that the radio corresponding to a quantity of 5 is now selected',
+                onload: function() {
+                    var radio = $('.js-pqt-input[value="5"]');
+                    if (!radio.is(':checked')) {
```
oscarolar, line 70: I don't think including `if` statements in tests is a good practice, can't this be replaced by?:

``` js
waitFor: '.js-pqt-input[value="5"]:checked', 
```

website_sale_qty/static/tests/js/website_sale_qty.tour.js
```diff
@@ -0,0 +1,105 @@
...
+                onload: function() {
+                    // Input event should be fired to properly replicate user action
+                    $('.css_quantity input').val(6).trigger('input');
+                },
+            },
+            {
+                title: 'Check that none of the radios are selected anymore',
+                onload: function() {
+                    var radios = $('.js-pqt-input');
+                    if (radios.is(':checked')) {
```
oscarolar, line 53: I don't think including `if` statements in tests is a good practice, can't this be replaced by?:

``` js
waitFor: '.js-pqt-input:not(:checked)', 
```

website_sale_qty/static/tests/js/website_sale_qty.tour.js
```diff
@@ -0,0 +1,105 @@
...
+        id: 'test_website_sale_qty',
+        name: _t('Test JS for price/quantity radio buttons on product page in website shop'),
+        path: '/shop/product/ipad-mini-9',
+        mode: 'test',
+        steps: [
+            {
+                title: 'Check that the radio corresponding to a quantity of 1 is selected on load',
+                onload: function() {
+                    var radio = $('.js-pqt-input[value="1"]');
+                    if (!radio.is(':checked')) {
```
oscarolar, line 22: Same for this case

website_sale_qty/static/tests/js/website_sale_qty.tour.js
```diff
@@ -0,0 +1,105 @@
...
+
+            {
+                title: 'Select the radio corresponding to a quantity of 5',
+                element: '.js-pqt-input[value="5"]',
+            },
+            {
+                title: 'Check that the main quantity input has been updated to 5',
+                onload: function() {
+                    var quantityInput = $('.css_quantity input');
+                    if (quantityInput.val() != 5) {
```
oscarolar, line 36: same for this case.

oscarolar (commented):
I think it this all if inside the test are not necessary, you can do all that with `waitFor` and `waitNot` keys.

website_sale_qty/__openerp__.py
lasley: @obulkin - please heed this advice for all of our submissions to OCA

lasley:
It should not show the radio button when there is only one price tier

image

lasley:
Error 500 when no pricelist tier available:

```
2016-10-17 23:59:01,684 32339 ERROR test_db openerp.addons.website.models.ir_http: 500 Internal Server Error:
...
  File "/opt/odoo/community/oca-e-commerce/website_sale_qty/models/product_template.py", line 44, in _compute_price_quantity_tiers
    record.product_variant_ids[0].id, min_qty
  File "/media/enc-data-02/odoo/openerp/models.py", line 5761, in __getitem__
    return self._browse(self.env, (self._ids[key],))
QWebException: "tuple index out of range" while evaluating
'product.price_quantity_tiers[pricelist.id]'
```

website_sale_qty/static/tests/js/website_sale_qty.tour.js
obulkin: Actually, the `waitFor`/`waitNot` approach doesn't seem to work properly here. For example, the tour will still finish successfully even if this step includes a line like `waitFor: '.js-pqt-input[value="1"]:not(:checked)'`. Something similar happens with the last step in the tour, where `waitNot: '.js-pqt-input[value="5"]:checked'` doesn't generate an error. And I don't think there's any way to check the value **property** of an `<input>` as part of a jQuery selector, which would be necessary to replace the `if` statement in the step below.

In light of all these issues, I decided to just go with the `onload` callback throughout for consistency.

obulkin:
Thanks for the feedback, everyone. I just pushed up a new commit that should address all of this with the exception of the tour feedback left by @oscarolar. I've commented on that inline.

lasley (approved):
Thanks @obulkin

lasley:
Any PSC with thoughts on this? I'd like to see this merged soon

pedrobaeza:
What about the comments from @oscarolar ?

obulkin:
@pedrobaeza - I addressed those [in line](https://github.com/OCA/e-commerce/pull/131#discussion_r83781815).

website_sale_qty/README.rst
```diff
@@ -0,0 +1,74 @@
...
+==========================
+Website Sale - Price Tiers
+==========================
+
+This module looks at the price rules for a product and uses them to calculate 
+price tiers (quantity ranges with different unit pricing). It then adds radio 
+buttons to product pages in the website shop, which allow users to easily 
+select the minimum quantity corresponding to each tier.
+
+Installation
```
pedrobaeza: Don't add these "empty" sections.

pedrobaeza (approved):
Little comment, but overall good (code review)

website_sale_qty/README.rst
obulkin: Sure thing. I'll drop them. I figured I needed to put something there to align with the OCA template.

lasley:
Looks like we're good to merge?

website_sale_qty/README.rst
```diff
@@ -0,0 +1,59 @@
...
+==========================
+Website Sale - Price Tiers
+==========================
+
+This module looks at the price rules for a product and uses them to calculate 
+price tiers (quantity ranges with different unit pricing). It then adds radio 
+buttons to product pages in the website shop, which allow users to easily 
+select the minimum quantity corresponding to each tier.
+
+Known Issues / Roadmap
```
elicoidal: I miss the runbot section

elicoidal (approved):
minor non-blocking in the README

website_sale_qty/README.rst
obulkin: Good catch. I've added a Runbot link.

lasley:
Good to merge?

PR #132, opened 2016-09-30
[9.0][MIG][website_sale_checkout_country_vat] Migrate.

yajo:
Migrated to v9, with corresponding changes.

@Tecnativa


lasley (approved):
Looks good 👍 

Also TIL `Website Snippet Country Code Dropdown` - spiffy

website_sale_checkout_country_vat/README.rst
```diff
@@ -0,0 +1,79 @@
...
+
+If you want to have a sane default also in the country address, check the
+`website_sale_default_country
+<https://github.com/OCA/e-commerce/tree/8.0/website_sale_default_country>`_
+module.
+
+Usage
+=====
+
+when a customer makes a purchase and do the checkout process the VAT country
```
pedrobaeza: s/when/When

pedrobaeza: comma after `checkout process`

website_sale_checkout_country_vat/README.rst
```diff
@@ -0,0 +1,79 @@
...
+code can be selected separated from the code itself.
```
pedrobaeza: You can add:

```
This is useful in 2 ways:

- User is implicitly forced to input (select) a country with its code.
- Error in the VAT now only can be raised by an incorrect vat, not the absence of the country code, which usually confuses user.
```

website_sale_checkout_country_vat/__openerp__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# © 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# © 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Website Sale Checkout Country VAT",
+    "summary": "Autocomplete VAT in checkout process",
+    "version": "9.0.1.0.0",
+    "category": "Website",
```
pedrobaeza, line 10: Isn't a specific category for e-commerce? We should create it if not.

pedrobaeza (changes requested):
You should squash also in pairs the two 8.0 commits from you, the OCA transbot and the removal of the en language, and finally your two 9.0 commits (till now). You can leave separated for now the commit amending these review, but we should squash it at the end also. The last 2 can be leaved until the "final squash" before merging.

website_sale_checkout_country_vat/__openerp__.py
yajo: [`website_sale` has this](https://github.com/odoo/odoo/blob/9.0/addons/website_sale/__openerp__.py#L3)...

yajo:
Readme improved, commits logically squashed.

website_sale_checkout_country_vat/__openerp__.py
pedrobaeza: But as said, we are creating a new whole set of modules that should have an specific category although the main one is not differentiated.

pedrobaeza: Let's go this way for now, as there's no convention on this on the rest of the modules.

pedrobaeza (approved):

PR #142, opened 2016-11-23
[FIX] travis: Backport travis file

lasley:
* Backport travis from v10 to fix no coverage

Note that I also removed a `pip install pytz` in the `install`. This should be in `requirements.txt` if required, so if the build fails I'll add it.


pedrobaeza (approved):
Fast-tracking this

PR #147, opened 2016-11-28
[IMP] Translate website_sale_suggest_create_account in French

cvinh:
Translate website_sale_suggest_create_account in French (v9.0)


pedrobaeza:
Please make translations on Transifex instead.

PR #149, opened 2016-12-13
[FIX][9.0][website_sale_wishlist] Make tour work in Travis.

yajo:
Remove unneeded `waitFor` condition, to see if that's what kills Travis.

This is mostly a blind shot, given it works fine when running the test in a normal browser directly, so I'll mark as WIP until travis goes green somehow.

@Tecnativa


pedrobaeza:
I'm afraid not

yajo:
Yikes, this is stupid, let's see other possibility.

yajo:
OK, last patch is still stupid but it works. I'm willing to taste the new tour engine for v10 😓

PR #150, opened 2016-12-23
[ADD] Added website_sale_line_comment

JayVora-SerpentCS:


pedrobaeza:
This module is very similar to https://github.com/OCA/e-commerce/tree/8.0/website_sale_checkout_comment. Better to migrate this one.

yajo:
Yes, and that implementation seems better: add the field to checkout form instead of submitting it independently by ajax.

I'm closing this to keep PR list clean, hoping that you migrate that addon, but please reopen if you see it convenient 😉

PR #153, opened 2017-01-17
[MIG|IMP] website_sale_qty: v10 + Default Tier

kellylougheed:
- Forward port of #131 
- Upgrade website_sale_qty to v10
- Add feature to select the lowest tier quantity as option to the "Customize" menu of the shop page
- The radio button for the lowest tier is automatically selected on the product page


lasley:
@kellylougheed's ECLA was submitted recently under "Specialty Medical Drugstore". Advise if this isn't the case & I'll follow-up.

lasley:
@Yajo - any chance you've dug into the new tours yet?

yajo:
All I have is https://www.tecnativa.com/slides/slide/the-new-way-to-develop-automated-tests-beautiful-tours-36 for now, still did not write a v10 tour personally.

lasley:
Damn guess I have to break some ground here then. Thanks for the resource @Yajo I'll point you towards whatever I come up with once it exists!

lasley:
PS can someone mark this as WIP please? Outside of the tour, we have another feature in the works too.

obulkin:
I've cleaned up this PR and made some additional improvements, so it should now be ready for review. The changes here are as follow:
* Migrate ``website_sale_qty`` to v10
* Move logic for selecting lowest price tier by default to template and make it more general
* Add demo data to remove dependence on demo data from other modules
* Make style/organizational changes to bring module closer to OCA guidelines

website_sale_qty/README.rst
```diff
@@ -0,0 +1,67 @@
...
+price tiers (quantity ranges with different unit pricing). It then adds radio 
+buttons to product pages in the website shop, which allow users to easily 
+select the minimum quantity corresponding to each tier.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/113/9.0
```
lasley: 10.0

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,41 @@
...
+    base.ready().done(function() {
+        var radios = $('.js-pqt-input');
+
+        if (radios.length > 0) {
+            var radioMap = {};
+            _.each(radios, function(element) {
+                var $element = $(element);
+                radioMap[$element.val()] = $element;
+            });
+            var quantityInput = $('.css_quantity input');
```
lasley: Prefix vars representing jQuery objects with `$`. `$quantityInput`

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,41 @@
...
+            var quantityToggles = $('.css_quantity a');
```
lasley: `$quantityToggles`

website_sale_qty/static/src/js/website_sale_qty.js
```diff
@@ -0,0 +1,41 @@
...
+
+                var newValue = quantityInput.val();
+                if (newValue in radioMap) {
+                    radioMap[newValue].prop('checked', true);
+                }
+            };
+
+            quantityInput.on('input', updateRadios);
+            quantityToggles.click(function() {
+                // This allows website_sale.website_sale to update the input first
```
lasley: might want to mention that the reasoning behind this behavior is that the timeout will put this operation to the bottom of the event queue. Typically using a timeout creates a race condition, but we're totally safe here

lasley (changes requested):
Thanks @obulkin  - Few minor comments.

website_sale_qty/static/src/js/website_sale_qty.js
obulkin: My understanding is that there are no race conditions in JS. If the same event leads to multiple messages being added to the queue using ``setTimeout`` with a delay of ``0``, then their order of execution will be unclear, but it'll still be sequential. And since ``setTimeout`` always adds messages to the queue (not just in this specific instance), it feels like overkill to explain this in a comment.

obulkin: Or were you mostly just suggesting that `updateRadios` may be dependent on the completion of some other code that could also get called using `setTimeout` and thus could be placed in the queue after `updateRadios`? This is possible, but it doesn't seem to be the case at the moment. If this is a concern, I could look for another way to make sure that it's called at the right time, such as listening for a different event, but IIRC there weren't any options like that.

lasley: > My understanding is that there are no race conditions in JS

A race condition can exist in anything. It is simply when one process depends upon something from another without an explicit sequential chain between the two. They're actually more common in JS due to the non-sequential nature of it.

Using a `setTimeout` is typically a trick that developers will use to solve race conditions. To put it into a simple example, let's pretend that the triggers for page load completion don't exist (or we don't know about them). We have a method that depends on something to be in the DOM, but the DOM isn't ready for a bit. A nasty hack would be to add a `setTimeout` of some interval that works on the developer's computer, but may not work on other slower or heavily loaded machines. It's a newb hack really, and is something I explicitly look for when reviewing JS.

In your instance you're actually doing an expert hack. The `setTimeout` of `0` is only guaranteeing that your operation is shifted to the bottom of the event queue, of which the `website_sale.website_sale` event actions are also a part of. Because those actions already exist in the queue, you are guaranteeing a solution no matter the environment.

This is alluded to in your comment, I was just suggesting that you possibly mention why it works so that someone doesn't come and try to re-engineer for some reason.

obulkin: Fair enough. I was thinking of a race condition as a scenario where multiple threads or processes are interacting with the same memory in parallel. This doesn't happen in JS, where callbacks are atomic and get executed sequentially, but if the sequence is not predictable, you basically run into the same problem of memory being in an unexpected state, so in this case it's mostly just semantics.

I'll expand on the comment a bit.

obulkin:
@lasley - I've pushed up a commit that should address your comments.

lasley (approved):

obulkin:
I've updated this PR slightly to make the [unit price wording](https://github.com/OCA/e-commerce/pull/153/files#diff-52b964a0aedbbc612495eaec753148ebR6) inactive by default. This helps to mitigate a conflict with the `website_sale_price_subtotal` module introduced [here](https://github.com/OCA/e-commerce/pull/165).

lasley:
Thanks @obulkin! I'm going to set this as a work in progress while we bring the features in from the other PR, as well as battle test this module.

obulkin:
@lasley - This PR should be ready for review now. Travis is failing due to a seemingly unrelated issue in ``website_sale_wishlist``.

lasley (approved):
Thanks @obulkin

obulkin:
I've added a fix to the ``website_sale_wishlist`` issue in #191, which is now a prerequisite for this PR.

website_sale_qty/wizards/website_config_settings.xml
```diff
@@ -0,0 +1,20 @@
...
+<!-- Copyright 2017 LasLabs Inc.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html) -->
+
+<odoo>
+    <record id="view_website_config_settings" model="ir.ui.view">
+        <field name="name">Website Settings - Implicit Price Tier</field>
+        <field name="model">website.config.settings</field>
+        <field name="inherit_id" ref="website.view_website_config_settings"/>
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='delivery']" position="inside">
```
lasley: This requires a dependency not named in the manifest

```
7/11/2017 11:51:17 AMTraceback (most recent call last):
7/11/2017 11:51:17 AM  File "/usr/local/lib/python2.7/dist-packages/werkzeug/serving.py", line 193, in run_wsgi
7/11/2017 11:51:17 AM    execute(self.server.app)
7/11/2017 11:51:17 AM  File "/usr/local/lib/python2.7/dist-packages/werkzeug/serving.py", line 181, in execute
7/11/2017 11:51:17 AM    application_iter = app(environ, start_response)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/service/server.py", line 250, in app
7/11/2017 11:51:17 AM    return self.app(e, s)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/service/wsgi_server.py", line 184, in application
7/11/2017 11:51:17 AM    return application_unproxied(environ, start_response)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/service/wsgi_server.py", line 170, in application_unproxied
7/11/2017 11:51:17 AM    result = handler(environ, start_response)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 1308, in __call__
7/11/2017 11:51:17 AM    return self.dispatch(environ, start_response)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 1282, in __call__
7/11/2017 11:51:17 AM    return self.app(environ, start_wrapped)
7/11/2017 11:51:17 AM  File "/usr/local/lib/python2.7/dist-packages/werkzeug/wsgi.py", line 599, in __call__
7/11/2017 11:51:17 AM    return self.app(environ, start_response)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 1463, in dispatch
7/11/2017 11:51:17 AM    odoo.registry(db).check_signaling()
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/__init__.py", line 52, in registry
7/11/2017 11:51:17 AM    return modules.registry.Registry(database_name)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 59, in __new__
7/11/2017 11:51:17 AM    return cls.new(db_name)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 82, in new
7/11/2017 11:51:17 AM    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 334, in load_modules
7/11/2017 11:51:17 AM    force, status, report, loaded_modules, update_module)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 237, in load_marked_modules
7/11/2017 11:51:17 AM    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 156, in load_module_graph
7/11/2017 11:51:17 AM    _load_data(cr, module_name, idref, mode, kind='data')
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 95, in _load_data
7/11/2017 11:51:17 AM    tools.convert_file(cr, module_name, filename, idref, mode, noupdate, kind, report)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 845, in convert_file
7/11/2017 11:51:17 AM    convert_xml_import(cr, module, fp, idref, mode, noupdate, report)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 915, in convert_xml_import
7/11/2017 11:51:17 AM    obj.parse(doc.getroot(), mode=mode)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 799, in parse
7/11/2017 11:51:17 AM    self._tags[rec.tag](rec, de, mode=mode)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/tools/convert.py", line 707, in _tag_record
7/11/2017 11:51:17 AM    id = self.env(context=rec_context)['ir.model.data']._update(rec_model, self.module, res, rec_id or False, not self.isnoupdate(data_node), noupdate=self.isnoupdate(data_node), mode=self.mode)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/addons/base/ir/ir_model.py", line 1228, in _update
7/11/2017 11:51:17 AM    record = record.create(values)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/addons/base/ir/ir_ui_view.py", line 362, in create
7/11/2017 11:51:17 AM    return super(View, self).create(self._compute_defaults(values))
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/connector/connector/producer.py", line 48, in create
7/11/2017 11:51:17 AM    record = super(Base, self).create(vals)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3838, in create
7/11/2017 11:51:17 AM    self._fields[key].determine_inverse(record)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1000, in determine_inverse
7/11/2017 11:51:17 AM    getattr(records, self.inverse)()
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/addons/base/ir/ir_ui_view.py", line 220, in _inverse_arch
7/11/2017 11:51:17 AM    view.write(data)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/addons/base/ir/ir_ui_view.py", line 380, in write
7/11/2017 11:51:17 AM    return super(View, self).write(self._compute_defaults(vals))
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/connector/connector/producer.py", line 56, in write
7/11/2017 11:51:17 AM    result = super(Base, self).write(vals)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3564, in write
7/11/2017 11:51:17 AM    self._write(old_vals)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 3715, in _write
7/11/2017 11:51:17 AM    self._validate_fields(vals)
7/11/2017 11:51:17 AM  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 1083, in _validate_fields
7/11/2017 11:51:17 AM    raise ValidationError("%s\n\n%s" % (_("Error while validating constraint"), tools.ustr(e)))
7/11/2017 11:51:17 AMParseError: "Error while validating constraint
7/11/2017 11:51:17 AM
7/11/2017 11:51:17 AMElement '<xpath expr="//group[@name='delivery']">' cannot be located in parent view
7/11/2017 11:51:17 AM
7/11/2017 11:51:17 AMError context:
7/11/2017 11:51:17 AMView `Website Settings - Implicit Price Tier`
7/11/2017 11:51:17 AM[view_id: 1812, xml_id: n/a, model: website.config.settings, parent_id: 826]
7/11/2017 11:51:17 AMNone" while parsing /opt/odoo/auto/addons/website_sale_qty/wizards/website_config_settings.xml:6, near
7/11/2017 11:51:17 AM<record id="view_website_config_settings" model="ir.ui.view">
7/11/2017 11:51:17 AM        <field name="name">Website Settings - Implicit Price Tier</field>
7/11/2017 11:51:17 AM        <field name="model">website.config.settings</field>
7/11/2017 11:51:17 AM        <field name="inherit_id" ref="website.view_website_config_settings"/>
7/11/2017 11:51:17 AM        <field name="arch" type="xml">
7/11/2017 11:51:17 AM            <xpath expr="//group[@name='delivery']" position="inside">
7/11/2017 11:51:17 AM                <label for="show_implicit_price_tier"/>
7/11/2017 11:51:17 AM                <div>
7/11/2017 11:51:17 AM                    <field name="show_implicit_price_tier" nolabel="1"/>
7/11/2017 11:51:17 AM                    <label for="show_implicit_price_tier" string="Add implicit tier with quantity 1 to all product pages irrespective of price rule quantity cutoffs"/>
7/11/2017 11:51:17 AM                </div>
7/11/2017 11:51:17 AM            </xpath>
7/11/2017 11:51:17 AM        </field>
7/11/2017 11:51:17 AM    </record>
```

lasley: Hmmm ok I take that back, I think this may have been due to a partial upgrade caused by lack of column. Carry on!

lasley (approved):

website_sale_qty/wizards/website_config_settings.xml
lasley: @obulkin - I figured out what this is, and it's most certainly an issue (not sure why tests aren't failing). 

You're targeting `website.view_website_config_settings` as the inherit here, but the `delivery` group is introduced in [`website_sale.website_config_settings_view_form`](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/website_sale/views/res_config_view.xml#L3)

lasley (changes requested):
https://github.com/OCA/e-commerce/pull/153#discussion_r130204639

website_sale_qty/wizards/website_config_settings.xml
obulkin: I just pushed up a fix for this. The original code was inheriting from the parent view in ``website``, which should work as long as the ``website_sale`` changes are applied first, and this seems like the expected order since ``website_sale`` is a depend for this module. I wonder why this wasn't happening in your scenario. In any event, the new approach should be less brittle.

lasley (changes requested):
Hah the PR that keeps on giving! When implicit tier is disabled, we need the JS to use the bottom tier when the page is loaded:

image

obulkin:
@lasley - I've pushed up a commit that addresses this issue by moving the logic for selecting the lowest tier from the template to the JS. The tour should cover this edge case now, and I've also changed the implicit price tier setting to be off by default.

lasley (approved):

lasley:
We're adding another feature to this so setting to WIP. Feature is to display the lowest price tier info to the products list/search page, instead of the per unit price.

obulkin:
@lasley - The new feature has been added, and the PR should be ready for review again.

website_sale_qty/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016-2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Website Sale - Price Tiers',
+    'summary': 'Add price tier radio buttons to product pages in website shop',
```
bealdav: Thanks for your contribution.

not blocking for me but why technical name and friendly name are different ?
If technical name is not explicit why not rename it ?

lasley: I'm ok with a rename

obulkin: Fair enough. The name could definitely be more descriptive, so I've changed it to ``webite_sale_price_tiers``.

pedrobaeza:
`tiers` is a plural name? You should stick to singular in module names.

obulkin:
@pedrobaeza - Singular it is.

pedrobaeza:
OK, thanks.

pedrobaeza:
Any news with this?

lasley:
Are there additional changes required/requested? We've had this deployed in prod long ago, so from our perspective it is very much ready for merge.

yajo:
This just lacks enough approvals 😕

lasley:
Welp there's the news then 😆

pedrobaeza:
Please squash at taste and we can make optimistic merging after a quick code review and green CIs.

lasley:
Squashed!

pedrobaeza (approved):
Code review

PR #157, opened 2017-02-01
[ADD] website_sale_search_autocomplete: New module

lasley:
This module utilizes `OCA/website/website_field_autocomplete` to add an autocomplete to the search for website_sale.

The drawback to the current implementation is that the products autocomplete is being appended into the website search, because there isn't an input specifically for the products search. I'm honestly not sure what other contexts this search is used in, but I'm pretty sure this could have unintended consequences somewhere.

The only thing I can think of is to recreate the input element in this module, but that kind of circumvents the whole centralization of the search field which I think would be a bad choice. Pretty sure there's no such thing as a `mode="primary"` for templates to copy like we have in the backend, which would be perfect here.

This does work FWIW, so unless someone can point me to some other uses for the search that I'm not considering, this is good for review.


lasley:
Note that if you actually perform a search, you will get a 500. This is unrelated and reported in #158

PR #155, opened 2017-01-24
[IMP] website_sale_qty: Display quantity and price of lowest tiered option

kellylougheed:
- Enhancement of #153 
- Displays quantity and price of lowest tiered option in the search results.
- Available as an option in the "Customize" dropdown menu.

PR #166, opened 2017-03-17
[10.0][IMP] website_sale_qty: Add option to hide implicit price tier

kellylougheed:
- Forward port of #155 
- website_sale_qty adds an implicit price tier with a minimum quantity of one. This PR includes an option to hide that minimum quantity if it does not appear in the price list.

no implicit


lasley:
@obulkin please take this over. I'm not sure the difference between this and #155 either - I think the two can be merged?

lasley:
Closing - we are integrating this feature into #153

PR #36, opened 2015-02-13
Remove .coveragerc

yvaucher:

PR #167, opened 2017-04-06
[10.0][ADD][website_sale_one_step_checkout] Added possibility to use one step checkout in your e-commerce w…

rruebner:
…ebsite & added handling of delivery selection in one step checkout.

This PR is a port + improvements of our origin PR [#36](https://github.com/OCA/website/pull/36). We covered all issues mentioned in the origin PR and did improvements in general.


pedrobaeza:
Isn't one step checkout available already in core?

rruebner:
@pedrobaeza Don't make me cry. But we checked odoo core and we didn't find a one step checkout implementation so far.

pedrobaeza:
OK, you're right. I read in some place about Odoo including this in core, but I have checked also and I don't find any trace of it, so I should dream it, hehe.

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+                if not order.partner_id[f]:
+                    result['partner_id'] = order.partner_id.id
+                    return result
+
+            result['success'] = True
+
+            return result
+
+        elif order.partner_id.id == request.website.user_id\
+                .sudo().partner_id.id:
```
paulius-sladkevicius: No need of this elif block, you can do `return result` instead.

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+            return result
+
+    @http.route()
+    def address(self, **post):
+        if not request.website.use_osc:
+            return super(WebsiteSale, self).address(**post)
+        else:
+            return request.redirect('/shop')
```
paulius-sladkevicius: No need to use else, you can do `return`

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @http.route(['/shop/checkout/render_address'], type='json', auth='public',
+                website=True, multilang=True)
+    def render_address(self, **kw):
+        Partner = request.env['res.partner']\
+            .with_context(show_address=1).sudo()
```
paulius-sladkevicius: Better to use boolean `True` for `show_address`

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+        order = request.website.sale_get_order(force_create=1)
```
paulius-sladkevicius: Same here

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+        def_country_id = order.partner_id.country_id
+        values, errors = {}, {}
+
+        mode = (False, False)
+        partner_id = int(kw.get('partner_id', -1))
```
paulius-sladkevicius: I would go better with something similar `int(kw['partner_id']) if 'partner_id' in 'kw' else None`, then below you can do `if partner_id` instead `if partner_id > 0`.

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+                else:
+                    shippings = Partner.search([
+                        ('id', 'child_of',
+                         order.partner_id.commercial_partner_id.ids)],
+                        order='id desc')
+                    if partner_id in shippings.mapped('id'):
+                        mode = ('edit', 'shipping')
+                    else:
+                        return Forbidden()
+                if mode:
```
paulius-sladkevicius: In case if `mode` wasn't set above `(False, False)` value will return `True`. I suggest `mode` default value set to `None`

paulius-sladkevicius: If `mode` can't be `None` then use `any(mode)`.

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+                                                              errors,
+                                                              error_msg)
+
+            if errors:
+                errors['error_message'] = error_msg
+                return {
+                    'success': False,
+                    'errors': errors
+                }
+            else:
```
paulius-sladkevicius: suggest to remove `else` and move below code to right.

website_sale_one_step_checkout/controllers/main.py
```diff
@@ -0,0 +1,225 @@
...
+    @http.route()
+    def address(self, **post):
+        if not request.website.use_osc:
+            return super(WebsiteSale, self).address(**post)
+        else:
+            return request.redirect('/shop')
+
+    @http.route(['/shop/checkout/render_address'], type='json', auth='public',
+                website=True, multilang=True)
+    def render_address(self, **kw):
```
paulius-sladkevicius: This function requires refactor and to be split to a separate functions, now it's too big.

website_sale_one_step_checkout_delivery/controllers/main.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @http.route(['/shop/checkout/change_delivery'], type='json', auth="public",
+                website=True, multilang=True)
+    def change_delivery(self, **post):
+        """If delivery method was changed in frontend.
+
+        Change and apply delivery carrier / amount to sale order.
+        """
+        order = request.website.sale_get_order()
+        carrier_id = int(post.get('carrier_id'))
```
paulius-sladkevicius: no need to do `get()` if it always will have carrier_id, otherwise it can fail when `get()` will return `False`

paulius-sladkevicius (changes requested):
Great addon, now one step checkout is popular. However needs some adjustments to have cleaner code.

Got warning e.g. when adding address:
`WARNING 24take odoo.http: <function odoo.addons.website_sale_one_step_checkout.controllers.main.checkout> returns an invalid response type for an http request`

rafaelbn:
Hi @rruebner , thanks a lot for this module! Grear contribution

How is the state of this PR? Could you please give a feedback? Are you using it in a Customer Production environment?

Could you please review comments from @paulius-sladkevicius ?

Let me know I will review it

Thanks

rafaelbn:
Rebuilding for testing 😄

rruebner:
Hi @rafaelbn, we plan to use it in one of our project this year. Because of other circumstances we couldn't continue in a faster way.

CC @reinka, if you have minute can you check if we already updated our project code.

reinka:
@rruebner @rafaelbn  Actually we have a refactored version now. We changed the address handling for public users, the form is now being shown on the page instead of a modal. So we would have to do another pull request with the most recent version I guess?

rafaelbn:
You can close this PR and open a new one or you can update this PR with the new code

Please once done, ping me! I will review the refactored version

rafaelbn:
@rruebner @reinka 

You can close this PR and open a new one or you can update this PR with the new code

Please once done, ping me! I will review the refactored version

website_sale_one_step_checkout/controllers/main.py
reinka: Done.

website_sale_one_step_checkout/controllers/main.py
reinka: This is actually taken from the original Odoo controller: https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L542

website_sale_one_step_checkout/controllers/main.py
reinka: Same, e.g. here: https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L422

website_sale_one_step_checkout/controllers/main.py
reinka: Again, see original: https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L549 and https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L565 and https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L576

website_sale_one_step_checkout/controllers/main.py
reinka: Odoo standard: https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L571

website_sale_one_step_checkout_delivery/controllers/main.py
reinka: "otherwise it can fail when `get()` will return False". Could you elaborate on this a bit, please? I'm afraid I can not follow.

website_sale_one_step_checkout/controllers/main.py
reinka: Done.

website_sale_one_step_checkout/controllers/main.py
paulius-sladkevicius: I don't think it's a good argument if it's right way if it's in Odoo core.

website_sale_one_step_checkout_delivery/controllers/main.py
paulius-sladkevicius: [get()](https://docs.python.org/2.7/library/stdtypes.html#dict.get) returns `False` if there is no key in it, so `int(False)` will fail.

website_sale_one_step_checkout/controllers/main.py
paulius-sladkevicius: I don't think it's a good argument if it's right way if it's in Odoo core.

website_sale_one_step_checkout/controllers/main.py
paulius-sladkevicius: I don't think it's a good argument if it's right way if it's in Odoo core.

website_sale_one_step_checkout/controllers/main.py
paulius-sladkevicius: I don't think it's a good argument if it's right way if it's in Odoo core.

website_sale_one_step_checkout_delivery/controllers/main.py
reinka: `int(False)` should be actually broadcasted to `0`. But since `get` defaults to `None` and `None` can not be broadcasted, this will break it. So you are right. Thanks for pointing it out. I will fix this.

paulius-sladkevicius: Sorry, yeah, it's `None`.

website_sale_one_step_checkout/controllers/main.py
reinka: Done.

website_sale_one_step_checkout/controllers/main.py
reinka: Done.

website_sale_one_step_checkout/controllers/main.py
reinka: Your suggestion breaks the functionality, this is due to the following line: https://github.com/odoo/odoo/blob/10.0/addons/website_sale/views/templates.xml#L1251 which is also used here https://github.com/OCA/e-commerce/pull/167/files#diff-beec764b1e34f3571493d0cfbf32e65eR276 

So going with your suggestion would require to adapt that line. If I remove the `or '0'` then I get a `None`. But then it is not possible to use `int(kw['partner_id'])`. So it seems to me to make your suggestion work, I either have to check for value `0` and replace it with `None` or check for `None` before doing the `int` type cast. Not sure what the best approach is here.

paulius-sladkevicius: I think it's more cleaner to check if it's `None` and if yes use `None` instead trying to assign `-1` and do further check based on it what looks like a magic if you read only `if partner_id > 0`.

reinka: Done.

website_sale_one_step_checkout/controllers/main.py
reinka: To left? Done.

website_sale_one_step_checkout/controllers/main.py
reinka: Actually I just realised the case where `mode` is not set does not occur. At this point, the `mode` will be either `('edit', 'billing')` or `('edit', 'shipping')`, otherwise `Forbidden()` would have been returned. So I think it's save to remove the `if mode:` and just move code to the left?

reinka:
@rafaelbn Sorry, I missed out your comment to ping you once the refactored version is pushed. After taking care of the previous review, I had a problem to get green travis for the PhantomJS test, even though the tests ran fine on my local machine. Didn't have immediately the time to look further into it and forgot about this PR. Maybe you can have a look now and check if the tests work fine for you  locally?

I also haven't gotten the time to look into the warning mentioned above.

rafaelbn:
Thank you reinka! I am in debt with @Benniphx because the headache in https://github.com/OCA/website/pull/36 and https://github.com/OCA/e-commerce/pull/45 :fearful: 

Great work Bloopark!

rafaelbn:
Thank you @reinka ! I am in debt with @Benniphx because the headache in https://github.com/OCA/website/pull/36 and https://github.com/OCA/e-commerce/pull/45 :fearful: 

Great work Bloopark!

rafaelbn:
Hello @paulius-sladkevicius , your review is very welcome!!! :smile: Thank you!

paulius-sladkevicius (approved):
LGTM!

pedrobaeza:
Travis is red

reinka:
As mentioned earlier, this seems to be due to the failed PhantomJS test. I could not fix it and on my local machine I did not have any errors running the tests. Has anyone else looked into this?

pedrobaeza:
Changing standard flow makes standard tests to fail obviously. Maybe what you have to do is to detect when you are in test mode, and not testing your development, to overpass it, and leave the standard behavior. Here there are 2 things involved:

* Controllers. This is easy, as you can detect easily and call super. Although not exactly the same, you can use this technique: https://github.com/OCA/partner-contact/pull/580/commits/2cf571940e5623a1e706adafb972c68b072b1ddb#diff-a3e8fe25a5f7b37dbec19d94efe671d0R18
* QWeb templates: This is more difficult. You should try to leave the xpath that are used for the tour untouched.

rafaelbn:
Hi @rruebner @reinka @paulius-sladkevicius ,

We couldn't merge this with travis red :cry: 

I get an idea from @Yajo , maybe if you propose this PR to Odoo core in master (v12) they would accept it? This would be definitely more easy to maintain! :smile:

yajo:
Yes, the main problem if merging such addon would be that it would introduce many incompatibilities:

- break tests
- cannot work along with many other add-ons that extend the checkout process

This makes it a good idea in the wrong place. Just set it as the default upstream and all other add-ons will get adapted to it wherever merged.

website_sale_one_step_checkout/models/website.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import fields, models
+
+
+class Website(models.Model):
+    _inherit = 'website'
+
+    use_osc = fields.Boolean(
+        string='Use OSC',
+        default=True,)
```
chienandalu, line 12: Maybe if you set this default to False and run your tests after install you could avoid collissions with other tests

rafaelbn:
This functionality will be available in Odoo 18 😄

PR #138, opened 2016-11-02
10.0 mig website sale require login

iledarn:
This depends on #137 and includes its commits. You have to review only commit e0cd9e355025e75465b84fe62e59b9731c6fad45

PR #170, opened 2017-04-11
[10.0][IMP] migrated sale_require_login and dependency

ovnicraft:
Module from 9.0 is working ok, migrate to 10.0 without major changes


pedrobaeza:
There's already a PR for this module. Please review #138 and maybe propose any change to that branch.

ovnicraft:
@pedrobaeza the migration hasta latest commit Nov/2016 this is an issue :(. So if there is no response about this, OCA has any definition about close the abandon PR and continues with another proposed by community ?

pedrobaeza:
Well, not exactly a procedure, but when someone doesn't respond in the PR, usual steps is to take over his/her work and create another PR. Continuing from the previous PR work or start from the scratch will depend on the complexity of the changes. Anyway, I see that this module depends on another one, so you should first migrate that one also, isn't it?

ovnicraft:
@pedrobaeza I migrate both in this PR, so i review their changes and has the same changes (module works without major changes).

Open a new PR for dependency module?

pedrobaeza:
Let's see if tomorrow there's an answer, and if not, please open first a PR for the dependency, and then we reuse this one for the website_sale_require_login.

ovnicraft:
Ok 👍

iledarn:
@pedrobaeza I would like you to take this instead of mine https://github.com/OCA/e-commerce/pull/138, since @ovnicraft has fixed what was not right and so much time has elapsed . Sorry I'm messing up

yelizariev:
Seems like @Yajo's [review](https://github.com/OCA/e-commerce/pull/138#pullrequestreview-6975264) are not taken into account.  At least two obvious requests (move assets.xml to demo and runbot link) are not done

PR #173, opened 2017-04-28
[website_sale_b2c] B2C mode in e-commerce

yajo:
This is a functional backport form Odoo 10's B2C mode. Allows you to display prices with taxes included in the online shop.

@Tecnativa


website_sale_b2c_v10/README.rst
```diff
@@ -0,0 +1,82 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/113/9.0
+
+Known issues / Roadmap
+======================
+
+* No tests.
+* When upgrading your database to Odoo 10.0, you will have to uninstall this
```
pedrobaeza: This is not needed if migration scripts are provided on OpenUpgrade, so just put "This module is not needed on v10 and without proper migration scripts, a version 9 upgraded DB can have problems".

website_sale_b2c_v10/README.rst
```diff
@@ -0,0 +1,82 @@
...
+
+Credits
+=======
+
+Contributors
+------------
+
+* Jairo Llopis <jairo.llopis@tecnativa.com>
+
+Funders
```
pedrobaeza: Remove this section, as it's used for crowdfunding and similar, and actually it hasn't been funded by Tecnativa, but the customer. Anyway, the customer here doesn't want to appear.

pedrobaeza (changes requested):
I miss some tests for checking computed fields (not website part).

pedrobaeza:
@chienandalu @luismontalba please review this functionally. Do you understand the purpose?

yajo:
Now that you point at it @pedrobaeza... I was writing the tests, but I found extremely difficult to recreate the data to work with. Basically: A product template with 2 variants, one of which needs a price supplement, and with at least one tax.

I did not want to lose so much time in testing an addon with such a short life span, but do you have any recipe for creating that data?

Thanks.

pedrobaeza:
Well, I don't have it, but it's what you have mentioned. Not too much records to create IMO...

chienandalu:
I understand it @pedrobaeza but runbot is not rebuilding :-/

pedrobaeza:
OK, wait as there are 58 in queue that hasn't been executed yet. This is due to the "OCA Transbot" translations commits that are getting late this week. You can see in the runbot top part this status:

seleccion_001

pedrobaeza (approved):

pedrobaeza:
@sergio-teruel please review

chienandalu (approved):
@Yajo @pedrobaeza I've just patched with a little fix so taxes aren't added to the computed price for every company tax configured in the product template.

For instance, if I had in a product two company taxes the calculation was:

`pricelist_price + tax_company_1 + tax_company_2`

That aside, I've done the functional testing and everything seems ok

pedrobaeza:
@cubells please review

website_sale_b2c_v10/__openerp__.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "e-Commerce B2C mode",
+    "summary": "Display prices with taxes included in your online shop",
+    "version": "9.0.1.0.0",
+    "category": "Website",
+    "website": "https://www.tecnativa.com/",
+    "author": "Tecnativa",
```
cubells: README file says module is maintained by OCA. Where is OCA here?

website_sale_b2c_v10/README.rst
```diff
@@ -0,0 +1,74 @@
...
+   taxes included.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/113/9.0
+
+Known issues / Roadmap
+======================
+
+* No tests.
```
cubells: There is not tests? Sure?

website_sale_b2c_v10/models/sale_order_line.py
```diff
@@ -0,0 +1,25 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from openerp import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    # HACK Redefine computation or we cannot override it
+    discounted_price = fields.Float(compute="_fnct_get_discounted_price")
```
cubells: In theory, compute method must start with _compute

cubells (changes requested):
there is no translation file.

Review travis because tests are failing.

website_sale_b2c_v10/models/sale_order_line.py
chienandalu: They are Odoo core methods

chienandalu:
@pedrobaeza @cubells I just sent a patch to fix a bug in the shown price when quantities are changed in the product page. Could you review again? Thanks!

cubells (changes requested):
Tests are failing. Please review.

elicoidal (changes requested):
I am very uncomfortable with the name of the module including the version...
I assume there might be a reason for that but honestly even though would strongly suggest to find a better name

pedrobaeza:
@elicoidal, yes, there's a very strong reason, and no, it mustn't be removed. Please check the README where all is perfectly explained, including its lifetime.

pedrobaeza:
This is the text:

`This is a functional backport form Odoo 10's B2C mode. Allows you to display prices with taxes included in the online shop.`

elicoidal:
@pedrobaeza The name should reflect the function, not the version. Besides it is even more confusing because naming suggest that this is for v10 actually while this is the other way around.
I still do not see any reason to have the version while it is perfectly explained in the README where and how to use this module.

pedrobaeza:
No, it doesn't suggest that. The module version is clear. The name indicates exactly the functionality: these are the B2C prices feature from v10. 

This technique is used in a lot of modules, even already merged (https://github.com/OCA/web/tree/10.0/web_widget_domain_v11).

elicoidal:
I still do not think that is a good idea: you dont care whether it has been backported or forwardported from where. Modules names should not be linked to versions but functions: otherwise are you gonna start renaming account_analytic_v8 because we forward port the function? 
Version does not bring clarity and brings a lot more issues and questions.
You can always explain where the function comes from in the README (as you did).

pedrobaeza:
I repeat: this is a very special case, as this is something only valid for this version and it helps to identify that migrating a v9 DB, this module should be removed. I vote for keeping this name.

yajo:
Hi there @elicoidal, this is a common pattern that happened many times. It's an explicit way to say that an addon backports feature X from version Y. It also means that no migrations to that version should be done. So to speak, we call this addon with an ugly name to make sure you understand it's a temporary addon that will not be required in v10+. Also this way you know you should expect the same behavior as if you were in v10, an easy way to spot bugs.

it's not the same as when Odoo removes a feature in v10 that was present in v8. If we forward-port that, it's because it's a feature that exists no more, and as such the addon will not be volatile for just one version, but will be maintained for the foreseeable future. In those cases, adding the version to the name makes no sense indeed.

yajo:
@chienandalu it seems that this feature breaks some other tours, possibly ones that expect to check the product price. Any ideas on how to fix that?

chienandalu:
@Yajo Well, it seems that website_sale_qty dislikes this part of the patch: https://github.com/OCA/e-commerce/pull/173/commits/4657cdbfca5bb64ce1640208eaf58fe911272751#diff-0cbe07dd41a95cc435ca24c4954a29ed Anyway, on a step by step check of its tests I'm getting the same result from super and in the end of the override (as it should be when no b2c is set). So no clue on why it's failing :confused:

elicoidal:
@yajo I understand the constraints or needs and the message. Still I do not think using the name is good enough/useful/rightful. And sorry to say: but you have to provide a 10-line extra explanation to it so it is definitively not a good way. Nobody in Odoo world uses this convention except for some exceptions in our repo: given the volume and audience, we should not maintain/incentive this kind of exceptions.
Explanation should be provided in the README file where it belongs with all necessary warnings and alarm bells.

@lasley @dreispt @jgrandguillaume @sbidoul @jbeficent @gurneyalex

yajo:
OK, nevermind, renamed.

yajo:
There was a bug on pricelist price computation, it's now fixed (and code is clearer by using unpacking) so tests should go 💚 now.

elicoidal (approved):

cubells (approved):
:+1: 

LGTM

pedrobaeza (approved):

PR #178, opened 2017-06-09
[9.0][ADD] website_sale_hide_discount

lasley:
* Add new module to hide the original price and discount from website sales, only showing the price dictated by the pricelist


website_sale_hide_discount/README.rst
```diff
@@ -0,0 +1,50 @@
...
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
```
yajo, line 30: Not needed.

website_sale_hide_discount/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+{
+    'name': 'Website Sale Product Hide Discount',
+    'summary': 'Only show the pricelist price in website sale, without '
+               'the original price or discount.',
+    'version': '9.0.1.0.0',
+    'category': 'Website',
+    'website': 'https://laslabs.com/',
+    'author': 'LasLabs',
```
yajo, line 12: Add OCA.

website_sale_hide_discount/views/website_sale_template.xml
```diff
@@ -0,0 +1,28 @@
...
+<!--
+    Copyright 2016-2017 LasLabs Inc.
+    License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+-->
+
+<odoo>
+
+    <template id="products_item" inherit_id="website_sale.products_item" name="Products item no list price">
+        <xpath expr="//del" position="attributes">
+            <attribute name="style">display: none;</attribute>
```
yajo, line 12: Can't this be made with a CSS sheet? (same for all replacements)

yajo (changes requested):

website_sale_hide_discount/__openerp__.py
lasley: Whoops I knew I forgot something when bringing this in from LasLabs repos!

website_sale_hide_discount/README.rst
lasley: Really? We're not including the source file though & this is included in the template. IIRC there was even a conversation specifically mentioning that this credit was here, which is why the `svg` wasn't a required inclusion. 

@pedrobaeza - you were involved in this convo IIRC, do you remember the specifics? Am I mis-remembering?

pedrobaeza: Well, if the PNG file is included, you can perfectly mention the source (although the SVG is not included). It's the same if you include an SVG from openclipart.org or similar.

yajo: Last time I talked about this, we came to the conclusion that self-crediting was not required, since OCA is already an author of all addons, so it is expected that all code & assets are automatically credited to them if not stated otherwise. This was in the template as a placeholder, just in case you use external sources for images. Some of them have CC licenses that require crediting for instance.

In any case, it does not hurt if you prefer to keep it 😉

PR #187, opened 2017-06-26
[MIG][website_sale_checkout_country_vat] Migrate to 10.0 

yajo:
@Tecnativa


website_sale_checkout_country_vat/README.rst
```diff
@@ -0,0 +1,83 @@
...
+Configuration
+=============
+
+The default country that will appear is the one of the current user, or the one
+of the Public User assigned to the website, or the one of the company that owns
+the website. Change it in the corresponding place to get a sane default.
+
+If you want to have a sane default also in the country address, check the
+`website_sale_default_country
+<https://github.com/OCA/e-commerce/tree/8.0/website_sale_default_country>`_
```
chienandalu: s/8.0/10.0

website_sale_checkout_country_vat/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+# Copyright 2016 Sergio Teruel <sergio.teruel@tecnativa.com>
+# Copyright 2016-2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Website Sale Checkout Country VAT",
+    "summary": "Autocomplete VAT in checkout process",
+    "version": "10.0.1.0.0",
+    "category": "Website",
+    'website': 'http://www.tecnativa.com',
```
chienandalu: https

chienandalu (commented):
Little comments. Runbot and Travis aren't catching this module's dependency. Rebase maybe?

chienandalu:
Runbot and Travis are failing because: `Could not execute command 'sass'`.

pedrobaeza:
@Yajo, for Travis, you need to define https://github.com/OCA/maintainer-quality-tools/blob/f54d6e33c5c48e6bd8051efe018e88ff3db669f8/sample_files/.travis.yml#L36

lasley:
@pedrobaeza - the value in the sample file is actually the default, so that isn't the issue. That key would be used to set it back versions, or use the OS one.

The issue here is that we need [bootstrap-sass](https://github.com/OCA/website/blob/10.0/.travis.yml#L31)

pedrobaeza:
But we have also an error in PhantomJS tests, so I figured out that the problem was that...

lasley:
I believe that's because the page assets [weren't able to be bundled](https://travis-ci.org/OCA/e-commerce/jobs/247416866#L901), which caused the subsequent [Phantom error due to lack of assets defined in the bundle](https://travis-ci.org/OCA/e-commerce/jobs/247416866#L929)

pedrobaeza:
OK, then let's go to get the dependency and see

pedrobaeza:
Same issue as https://github.com/OCA/maintainer-quality-tools/issues/471

lasley:
@pedrobaeza - this is fixed in v10. Just needs to rebase & remove the manual sass installation.

lasley:
At least I hope that's the case...

pedrobaeza:
@lasley I'm afraid not. There are still errors with PhantomJS and for sure in runbot...

lasley:
Damn so I'm not sure if I know how to fix this then - the are no more install errors on the RVM/compass part. How do I replicate the error on Runbot?

pedrobaeza:
@Yajo, any clue about the error?

```
AssertionError: None is not true : PhantomJS test completed without reporting success; the log may contain errors or hints.
```

lasley:
This looks like the real error I think:
```
phantomjs: u"Could not get content for /website_sale_checkout_country_vat/static/src/js/change_country.js defined in bundle 'web.assets_frontend'.\n at http://127.0.0.1:8069/web/content/506-68a014f/web.assets_frontend.js:42 (in global code)\n(leaf frame on top)"
```

pedrobaeza (approved):
Now we get it!

The error is a bit cumbersome to be read, but @lasley hit it.

@chienandalu please review

chienandalu (approved):
Tested on rubot :+1:

Issue #152, opened 2017-01-09
Error while connecting through API


flotho:
Hi,

I have the following error through API :
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/openerp/service/wsgi_server.py", line 56, in xmlrpc_return
    result = openerp.http.dispatch_rpc(service, method, params)
  File "/usr/lib/python2.7/site-packages/openerp/http.py", line 114, in dispatch_rpc
    result = dispatch(method, params)
  File "/usr/lib/python2.7/site-packages/openerp/service/common.py", line 57, in dispatch
    return g[exp_method_name](*params)
  File "/usr/lib/python2.7/site-packages/openerp/service/common.py", line 23, in exp_login
    res = security.login(db, login, password)
  File "/usr/lib/python2.7/site-packages/openerp/service/security.py", line 9, in login
    return res_users._login(db, login, password)
  File "/usr/lib/python2.7/site-packages/openerp/addons/base/res/res_users.py", line 462, in _login
    self.check_credentials(cr, user_id, password)
  File "/usr/lib/python2.7/site-packages/openerp/api.py", line 250, in wrapper
    return old_api(self, *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/openerp/api.py", line 354, in old_api
    result = method(recs, *args, **kwargs)
  File "/opt/odoo/additional_repos/OCA-ecommerce/website_sale_wishlist/models/res_users.py", line 27, in check_credentials
    self.env["product.wishlist"]._join_current_user_and_session()
  File "/usr/lib/python2.7/site-packages/openerp/api.py", line 248, in wrapper
    return new_api(self, *args, **kwargs)
  File "/opt/odoo/additional_repos/OCA-ecommerce/website_sale_wishlist/models/product_wishlist.py", line 72, in _join_current_user_and_session
    ("session", "=", request.session.sid),
  File "/usr/lib/python2.7/site-packages/werkzeug/local.py", line 343, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/usr/lib/python2.7/site-packages/werkzeug/local.py", line 302, in _get_current_object
    return self.__local()
  File "/usr/lib/python2.7/site-packages/werkzeug/local.py", line 135, in _lookup
    raise RuntimeError('object unbound')
RuntimeError: object unbound
```
Impossible to connect via the API

lasley:
[...]

PR #190, opened 2017-06-26
[FIX#152]Defensive test when request is not correctly set through

flotho:
Try to patch #152 XMLRPC-API

### DIFF
--- a/website_sale_wishlist/models/product_wishlist.py
+++ b/website_sale_wishlist/models/product_wishlist.py
@@ -68,10 +68,12 @@ class ProductWishlist(models.Model):
         user_products = self.search([
             ("user_id", "=", self.env.uid),
         ]).mapped("product_tmpl_id")
-        session_domain = [
-            ("session", "=", request.session.sid),
-            ("user_id", "=", False),
-        ]
+
+        if request:
+            session_domain = [
+                ("session", "=", request.session.sid),
+                ("user_id", "=", False),
+            ]
         # Remove session products already present for the user
         self.search(
             session_domain + [("product_tmpl_id", "in", user_products.ids)]) \


website_sale_wishlist/models/product_wishlist.py
```diff
@@ -68,10 +68,14 @@ def _join_current_user_and_session(self):
         user_products = self.search([
             ("user_id", "=", self.env.uid),
         ]).mapped("product_tmpl_id")
-        session_domain = [
-            ("session", "=", request.session.sid),
-            ("user_id", "=", False),
-        ]
+
+        if not request:
+            session_domain = [("user_id", "=", False)]
```
yajo: You'd be assigning all dangling products from all dangling sessions as favourites for this user when logging in without session.

If there's no session, just skip this whole method, since it makes no sense in such circumstance.

yajo (changes requested):

website_sale_wishlist/models/product_wishlist.py
```diff
@@ -68,10 +68,12 @@ def _join_current_user_and_session(self):
         user_products = self.search([
             ("user_id", "=", self.env.uid),
         ]).mapped("product_tmpl_id")
-        session_domain = [
-            ("session", "=", request.session.sid),
-            ("user_id", "=", False),
-        ]
+
+        if request:
```
lasley, line 72: I agree with @Yajo - this method should just be skipped with no session, because there's no possible wishlist.

IMO this should be changed back & the following should be added to top of this method:

```
if not request:
    return
```

lasley: It's also worth noting that this would still generate an error in the event of no session, because the domain is not defined in the no request case.

PR #191, opened 2017-07-11
[FIX] website_sale_wishlist: 404 in tour

obulkin:
* Update JS tour to address 404 issue happening when module is combined with OCB and ``website_sale_qty``

I'm not sure why, but the approach used here to prevent a 404 no longer works in conjunction with OCB and the latest version of ``website_sale_qty`` in https://github.com/OCA/e-commerce/pull/153: https://travis-ci.org/OCA/e-commerce/jobs/252199821#L1189. The 404 is likely happening because the next step of the tour begins before the logout process can be completed, so a better solution is to add an extra trigger that will only appear once the user is fully logged out, forcing the tour to wait.

This PR is required for merging #153.

### DIFF
--- a/website_sale_wishlist/static/src/js/wishlist_tour.js
+++ b/website_sale_wishlist/static/src/js/wishlist_tour.js
@@ -95,15 +95,9 @@ odoo.define("website_sale_wishlist.tour", function (require) {
                 run: "click",
             },
             {
-                //For some reason the first a[href=...] leads to a 404 so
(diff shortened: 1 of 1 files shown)


obulkin:
@lasley, @Yajo, @pedrobaeza

lasley (approved):

elicoidal (approved):

yajo (approved):
This seems a remainder from v9 buggy tours, thanks for removing it

PR #197, opened 2017-08-08
[ADD] website_quote_payment_messages: Add acquirer messages

lasley:
* Replace static payment status messages with the acquirer messages in the website quotes


lasley:
Hrmmm the testing is rough on this one too. There's got to be a way to use splats in the tour URIs or something...

PR #198, opened 2017-08-08
[ADD] website_quote_product_collapse: Collapsible product descriptions

lasley:
* Encapsulate product descriptions in a panel & make it collapsible (collapsed by default) in website quotes


lasley:
Crap so it seems this one is not testable. From what I can tell the tours require a static URI, but the quote needs to be dynamic due to the ID and token. Setting to needs review unless someone has some idea there.

yajo:
You can create the quote record in the unit test, and then get its url and launch the tour from there

lasley:
This is exactly the strategy I started with - problem came around with the `url` key in the tour itself. Is there a way to pass my Python-determined URL to the Javascript for the tour? 

I looked for examples in `website_quote` & a few other modules with tokens - none of them are tested though 😑

yajo:
[The `phantom_js()` 1st param is the url where to start](https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L413). Just use it there and set the tour to not require a specific url.

lasley:
> Just use it there and set the tour to not require a specific url.

Ohhhhh so this I think was the part I was missing. For some reason I guess I thought the URL key was required in the tour.

PR #199, opened 2017-08-10
[ADD] website_sale_affiliate: Create module

hughesbm:
This module adds sales affiliate management and conversion tracking to ```website_sale```.

Affiliate Program
=================

This module allows for the creation of affiliates and the tracking of sales conversions generated by them.

Usage
=====

To use this module, you must first create at least one affiliate (found in Sales/Affiliate Program).

Once an affiliate has been created, append the following to any shop url to track the affiliate's conversions:

```?ref=affiliate_id``` (where ```affiliate_id``` is the ID displayed on the affiliate's record, e.g. ```1```)

Optionally, you can add a custom key to affiliate links for tracking specific campaigns, using the following format:

```?ref=affiliate_id&key=custom_key``` (where ```custom_key``` is a url-friendly string of your choice, e.g. ```anniversary_sale```)

RFC: https://github.com/OCA/e-commerce/issues/182

### CODE REVIEW

File: website_sale_affiliate/README.rst [hunk 1]
```diff
@@ -0,0 +1,64 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl
+   :alt: License: LGPL-3
+
+=================
+Affiliate Program
+=================
+
+This module allows for creating affiliates and track their conversions.
```
lasley: This module allows for the creation of affiliates and the tracking of sales conversions generated by them.

website_sale_affiliate/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+To use this module, you must first create at least one affiliate (found in Sales/Affiliate Program).
+
+Once an affiliate has been created, append the following to any shop url to track the affiliate's conversions:
+
+```?ref=affiliate_id``` (where ```affiliate_id``` is the ID displayed on the affiliate's record, e.g. ```1```)
+
+Optionally, you can add a custom key to affiliate links for tracking specific campaigns, using the following format:
+
+```?ref=affiliate_id&key=custom_key``` (where ```custom_key``` is a url-friendly string of your choice, e.g. ```anniversary_sale```)
```
lasley: The triple tick is markdown. One tick is interpreted text in RST IIRC

website_sale_affiliate/controllers/main.py
```diff
@@ -0,0 +1,23 @@
...
+from odoo.addons.website_sale.controllers.main import WebsiteSale as Base
+
+_logger = logging.getLogger(__name__)
+
+
+class WebsiteSale(Base):
+    @route()
+    def shop(self, page=0, category=None, search='', ppg=False, **post):
+        affiliate_id = post.get('ref')
+        if affiliate_id:
```
lasley: [EAFP](https://docs.python.org/2/glossary.html#term-eafp). 
```
try:
    ...
except KeyError:
    ...
except ValueError:
    ...
# OR 
try:
    ...
except (KeyError, ValueError):
    ...
```

website_sale_affiliate/models/sale_affiliate.py
```diff
@@ -0,0 +1,74 @@
...
+    @api.model
+    def _default_sequence_id(self):
+        return self.env.ref(
+            'website_sale_affiliate.request_sequence',
+            raise_if_not_found=False,
+        )
+
+    @api.model
+    def find_from_session(self):
+        affiliate_id = request.session.get('affiliate_id')
```
lasley: Avoid the search if no affiliate in session

```
if not affiliate_id:
    return
```

website_sale_affiliate/models/sale_affiliate.py
```diff
@@ -0,0 +1,74 @@
...
+    @api.multi
+    def get_request(self):
+        self.ensure_one()
+        matching_request = self.request_ids.find_from_session()
+        if not matching_request:
+            Request = self.env['sale.affiliate.request']
+            matching_request = Request.create_from_session(self)
+        if matching_request.conversions_qualify():
+            return matching_request
+        return None
```
lasley: `return` is functionally equivalent to `return None`

website_sale_affiliate/models/sale_affiliate.py
```diff
@@ -0,0 +1,74 @@
...
+    )
+
+    @api.model
+    def _default_sequence_id(self):
+        return self.env.ref(
+            'website_sale_affiliate.request_sequence',
+            raise_if_not_found=False,
+        )
+
+    @api.model
```
lasley: [`api.model_cr_context`](https://github.com/odoo/odoo/blob/10.0/odoo/api.py#L392)

website_sale_affiliate/models/sale_affiliate_request.py
```diff
@@ -0,0 +1,97 @@
...
+    sale_ids = fields.One2many(
+        'sale.order',
+        'affiliate_request_id',
+        string='Sales',
+        help='Qualified conversions generated as a result of affiliate request'
+    )
+
+    @api.model
+    def create_from_session(self, affiliate):
+        key = request.session.get('affiliate_key')
```
lasley:
```
try:
    name = request.session['affiliate_key']
except KeyError:
    name = affiliate.sequence_id.next_by_id() 
```

website_sale_affiliate/models/sale_affiliate_request.py
```diff
@@ -0,0 +1,97 @@
...
+            name = affiliate.sequence_id.next_by_id()
+        return self.create({
+            'name': name,
+            'affiliate_id': affiliate.id,
+        })
+
+    @api.multi
+    def find_from_session(self):
+        key = request.session.get('affiliate_key')
+        if key:
```
lasley: try/except

website_sale_affiliate/models/sale_order.py
```diff
@@ -0,0 +1,23 @@
...
+        'sale.affiliate.request',
+        string='Affiliate request',
+        help='Affiliate request associated with sale order',
+    )
+
+    @api.model
+    def create(self, vals):
+        res = super(SaleOrder, self).create(vals)
+        affiliate = self.env['sale.affiliate'].find_from_session()
+        if affiliate:
```
lasley: try/except. If I'm guessing correctly, `AttributeError`

lasley (changes requested):
Amazing test coverage dude. Great job on all the edge cases, and even including the lambda defaults that our coverage checkers give us for free.

lasley:
Tried for a functional, and I didn't see an affiliate request come in. Following are replication instructions:

* http://3300500-199-a5169a.runbot2.odoo-community.org/shop/product/e-com10-apple-wireless-keyboard-16?ref=1&key=ultra-sale#
* Add to cart, checkout, pay
* Sales => Affiliate Requests
* None (sadpanda)

hughesbm:
Hmm, you're right. 😕 The same steps work locally, but not on runbot. Looking into it...

Edit: Just kidding, doesn't work locally either. I didn't account for all the routes.

website_sale_affiliate/controllers/main.py
```diff
@@ -0,0 +1,36 @@
...
+import logging
+
+from odoo.http import request, route
+from odoo.addons.website_sale.controllers.main import WebsiteSale as Base
+
+_logger = logging.getLogger(__name__)
+
+
+class WebsiteSale(Base):
+    def _store_affiliate_info(self, **kwargs):
```
simahawk: maybe you can improve simplicity and readability like this:
```python
def _store_affiliate_info(self, **kwargs):
    if 'ref' in kwargs:
        if kwargs['ref'].isdigit():
            request.session['affiliate_id'] = int(kwargs['ref'])
            if 'key' in kwargs:
                request.session['affiliate_key'] = kwargs['key']
        else:
            _logger.debug('Invalid affiliate ID value')
```
BTW maybe is better to provide better naming for those vars, like `aff_key` and `aff_ref`?

website_sale_affiliate/models/sale_affiliate_request.py
```diff
@@ -0,0 +1,93 @@
...
+            name = request.session['affiliate_key']
+        except KeyError:
+            name = affiliate.sequence_id.next_by_id()
+        return self.create({'name': name, 'affiliate_id': affiliate.id})
+
+    @api.multi
+    def find_from_session(self):
+        try:
+            key = request.session['affiliate_key']
+            affiliate_request = self.filtered(lambda r: r.name == key)
```
simahawk: do you expect to have hundreds or thousands of records here? In that case you a search as this would be really slow.

website_sale_affiliate/models/sale_order.py
```diff
@@ -0,0 +1,25 @@
...
+        help='Affiliate request associated with sale order',
+    )
+
+    @api.model
+    def create(self, vals):
+        res = super(SaleOrder, self).create(vals)
+        affiliate = self.env['sale.affiliate'].find_from_session()
+        try:
+            res.affiliate_request_id = affiliate.get_request()
+        except AttributeError:
```
simahawk: `AttributeError`?

simahawk (changes requested):

website_sale_affiliate/models/sale_order.py
lasley: `find_from_session` returns `None` in the event of no session, so attempting `get_request` would yield an `AttributeError` I believe

website_sale_affiliate/models/sale_affiliate_request.py
```diff
@@ -0,0 +1,93 @@
...
+
+from odoo import api, fields, models
+from odoo.http import request
+
+
+class AffiliateRequest(models.Model):
+    _name = 'sale.affiliate.request'
+    _order = 'create_date desc'
+
+    name = fields.Char(required=True)
```
lasley: `index=True`

website_sale_affiliate/controllers/main.py
lasley: @simahawk - there's a lot of ifs there, vs EAFP which is more Pythonic. 

@hughesbm - Readability would be improved if you reduced nesting. Keep in mind, we want the less nested code as possible in basically all instances.

```
    def _store_affiliate_info(self, **kwargs):
        try:
            request.session['affiliate_id'] = int(kwargs['ref'])
        except KeyError:
            pass
        except ValueError:
            _logger.debug('Invalid affiliate ID value')

        try:
            request.session['affiliate_key'] = kwargs['key']
        except KeyError:
            request.session.pop('affiliate_key', None)
```

website_sale_affiliate/models/sale_affiliate_request.py
```diff
@@ -67,15 +67,17 @@ def create_from_session(self, affiliate):
...
+    @api.model_cr_context
+    def find_from_session(self, affiliate):
+        """Find affiliate request record based on session contents"""
+        domain = [('affiliate_id', '=', affiliate.id)]
         try:
             key = request.session['affiliate_key']
-            affiliate_request = self.filtered(lambda r: r.name == key)
+            domain.append(('name', '=', key))
         except KeyError:
             ip = request.httprequest.headers.environ.get('REMOTE_ADDR')
```
lasley: Can remove this part - we'll eval whether IP should be used later per readme

lasley (changes requested):

website_sale_affiliate/models/sale_order.py
simahawk: yes, "I believed" too :) But be explicit then and check if its `None`. Plus, if by error `get_request` goes on attr error you are going to swallow it without any good reason...

lasley: This goes against [EAFP](https://docs.python.org/2/glossary.html#term-eafp) though, which is pretty explicit in use case based on missing keys and attributes.

website_sale_affiliate/controllers/main.py
simahawk: in this case there's no real benefit and you are even writing more lines of code :stuck_out_tongue: 
Plus, as I understand the behavior you cannot have a `key` if no `id` is there so the last part of the code is totally useless if the 1st condition is not satisfied. That's why I wrapped the whole block inside ``if 'ref' in kwargs:``. I'm not a fun of too many IF blocks but in this case it makes it easier to read and understand what you are really doing :wink: 
Nevertheless, this is not blocking...

website_sale_affiliate/models/sale_order.py
simahawk: bah, I just read 4 lines of code instead of one :stuck_out_tongue_closed_eyes:

website_sale_affiliate/controllers/main.py
lasley: Yup you do have a point about the key being useless. I'm good with it either way, @hughesbm - you decide, it's your code 🚀

lasley: @hughesbm Keep in mind no rule is a hard rule - they can all be broken with a good reason

website_sale_affiliate/models/sale_order.py
lasley: > instead of one 

Whoah wait, one? In my head I see four (try, operation, except, pass) vs two (if, operation). Am I misunderstanding the suggestion?

Also note that when v11 comes around, things like this will instead be replaced with:
```
with contextlib.suppress(AttributeError):
    operation()
```

Which would be the same two lines, but more Pythonic. Harder to identify what to switch during migrations if we switch to LBYL IMO

simahawk: 1 line :wink:
```python
res.affiliate_request_id = affiliate.get_request() if affiliate is not None else False
```

lasley: Yeah but what about our 79 char line limit?

lasley: Although I'll totally give you the fact that your's is way cleaner if it weren't for that!

simahawk: use backslash, break to new line and shut up! :smile: :clown_face:

website_sale_affiliate/controllers/main.py
hughesbm: Thanks for the lively discussions! I appreciate the explanations/arguments for each. My nested try/except statements were pretty ugly, and either of your suggested changes would be an improvement. I'm going with the flatter try/except suggestion here. I think it's easier to follow, and though the key is useless without an ID, it's also harmless.

lasley (approved):

lasley (changes requested):
Doesn't seem to work functionally still:
* Runbot
* Create affiliate
* /shop/product/e-com08-apple-in-ear-headphones-14?aff_ref=1&aff_key=Test
  * Also tested http://3300745-199-2ddfdf.runbot2.odoo-community.org/shop?aff_ref=1&aff_key=Test
* No affiliate request created

Runbot (if still live): http://3300745-199-2ddfdf.runbot2.odoo-community.org/shop/product/e-com08-apple-in-ear-headphones-14?aff_ref=1&aff_key=Test

hughesbm:
Hmm, seems to work fine for me there. You didn't mention the checkout process this time, did you do that?

lasley:
Oops I did not. Let's save the request regardless of a sale though - this type of information will be useful in identifying which affiliates are providing the highest conversion rates

hughesbm:
@lasley Affiliate requests are now saved independently of sale orders, which provides much more useful information about affiliates. Conversion rate and average sales per request are now displayed on the Affiliate form view. Are there any other useful metrics I should add?

lasley (approved):
Thanks @hughesbm! I can't think of any more metrics at the moment

lasley:
@simahawk can you please check to make sure your comments have been attended to and the new code looks good?

website_sale_affiliate/models/sale_affiliate_request.py
```diff
@@ -0,0 +1,86 @@
...
+        ),
+        help='Request session accept language',
+    )
+    sale_ids = fields.One2many(
+        'sale.order',
+        'affiliate_request_id',
+        string='Sales',
+        help='Qualified conversions generated as a result of affiliate request'
+    )
+
```
lasley, line 67: We need to save the affiliate reference too

lasley (changes requested):

website_sale_affiliate/models/sale_affiliate_request.py
hughesbm: The `aff_key` kwarg is stored in the affiliate request's `name` field. I added a help parameter to the field for clarity.

lasley (approved):
Woot thanks @hughesbm

simahawk (approved):

PR #202, opened 2017-08-24
[10.0][ADD] website_sale_select_qty: Quantity URL param

obulkin:
* Overload ``website_sale`` controller action ``product`` and modify associated template to provide support for selecting a quantity via URL param on shop product pages
* Add unit tests for new controller logic


lasley (approved):

lasley (changes requested):
It looks like this doesn't work when `website_sale_qty` is installed - is there something we can do to support that here, or will we need a glue module?

_Edit: Disregard the price change thing that was here, I forgot that's part of a module_

obulkin:
@lasley - The fix for this compatibility issue actually needed to be in ``website_sale_qty``, so I've added a commit to https://github.com/OCA/e-commerce/pull/153 that should take care of it.

lasley (approved):
Thanks @obulkin

woodbrettm (approved):
LGTM :)

PR #204, opened 2017-09-06
[FIX][website_sale_checkout_country_vat] Do not die without country

yajo:
If you leave country field empty before pressing the confirm button at checkout, you were getting a 500 error because `country_id` was `""` (str), which cannot be searched in an int field.

Now we set a default `False` value, which will return an empty recordset without that problem.

@Tecnativa


pedrobaeza (approved):

luismontalba (approved):

rafaelbn (approved):
OK, tested in runbot 👍

PR #205, opened 2017-09-06
[MIG][website_sale_require_legal] Migrate to v10

yajo:
Normal migration.

@Tecnativa


rafaelbn (changes requested):
- Go to runbot (don't make login)
- Buy and ipad
- Fill the form
- You will see field "Country" cannot be modified + as is a mandatory field you cannot continue buying

Please fix, thanks!

rafaelbn:
@Yajo it doesnt' work as expected check https://github.com/OCA/e-commerce/pull/205#pullrequestreview-60898862

website_sale_require_legal/views/website_sale.xml
```diff
@@ -0,0 +1,25 @@
...
+                    name="accepted_legal_terms"
+                    id="accepted_legal_terms"
+                    required="required"/>
+                <t t-call="website_legal_page.acceptance_full"/>
+            </label>
+        </div>
+    </template>
+
+    <template id="address" inherit_id="website_sale.address">
+        <xpath expr="//div[@class='clearfix'][last()]" position="before">
```
chienandalu: Maybe if you put it after or add another `clearfix` before the legal block it wouldn't hide the country selector.

yajo:
It should work now.

pedrobaeza (approved):
Tested on runbot

chienandalu (approved):
Tested on runbot :+1:

yajo:
@rafaelbn please re-review

rafaelbn:
Hi @Yajo , rebuilding runbot for testing! Test are pending isn't it? Thanks!

yajo:
This has tests now

pedrobaeza:
Travis is failing

yajo:
The tour wasn't being loaded, let's see now.

pedrobaeza:
Still red.

pedrobaeza:
@Yajo Travis is still red. Maybe a rebase?

yajo:
Rebased.

pedrobaeza:
This is still in red

rafaelbn (approved):
Thanks but travis is stilll :red_circle:

rafaelbn:
If travis :green_heart: @Yajo merge!

website_sale_require_legal/views/website_sale.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <template id="accept_input">
+        <div t-attf-class="form-group col-md-12 #{'has-error' if error.get('accepted_legal_terms') else ''}">
```
grindtildeath: `if error.get(...)` should be `if error and error.get(..)` to avoid following rendering error :
```
QWebException: 'NoneType' object has no attribute 'get' Traceback (most recent call last): File "/opt/odoo/src/odoo/addons/base/ir/ir_qweb/qweb.py", line 315, in _compiled_fn return compiled(self, append, values, options, log) File "<template>", line 1, in template_website_sale_require_legal_accept_input_50 AttributeError: 'NoneType' object has no attribute 'get' Error to render compiling AST AttributeError: 'NoneType' object has no attribute 'get' Template: website_sale_require_legal.accept_input Path: /templates/t/div Node: <div t-attf-class="form-group col-md-12 #{'has-error' if error.get('accepted_legal_terms') else ''}"> <label for="accepted_legal_terms" class="control-label"> <input type="checkbox" name="accepted_legal_terms" id="accepted_legal_terms" required="required" data-oe-id="1617" data-oe-model="ir.ui.view" data-oe-field="arch" data-oe-xpath="/t[1]/div[1]/label[1]/input[1]"/> <t t-call="website_legal_page.acceptance_full"/> </label> </div>
```

I opened https://github.com/Tecnativa/e-commerce/pull/3 as it's blocking for us :wink:

grindtildeath (commented):
Just wondering if it's been considered, but address page won't appear for connected user with an address already existing. Shouldn't the acceptance checkbox be moved to 'checkout' or 'payment' page ?
Or at least state this in the README ?

pedrobaeza:
@grindtildeath do you give your approval here then?

grindtildeath:
@pedrobaeza I tried it out on runbot without our customization, and to me the module doesn't work as expected, since you can place an order without having to accept explicitly.
IMO the validation checkbox should appear on the payment page, because we need legal things to be confirmed to place an order, and not to enter any address....

pedrobaeza:
Not really, as with new GDPR, giving your address and clicking on continue, is registering personal information, so we need the legal page here.

yajo:
It should go :green_heart: now.

It turns out that on v10, tour matches must be case sensitive.

rafaelbn:
@Yajo this is still not GDPR compliance as metadata of acceptance is not stored in the customer. Please review. Thanks

yajo:
In respect to the doubt about if the checkbox is asked for already logged-in customers, it is, but in the form where personal data is filled, not in the pay form. Paying is not personal data:

peek 16-05-2018 08-16

But indeed, nothing is stored. I set as WIP to fix that.

grindtildeath:
@Yajo But you had to click on the small icon to change the address, for it to be displayed. If you instead click directly on Confirm, you'll skip the checkbox and can register an order without having to explicitly accept the conditions.
To me if the module is called website_address_require_legal or website_personal_informations_require_legal, there's no issue, but we can't say sale require legal if you can register an order without having to accept explicitly....

pedrobaeza:
@grindtildeath but that's because you are already logged-in. That doesn't happen if you are not logged-in, and the address introduction is mandatory without little icon.

yajo:
Maybe we should move the check one step before, before the "Process Checkout" button? So it covers all next steps?

pedrobaeza:
But I don't really see the problem: if you have already introduced the data in a previous checkout (or in the user creation), you don't need to put again the mark. If not, address will be shown and the check will be mandatory.

yajo:
I think his point is beyond GDPR: accepting terms of service for current order. In case such terms changed from last buy, it makes the user responsible for accepting them.

pedrobaeza:
But doesn't it require a new page?

yajo:
It wouldn't; we'd just move it to the cart page. When you pass the cart page, you enter the place where you'll start putting personal data and buying, etc., so putting it there makes the user aware of things before adding anything meaningful to the SO.

Yikes, I hope somebody ever reads some terms of use after all this work :sweat:

pedrobaeza:
OK, go ahead.

grindtildeath:
@Yajo You got it :sweat_smile: 
@pedrobaeza I don't know what's the best solution to this issue, but I'm just pointing towards a loophole allowing to register an order without accepting every time, what IMO doesn't match module's name and description.

pedrobaeza:
@Yajo please finish this

rafaelbn:
This matter has been double-checked with Lawyer and this approach is valid and legal.

About your @grindtildeath doubt 

> you'll skip the checkbox and can register an order without having to explicitly accept the conditions

This happens when the customer is already registered so he/she has accepted the terms before. If you change terms, then you should inform to all customers in other way. But not, is not needed to put another checkbox. Said that, it could be an improvement to allow more options in this module but we cannot afford this now because the main objective it migrating this module with GDPR compliance approach which is in the form when user fill the personal data and accept legal advice and term and conditions of the ecommerce.

Please @chienandalu finish this PR with the metadata info

Thanks

grindtildeath:
@rafaelbn Again, I understand your point of view and the goal is legit, but I persist in thinking you need to reconsider the name of the module as well as its description.
For example, if this module is called website_gdpr_compliance, or website_address_registration_require_legal, or anything which is not what it is actually, I have no issues with it.
But IMO my point still stands, that is : website_sale_require_legal is misleading, because acceptance of legal stuff is required to register information and not to register a sale order.

As I don't want to block your work, I see two easy solutions :
- Use a module name which is closer to what the module does
or
- At the very least explain this loophole in the issues roadmap section of the readme.

Don't you think that would make it much easier for who wants to use this module ?

pedrobaeza:
OK, we will add it to the know issues / roadmap section, as the technical name is correct due to the dependencies and this module is already in several versions with this name.

rafaelbn:
OK, let's go with `At the very least explain this loophole in the issues roadmap section of the readme.`

This modules was needed before GDPR (comes from 2016 in v8), this is for improving it making it GDPR compliance.

@grindtildeath I really thank for you point of view (22 days ago https://github.com/OCA/e-commerce/pull/205#pullrequestreview-117173657)  which make me to call and write some emails to be sure with several opinions (outsite Tecnativa, outside Odoo software). This make me be sure, really sure on what we make. Asking Lawyers, Ecommerce Manager and trying some other ecommerce that are not Odoo but Prestashop o Magento. :smile:  Really we want the best solution :+1: 

IMHO:
- The module name is correct  as it requires accept legal advice.
- The module is missing (in you opinion) is a new one called website_sale_require_term_of_use which apply in the Payment doesn't need yo have a checkbox. Just a text bellow (see Amazon example when I'm already registered)

term_of_use

Said that, please send me 2 or 3 ecommerce in EU to test them in both, legal and terms of use cases.

Thanks

rafaelbn (changes requested):
Tested in runbot. After filling the form <http://3327035-205-68b4d3.runbot2.odoo-community.org/shop/address> and click next you get 
500: Internal Server Error

website_sale_require_legal/controllers/main.py
```diff
@@ -34,3 +34,21 @@ def checkout_form_validate(self, mode, all_form_values, data):
...
+            metadata = "Website legal terms acceptance metadata: "
+            metadata += "\n".join(
+                "%s: %s" % (val, environ.get(val)) for val in (
+                        "REMOTE_ADDR",
+                        "HTTP_USER_AGENT",
+                        "HTTP_ACCEPT_LANGUAGE",
+                    )
+                )
+            partner_id = request.env['res.partner'].browse(res)
+            partner_id.message_post(body=metadata)
```
yajo: I imagine this message should be of type log. Don't you need to specify that? We don't want this to reach the buyer, right?

chienandalu:
@rafaelbn I couldn't reproduce your error. Can you detail STR?

rafaelbn:
@chienandalu please check video: https://drive.google.com/file/d/1ppuxpDWWMciCQGWj_C5IFA6dBPptmABm/view?usp=sharing

website_sale_require_legal/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+# Copyright 2015, 2017 Tecnativa - Jairo Llopis
+# Copyright 2016 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Require accepting legal terms to checkout",
+    "summary": "Force the user to accept legal tems to buy in the web shop",
+    "version": "10.0.1.0.0",
+    "category": "Website",
+    "website": "https://www.tecnativa.com",
+    "author": "Tecnativa, "
```
MeetKD: `'website': 'https://github.com/OCA/e-commerce',`

chienandalu:
@rafaelbn Can you test again?

chienandalu:
I don't know why LINT is failing now :confused:

simahawk:
@chienandalu 
```
website_sale_wishlist/README.rst:7: [E7901(rst-syntax-error), ]  Duplicate explicit target name: "your shop".
```
this should be the cause (we have stricter rules now for RST).

chienandalu:
Right! Fix here: https://github.com/OCA/e-commerce/pull/234

rafaelbn (approved):
Tested :+1:

rafaelbn:
Hello @grindtildeath ,

> Shouldn't the acceptance checkbox be moved to 'checkout' or 'payment' page ?

For your need you have an standard functionality. You don't need this module.



Regards

cc @Yajo @carlosdauden @sergio-teruel @chienandalu @pedrobaeza

sergio-teruel:
Yes, I think it too...

PR #209, opened 2017-10-03
[FIX] website_sale_affiliate: Handle case of no session

lasley:
* Handle the case of no web session, which will raise a `RuntimeError('object unbound')` when attempting to access it. This happens during XML record creations, but likely other places such as in the shell


hughesbm (approved):
Looks good to me!

eLBati (approved):

PR #214, opened 2017-10-23
[9.0][NEW] website_sale_stock_control: New module

sergio-teruel:
This module does not allow buy products without stock in a shop on line
cc @Tecnativa


chienandalu (commented):
~~Shouldn't availability be checked before checkout? If a cart has a product that runs out of stock before the buyer finish the purchase (in online orders this can be periods of days or even weeks) there's going to be an oversell.~~
- Forget about it. I see how it's controlled now. I'll comment on the readme anyway :smile:

website_sale_stock_control/README.rst
```diff
@@ -0,0 +1,63 @@
...
+their variants without stock.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to products and create one of type "Stockable".
+#. Select an option in stock control field.
+#. Go to shop and if product do not have stock yo will see an icon.
```
pedrobaeza: `Go to the shop and if the product doesn't have stock, you will see an icon indicating that.`

website_sale_stock_control/controllers/main.py
```diff
@@ -0,0 +1,28 @@
...
+    def get_attribute_value_ids(self, product):
+        res = super(WebsiteSale, self).get_attribute_value_ids(product)
+        variant_ids = [r[0] for r in res]
+        for r, variant in zip(
+                res, request.env['product.product'].browse(variant_ids)):
+            r.extend([variant.website_qty_available])
+        return res
+
+    def checkout_redirection(self, order):
+        res = super(WebsiteSale, self).checkout_redirection(order=order)
```
pedrobaeza: Put a comment about what this overwrite does. For example:
```
"""Avoid the checkout if there's no stock of one of the 
products in the cart.
"""
```

website_sale_stock_control/controllers/main.py
```diff
@@ -0,0 +1,28 @@
...
+        order = request.website.sale_get_order(context=request.context)
+        lines = order.order_line.filtered(lambda x: (
+            x.product_id.inventory_availability == 'always' and
+            x.product_uom_qty > x.product_id.website_qty_available)
+        )
+        if lines:
+            return request.redirect("/shop/cart")
```
pedrobaeza: You should show any kind of error message, not just redirect.

website_sale_stock_control/i18n/es.po
```diff
@@ -0,0 +1,115 @@
...
+"almacenados en esta ubicación, o cualquiera de sus hijas.\n"
+"En un contexto de un solo almacén, esto incluye los bienes almacenados en la "
+"ubicación de existencias de ese almacén, o cualquiera de sus hijas.\n"
+"En cualquier otro caso, esto incluye los bienes almacenados en cualquier "
+"ubicación de existencias de tipo 'Interna'."
+
+#. module: website_sale_stock_control
+#: selection:product.template,inventory_availability:0
+msgid "Deny buy products without stock"
+msgstr "Denegar comprar productos sin stock"
```
pedrobaeza: s/Denegar/No permitir
s/stock/existencias (no utilicemos anglicismos)

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    inventory_availability = fields.Selection([
+        ('never', 'Allow buy products without control'),
```
pedrobaeza: Better to call it: `Always allow to buy products`

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+# Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    inventory_availability = fields.Selection([
```
pedrobaeza: Please prefix it with `website_` for indicating their purpose.

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+        ('never', 'Allow buy products without control'),
+        ('always', 'Deny buy products without stock'),
```
pedrobaeza: `Don't allow to buy product without stock`

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+    ], string='Stock Control', default='never',
```
pedrobaeza: `Website stock control`

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    # To be inherit by other modules
+    website_qty_available = fields.Float(
+        related='qty_available', readonly=True, compute_sudo=True)
```
pedrobaeza: If you put compute_sudo, the quantity will be incorrect in case of multi-company.

pedrobaeza: For avoiding to retranslate the texts (string and help) from upstream, which are very long, add both here.

website_sale_stock_control/views/templates.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3). -->
+<odoo>
+
+<template id="products_item"
+          inherit_id="website_sale.products_item">
+
+    <xpath expr="//a[@t-field='product.name']/../.." position="after">
+        <!-- TODO: Remove this field, only testing purpose -->
```
pedrobaeza: Remove this

website_sale_stock_control/tests/test_website_sale_stock_control.py
```diff
@@ -0,0 +1,95 @@
...
+# Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import urllib
+from lxml import html
+
+from openerp.tests.common import HttpCase
+
+
+class WebsiteSaleStockControlCase(HttpCase):
```
pedrobaeza: Is this working?

pedrobaeza (changes requested):

website_sale_stock_control/controllers/main.py
sergio-teruel: Ok. Yes i will show a error message with JS

rafaelbn:
I cannot test it until https://github.com/OCA/e-commerce/issues/216 is fixed.

By the way: is it a good practive to test buying a product in any website_sale module? Then we can know if something fails in a normal proccess. Isn't it?

@sergio-teruel codecov/patch — 57.69% of diff hit (target 79.44%) we should improve test please

Thanks a lot! I test a little and it look very nice! 😄

pedrobaeza:
@rafaelbn coverage for this module is not going to be improved as the budget is limited.

website_sale_stock_control/controllers/main.py
sergio-teruel: @pedrobaeza is not necessary because this controller renders the cart with a icons warning and message "Temporaly out of stock"

website_sale_stock_control/models/product.py
sergio-teruel: I called it by similarity as v11, I am thinking in upgrade mode

website_sale_stock_control/models/product.py
sergio-teruel: This field is inside "website" section

website_sale_stock_control/models/product.py
pedrobaeza: OK, then please put a comment before
```python
# Called the same way as v11 website_sale_stock_control module
```
and make sure also that the values provided are the same.

website_sale_stock_control/controllers/main.py
pedrobaeza: OK, then please put it a comment before explaining it. As general rule, any thing "weird" should be commented for avoiding code reviewers questions. It's not a question of commenting everything, like "create here the record for...", but things that aren't expected to be should be indicated. In this case:
```python
# redirecting again to the cart will show the text "Temporarily out of stock"
```

website_sale_stock_control/models/product.py
pedrobaeza: Yeah, but when for example making an advanced search over the fields, you find the field "Stock Control" and you are not sure what does it means. In this case, I prefer to have as `string` "Website Stock Control", and then add as view level only "Stock Control". I know this is a general problem with usability across all Odoo, but let's contribute at least with this one. I write down in my TODO list to write a general guideline about this.

website_sale_stock_control/tests/test_website_sale_stock_control.py
sergio-teruel: Yes of course!!! jajaj in new commit :wink:

website_sale_stock_control/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/113/9.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/113/issues>`_. In case of trouble, please
```
yajo: e-commerce here

website_sale_stock_control/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
```
yajo, line 3: Isn't it AGPL?

website_sale_stock_control/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+
+{
+    "name": "Website Sale Stock Control",
+    "summary": "Do not show products without stock in shop on line",
+    "version": "9.0.1.0.0",
+    "category": "Website",
+    "website": "http://www.tecnativa.com",
+    "author": "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
```
yajo: It seems you mixed licenses in some files, please fix the wrong ones.

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    # Called the same way as v11 website_sale_stock module
+    inventory_availability = fields.Selection([
+        ('never', 'Always allow to buy products'),
+        ('always', "Don't allow to buy product without stock"),
+    ], string='Website Stock Control', default='never',
```
yajo: required?

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+    # To be inherit by other modules
```
yajo: "inherited"

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,44 @@
...
+        for product in self:
+            product.website_qty_available = product.sudo().with_context(
+                force_company=request.env.user.company_id.id
+            ).qty_available
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    # To be inherit by other modules
```
yajo: "inherited"

website_sale_stock_control/models/product.py
```diff
@@ -0,0 +1,44 @@
...
+    website_qty_available = fields.Float(
+        compute='_compute_website_qty_available',
+        readonly=True)
+
+    @api.multi
+    def _compute_website_qty_available(self):
```
yajo: Isn't this inherited from `product.template` automatically?
If not, do you think it would make sense to add this to an `AbstractModel`, so you can inherit it in both `product.template` and `product.product` and DRY? 🤔

website_sale_stock_control/static/src/xml/website_sale_stock_control_product_availability.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<templates>
+
+    <t t-name="website_sale_stock_control.no_stock">
+        <div id="no_stock" class="text-danger mt16" style="padding: 15px;">
```
yajo: Can we style from Less and not from inline styles?

website_sale_stock_control/views/templates.xml
```diff
@@ -0,0 +1,50 @@
...
+
+<template id="products_add_to_cart"
+          inherit_id="website_sale.products_add_to_cart">
+
+    <xpath expr="//t[@t-if='len(product.product_variant_ids) == 1']" position="attributes">
+        <attribute name="t-if">len(product.product_variant_ids) == 1 and (product.inventory_availability == 'never' or (product.inventory_availability == 'always' and (product.website_qty_available &gt; 0.0)))</attribute>
+    </xpath>
+
+    <xpath expr="//t[@t-if='len(product.product_variant_ids) &gt; 1']" position="attributes">
+        <attribute name="t-if">len(product.product_variant_ids) &gt; 1 and (product.inventory_availability == 'never' or (product.inventory_availability == 'always' and (product.website_qty_available &gt; 0.0)))</attribute>
```
yajo: If the field `inventory_availability` is required, this check can be simplified (since it will always be either `never` or `always`)

yajo (changes requested):

website_sale_stock_control/models/product.py
sergio-teruel: And if product type is service??

pedrobaeza: with default='never' it's not a problem

website_sale_stock_control/views/templates.xml
sergio-teruel: The field is not required, but how?

sergio-teruel:
@pedrobaeza @Yajo @chienandalu Changes done and test passed ok

website_sale_stock_control/models/product.py
sergio-teruel: Ok. I have made an abstract model

website_sale_stock_control/models/product.py
sergio-teruel: yes.. yes is for other comment...

rafaelbn (approved):
Tested functionally in runbot! Perfect! 👍 Thanks @sergio-teruel

website_sale_stock_control/README.rst
```diff
@@ -1,6 +1,6 @@
-.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
-   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
-   :alt: License: LGPL-3
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
pedrobaeza: You should make it LGPL as this is for website.

website_sale_stock_control/i18n/es.po
```diff
@@ -101,8 +101,8 @@ msgstr "Cantidad a mano"
 #. module: website_sale_stock_control
 #: model:ir.model.fields,field_description:website_sale_stock_control.field_product_product_inventory_availability
 #: model:ir.model.fields,field_description:website_sale_stock_control.field_product_template_inventory_availability
-msgid "Stock Control"
-msgstr "Control de stock"
+msgid "Website Stock Control"
+msgstr "Control de stock en website"
```
pedrobaeza: Control de existencias en sitio web

website_sale_stock_control/models/product.py
```diff
@@ -2,28 +2,38 @@
...
+from openerp.http import request
+
+
+class WebsiteStockAvailable(models.AbstractModel):
+    _name = 'website.stock.available.mixin'
+
+    # To be inherited by other modules
+    website_qty_available = fields.Float(
+        compute='_compute_website_qty_available',
+        readonly=True)
```
pedrobaeza: A compute field is by definition readonly

pedrobaeza (changes requested):
Little changes

sergio-teruel:
OKS... ready :smile:

website_sale_stock_control/models/product.py
```diff
@@ -1,29 +1,38 @@
...
+    _inherit = ['product.template', 'website.stock.available.mixin']
 
+    # Called the same way as v11 website_sale_stock module
     inventory_availability = fields.Selection([
-        ('never', 'Allow buy products without control'),
-        ('always', 'Deny buy products without stock'),
-    ], string='Stock Control', default='never',
+        ('never', 'Always allow to buy products'),
+        ('always', "Don't allow to buy product without stock"),
+    ], string='Website Stock Control', default='never',
```
yajo: So, should this be required or not? It wasn't clear last time 😕

sergio-teruel: I think that is not necessary...

pedrobaeza (approved):

pedrobaeza:
Considering that all @Yajo's concerns have been addressed or answered (the only "hot" one is about the required flag, and I agree that let's keep it as optional for avoiding problems in other modules' tests with required fields - reflected as DB constraints - that are filled by defaults not yet loaded, and thus `Integrity error`), I merge the PR.

PR #217, opened 2017-11-03
[FIX] website_sale_qty: Product page permissions

obulkin:
* Fix permission errors on website shop product pages by granting portal and public users read access to the ``product.uom`` model

### DIFF
--- a/website_sale_qty/__openerp__.py
+++ b/website_sale_qty/__openerp__.py
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
-# Copyright 2016 LasLabs Inc.
+# Copyright 2016-2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 {
     'name': 'Website Sale - Price Tiers',
     'summary': 'Add price tier radio buttons to product pages in website shop',
-    'version': '9.0.1.0.0',
+    'version': '9.0.1.0.1',
     'category': 'E-commerce',
     'website': 'https://laslabs.com/',
     'author': 'LasLabs, Odoo Community Association (OCA)',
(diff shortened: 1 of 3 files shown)


lasley (approved):
Thanks @obulkin 

In light of this, should we switch the tour to portal user?

pedrobaeza:
I don't think portal users need this permission, but website (public) ones.

lasley:
I don't see any access declarations for `product.uom` and `portal`, so IMO the issue would still trigger if the portal user is logged in

pedrobaeza:
I don't get you. Have you seen https://github.com/OCA/e-commerce/pull/217/files#diff-9aac379b6fb363949febb85be7efc44bR2 ?

lasley:
Sorry, I was saying that there is not a declaration for `product.uom` and `portal` in core Odoo. Oleg's findings are that this permission is needed for execution of [this line](https://github.com/LasLabs/e-commerce/blob/354f52fbcfafff8e5c3f52bd8827576a347b2aa7/website_sale_qty/views/website_sale_qty.xml#L11). It probably should have been explained better in the PR

pedrobaeza:
Thanks for the explanation, but I still don't see why do you need to give permission to a portal user for seeing a website page.

lasley:
If portal user is logged in, they will not be using public permissions.

pedrobaeza:
Uhm, AFAIK, a portal user should have both groups: public and portal, isn't it?

lasley:
That doesn't seem to be the case, and I don't see any sort of relation between the two group-wise. There's also quite a few examples of Odoo SA treating them independently, such as [this](https://github.com/odoo/odoo/blob/11.0/addons/website/security/website_security.xml#L52)

obulkin:
@lasley - I'll update the tour accordingly.

@pedrobaeza - As far as I can tell, the groups are independent. For example, the demo portal user is not in the public group by default.

obulkin:
I've updated the tour to not use a login at all since this functionality should all be accessible to public users.

pedrobaeza:
OK then, thanks for the clarifications.

yajo (approved):
neat!

PR #218, opened 2017-11-03
[FIX] website_sale_qty: Product page permissions

obulkin:
* Fix permission errors on website shop product pages by granting portal and public users read access to the ``product.uom`` model

### DIFF
--- a/website_sale_price_tier/__manifest__.py
+++ b/website_sale_price_tier/__manifest__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Website Sale - Price Tiers',
     'summary': 'Add price tier radio buttons to product pages in website shop',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
     'category': 'E-commerce',
     'website': 'https://laslabs.com/',
     'author': 'LasLabs, Odoo Community Association (OCA)',
@@ -16,6 +16,7 @@
         'website_sale',
     ],
     'data': [
+        'security/ir.model.access.csv',
         'static/src/xml/website_sale_price_tier.xml',
         'templates/assets.xml',
         'wizards/website_config_settings.xml',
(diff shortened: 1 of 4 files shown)


website_sale_price_tier/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     'name': 'Website Sale - Price Tiers',
     'summary': 'Add price tier radio buttons to product pages in website shop',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.0.1',
```
pedrobaeza: It should be 10.0.1.1.0, as it requires to update module.

website_sale_price_tier/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+product_uom_portal_read_access,Product UoM - Portal Read Access,product.model_product_uom,base.group_portal,1,0,0,0
```
pedrobaeza, line 2: Why do you need portal access? This is for website visitors only.

pedrobaeza (changes requested):

website_sale_price_tier/security/ir.model.access.csv
obulkin: I commented on this in the v9 PR.

website_sale_price_tier/__manifest__.py
obulkin: A module update is required, but the change is backwards compatible and is just a bugfix rather than a new feature, so changing the patch version seemed like the best option. That being said, I'm happy to use ``10.0.1.1.0`` instead.

obulkin:
@pedrobaeza - I've updated the version number as requested and modified the tour to not use a login so that permissions issues like this one can be better detected in the future.

yajo (approved):
code ok

pedrobaeza (approved):

PR #219, opened 2017-11-09
[9.0][FIX] website_sale_stock_control: Fix not check stock for non stockable products

sergio-teruel:
cc @Tecnativa


rafaelbn (approved):
Thanks @sergio-teruel , please review coverage
👍 
Could you please review @carlosdauden ?

pedrobaeza (approved):
Coverage hasn't changed.

PR #221, opened 2017-12-07
[ADD] website_sale_credit_point

simahawk:
depends on https://github.com/OCA/sale-workflow/pull/579 

@pedrobaeza this is the companion for the website :stuck_out_tongue_winking_eye: 

**TODO**
- [ ] evaluate implementation alternatives indicated on https://github.com/OCA/sale-workflow/pull/579 
- [ ] add tests

### DIFF
--- a/oca_dependencies.txt
+++ b/oca_dependencies.txt
@@ -1,0 +1,1 @@
+sale-workflow https://github.com/simahawk/sale-workflow 11.0-add-sale_credit_point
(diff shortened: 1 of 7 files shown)


pedrobaeza:
Well, in this case there's no equivalent right now with the other modules. I don't know right now what will be the behavior in case of restricting limit (I suppose a 500 error).

PR #233, opened 2018-04-11
[ADD] Website product name

gfcapalbo:
Adds a module that allows the user to specify a website name for every product and show that name on the website.
if website name does not exist will default to the normal name.
Very convenient in cases of long and unattractive technical  product names that should not be shown on the website.


website_product_name/models/product_template.py
```diff
@@ -0,0 +1,23 @@
...
+        for this in self:
+            this.display_website_name = this.website_name or this.name
+
+    website_name = fields.Char(
+        string='Product name on website',
+        help='Name to be shown on website,if not set normal name will be '
+        'shown.'
+    )
+    display_website_name = fields.Char(
+        compute=_compute_display_website_name, store=True
```
chienandalu: Wouldn't be easier to override `name_get()`. That way you could avoid to overwrite the views

website_product_name/__openerp__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Website Product Name",
+    "version": "9.0.1.0.0",
+    "author": "Therp BV, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
```
MeetKD, line 8: `'website': 'https://github.com/OCA/e-commerce',`

website_product_name/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Images
+------
+
+* Odoo Community Association: 
+  `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
```
simahawk: @gfcapalbo your name here? :)

website_product_name/models/product_template.py
simahawk: but if you do that you have to use a ctx key or request key to not override it for backend too.

website_product_name/models/product_template.py
```diff
@@ -0,0 +1,23 @@
...
+    def _compute_display_website_name(self):
+        for this in self:
+            this.display_website_name = this.website_name or this.name
+
+    website_name = fields.Char(
+        string='Product name on website',
+        help='Name to be shown on website,if not set normal name will be '
+        'shown.'
+    )
+    display_website_name = fields.Char(
```
simahawk, line 15: pls, do not link methods by reference: use string `compute='_compute_display_website_name'`
this way you can also put it below fields def ;)

website_product_name/models/product_template.py
```diff
@@ -0,0 +1,23 @@
...
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.multi
+    @api.depends('name', 'website_name')
+    def _compute_display_website_name(self):
+        for this in self:
+            this.display_website_name = this.website_name or this.name
+
+    website_name = fields.Char(
```
simahawk, line 10: do you really need 2 fields?
I'd not store it also. The computation is so trivial than a simple method would work fine IMO:
```pytho
def display_website_name(self):
    return self.website_name or self.name
```

website_product_name/views/templates.xml
```diff
@@ -0,0 +1,44 @@
...
+        <template id="website_name_product_recommended_ext" inherit_id="website_sale.recommended_products"  
+            name="Product website name for recomendations">
+            <xpath expr="//h5/a/span" position="attributes">
+                <attribute name="t-att-title">alt_product.display_website_name</attribute>
+                <attribute name="t-field">alt_product.display_website_name</attribute>
+            </xpath>
+        </template>
+        <!-- Show website name in cart too -->
+        <template id="products_website_name_cart" inherit_id="website_sale.cart_lines"  name="Product website name in cart">
+            <xpath expr="//td[@t-if='line.product_id.product_tmpl_id']//div" position="replace">
```
simahawk, line 34: wouldn't be better to select the `<a />` element directly?

website_product_name/views/views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record model="ir.ui.view" id="product_template_form_view_extended">
+        <field name="name">product.template.product.website.form.extended</field>
+        <field name="model">product.template</field>
+        <field name="inherit_id" ref="website_sale.product_template_form_view"/>
+        <field name="arch" type="xml">
+            <field name="website_style_ids"  position="after">
+                    <field name="website_name" placeholder="Name to be shown in the website"/>
```
simahawk: bad indent

simahawk (changes requested):

website_product_name/README.rst
```diff
@@ -0,0 +1,61 @@
...
+=====
+In the Product form, under the tab "sales" you will find a field named
+"Product name on website", if filled, that will be the name of that
+product on the website.
+
+
+Known Issues / Roadmap
+======================
+
+* Will not allow to edit website name from website (no inverse for computed field)
```
rafaelbn, line 23: Hi @gfcapalbo , why don't you implement an internal name better a not a website name?

- [ ] You cannot edit website name from website
- [ ] How will you handle such amount of translations? Are translations supported?
- [ ] What will happen with promotion (SEO) tools of Odoo?

This is just my comment, an opinion about if you have 5 products (lucky company) with 2 languages you have to manage 10 names. So maybe is better approach to add a field like product_name_internal

Odoo product name is complex as Partner first name and second name! it was a headache :smile: but in this case + translations! :thinking: :face_with_head_bandage: 

Regards!

rafaelbn:
Hi @gfcapalbo i set need review label but you hace already some reviews done + travis :red_circle: please check it! :+1:

website_product_name/views/templates.xml
gfcapalbo: in my original project that element  had been changed a bit. Replacing the entire <a> even if it meant repeating a line of code was more convenient.

website_product_name/README.rst
gfcapalbo: @rafaelbn thank you for your observations I agree. For now i did not test/support translations we will be working on this soon, adding new modules to this repo and porting to newer versions.

PR #236, opened 2018-06-03
11.0 mig website_sale_vat_required

ernestotejeda:
cc @Tecnativa


website_sale_vat_required/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+
+{
+    'name': 'e-commerce required VAT',
+    'summary': 'VAT number required in checkout form',
+    'version': '11.0.1.0.0',
+    'category': 'Website',
+    'author': "Agile Business Group, "
+              "Tecnativa, "
+              "Odoo Community Association (OCA)",
+    'website': 'http://www.agilebg.com',
```
chienandalu: https://github.com/OCA/e-commerce

chienandalu (changes requested):
- Check Travis
- Use the new OCA readme system: https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

rafaelbn:
cc @Tecnativa

website_sale_vat_required/static/src/js/website_sale.tour.js
```diff
@@ -1,23 +0,0 @@
-/* Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
- * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
-
-odoo.define('website_sale_vat_required.tour', function (require) {
```
pedrobaeza: If you remove this patch, then the website_sale main tour fails (as it's happening right now)

website_sale_vat_required/views/templates.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <template id="address" inherit_id="website_sale.address" name="Address Management">
+        <xpath expr="//input[@name='vat']/.." position="replace"/>
+        <xpath expr="//input[@name='company_name']/../.." position="after">
+            <t t-if="mode[1] == 'billing'">
+                <div t-attf-class="form-group #{error.get('vat') and 'has-error' or ''} col-md-6 div_vat">
```
pedrobaeza, line 8: Why do you need this?

pedrobaeza (changes requested):

website_sale_vat_required/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    'license': 'AGPL-3',
+    'depends': [
+        'website_sale',
+        'base_vat',
+    ],
+    'demo': [
+        'views/templates.xml',
```
yajo: ~~This is not a demo file. Either put it in the `demo` folder or in the `data` key.~~

website_sale_vat_required/views/templates.xml
yajo: Good question. It seems [included upstream already](https://github.com/odoo/odoo/blob/d2b7be055ad3ca2223d4e8009388c80e5057dbf2/addons/website_sale/views/templates.xml#L1162-L1166).

yajo (changes requested):

website_sale_vat_required/views/templates.xml
ernestotejeda: This migration has been made from v9 directly . In v9 the only thing you need to do is set 'vat' as mandatory field using _get_mandatory_billing_fields method. In v10 and v11, if you only do that and you don't modify the template 'address', then, when an user is logged in, odoo will not permit to complete the form until this field is filled, but this field will be hide and the user will wonder what happened. This is, because now, odoo only show this field if there is no user logged in.

yajo: That's interesting. So, Odoo doesn't let a user modify his VAT number. IMHO that makes sense; if you need to change such field, you're probably better creating a new customer.

Then maybe you should only require `vat` when he's adding a new address. I think you can check that by making sure there's no `request.params["partner_id"]`.

rafaelbn:
Please @pedrobaeza @Yajo could you update you reviews after last commits? Thanks!

rafaelbn:
@ernestotejeda please review travis :red_circle: thanks!

yajo:
You need to fix the test: https://travis-ci.org/OCA/e-commerce/jobs/388963897#L3102

rafaelbn (approved):
Functionally tested :+1:

website_sale_vat_required/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,5 @@
+* Lorenzo Battistini <lorenzo.battistini@agilebg.com>
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+    * Jairo Llopis <jairo.llopis@tecnativa.com>
+    * Ernesto Tejeda <ernesto.tejeda87@gmail.com>
```
pedrobaeza: You shouldn't put your personal e-mail here. In fact, we are even removing our corporate ones in all places. I'm changing this myself if the rest is correct.

pedrobaeza:
Travis is failing. Please check that and make the change about the email address.

yajo (approved):

pedrobaeza (approved):

PR #240, opened 2018-06-06
[10.0] ADD website_sale_charge_payment_fee

eLBati:
modules
```website_sale_charge_payment_fee```
```website_sale_charge_payment_fee_delivery```

### CODE REVIEW

File: website_sale_charge_payment_fee/__manifest__.py [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+
+{
+    "name": "eCommerce: charge payment fee",
+    "summary": "Payment fee charged to customer",
+    "version": "10.0.1.0.0",
+    "development_status": "Alpha",
+    "category": "Website",
+    "website": "https://github.com/OCA/e-commerce",
+    "author": "Agile Business Group, Odoo Community Association (OCA)",
+    "maintainers": [],
```
MeetKD: Can you pls remove empty key?

MeetKD (changes requested):

eLBati:
@MeetKD done

website_sale_charge_payment_fee/controllers/website_sale.py
```diff
@@ -0,0 +1,31 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from odoo.http import request
+from odoo import http
+from odoo.addons.website_sale.controllers.main import WebsiteSale
+
+
+class WebsiteSaleFee(WebsiteSale):
+
+    @http.route()
```
SimoRubi, line 11: Shouldn't the route be `@http.route(['/shop/payment'], type='http', auth="public", website=True)` ?

hurrinico (approved):
👍 LGTM

SimoRubi (commented):
Looks good, waiting for runbot

Sashimi22 (approved):
Tested functionally 👍

website_sale_charge_payment_fee/controllers/website_sale.py
yajo: If the decorator was declared in the parent method, you don't need to repeat things.

SimoRubi:
👍 Thanks, didn't know that

website_sale_charge_payment_fee/controllers/website_sale.py
SimoRubi: Thanks, didn't know that, I usually keep repeating decorators in every overriding method.

yajo: Yes, you need the decorator; just not the arguments.

MeetKD (approved):

website_sale_charge_payment_fee/models/sale.py
```diff
@@ -0,0 +1,44 @@
...
+                    self.pricelist_id.currency_id.id
+                ):
+                    price = self.company_id.currency_id.with_context(
+                        date=self.date_order
+                    ).compute(price, self.pricelist_id.currency_id)
+            elif acquirer.charge_fee_type == 'percentage':
+                price = (
+                    acquirer.charge_fee_percentage / 100.0
+                ) * self.amount_total
+            self.env['sale.order.line'].create({
```
simahawk, line 28: what if the line is already there? Wouldn't be better to search for it via product?

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
```diff
@@ -0,0 +1,23 @@
+$(document).ready(function () {
```
simahawk, line 1: I'd switch this to snippet animation

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
```diff
@@ -0,0 +1,23 @@
...
+    var $oe_cart = $(".oe_cart");
+    var $selected_acquirer_id = $oe_cart.find("input[name='selected_acquirer_id']")
+    var selected_acquirer_id = $selected_acquirer_id.val()
+    $payment_method.find("input[name='acquirer']").click(function (ev) {
+        var payment_id = $(ev.currentTarget).val();
+        // we need to check this because website_sale module clicks on checked acquirer at page load:
+        // .find("input[name='acquirer']:checked").click();
+        if (payment_id != selected_acquirer_id) {
+            $pay_button.prop('disabled', true);
+            window.location.href = '/shop/payment?payment_fee_id=' + payment_id;
```
simahawk: what if you have extra args in qstring to keep?

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
```diff
@@ -0,0 +1,23 @@
...
+    // When choosing a payment, update the quotation. Disable the 'Pay
+    // Now' button to avoid being redirected to payment acquirer if the delivery carrier update is
+    // not over.
+    var $payment_method = $("#payment_method");
+    var $oe_cart = $(".oe_cart");
+    var $selected_acquirer_id = $oe_cart.find("input[name='selected_acquirer_id']")
+    var selected_acquirer_id = $selected_acquirer_id.val()
+    $payment_method.find("input[name='acquirer']").click(function (ev) {
+        var payment_id = $(ev.currentTarget).val();
+        // we need to check this because website_sale module clicks on checked acquirer at page load:
```
simahawk: ?

website_sale_charge_payment_fee/views/website_sale_view.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0"?>
+<odoo>
```
simahawk: /views/ - > /templates/

website_sale_charge_payment_fee/views/website_sale_view.xml
```diff
@@ -0,0 +1,60 @@
...
+    <template id="assets_frontend" inherit_id="website.assets_frontend">
+      <xpath expr="." position="inside">
+        <script type="text/javascript" src="/website_sale_charge_payment_fee/static/src/js/website_sale_fee.js"/>
+      </xpath>
+    </template>
+
+    <template id="payment_fee" inherit_id="website_sale.payment">
+        <xpath expr="//span[@t-if='acquirer.fees_active']" position="after">
+            <t t-if="acquirer.charge_fee">
+                <span>(Fee:</span>
```
simahawk: why closing the span here? you can nest them

website_sale_charge_payment_fee_delivery/models/delivery.py
```diff
@@ -0,0 +1,20 @@
...
+from odoo import models, api
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = 'delivery.carrier'
+
+    @api.multi
+    def get_price_available(self, order):
+        payment_fee_line = order.order_line.filtered('payment_fee_line')
+        if payment_fee_line:
```
simahawk: would be better to wrap the whole block inside `with self.env.do_in_draft()`

simahawk (changes requested):

SimoRubi (approved):
Tested in runbot

pedrobaeza:
Travis is red

website_sale_charge_payment_fee/models/sale.py
eLBati: @simahawk previous ```payment_fee_line``` would be deleted by line 13

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
eLBati: What is it? Do you have an example? Consider I followed the same approach of ```website_sale_delivery```

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
eLBati: https://github.com/odoo/odoo/blob/2198220/addons/website_sale/static/src/js/website_sale_payment.js#L40

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
eLBati: I assume I will not have. Same behaviour of ```website_sale_delivery```

website_sale_charge_payment_fee/views/website_sale_view.xml
eLBati: done

website_sale_charge_payment_fee/views/website_sale_view.xml
eLBati: done

website_sale_charge_payment_fee_delivery/models/delivery.py
eLBati: done

eLBati:
Travis :green_apple:

website_sale_charge_payment_fee/models/sale.py
simahawk: missed it :stuck_out_tongue_closed_eyes:

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
simahawk: quick example https://github.com/OCA/website-cms/blob/10.0/cms_form/static/src/js/lock_copy_paste.js
but if you search JS files for 'web_editor.snippets.animation' you'll find tons of examples.
Not mandatory at all, especially since is a migration :wink:

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
simahawk: ok, but I don't get the comment and the commented line: is a TODO then? If so, add 'TODO', if not, remove these lines :wink:

website_sale_charge_payment_fee_delivery/models/delivery.py
simahawk: hmm I don't think is going to do what you expect the way you did it.
As soon as the ctx manager is exited you won't have `is_delivery` set and your call to `get_price_available` is going to be the same.
You should wrap the call inside it  and you don't need to set it back again ;)

simahawk (changes requested):

website_sale_charge_payment_fee_delivery/models/delivery.py
eLBati: @simahawk thanks, done

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
eLBati: @simahawk improved comment: it is not a TODO, it is the reason for the presence of ```if (payment_id != selected_acquirer_id)```. Let me know if it is a bit more clear now

website_sale_charge_payment_fee/static/src/js/website_sale_fee.js
```diff
@@ -0,0 +1,25 @@
...
+    var $payment_method = $("#payment_method");
+    var $oe_cart = $(".oe_cart");
+    var $selected_acquirer_id = $oe_cart.find("input[name='selected_acquirer_id']")
+    var selected_acquirer_id = $selected_acquirer_id.val()
+    $payment_method.find("input[name='acquirer']").click(function (ev) {
+        var payment_id = $(ev.currentTarget).val();
+        // We need to check selected_acquirer_id because website_sale module clicks on checked acquirer at page load.
+        // See website_sale/static/src/js/website_sale_payment.js :
+        // .find("input[name='acquirer']:checked").click();
+        // Without this check, redirect would be performed right after page load
```
simahawk: @eLBati got it... "check" sounds more like "we have to check something" not that you have to tick/activate/enable a flag :wink: 
sounds better now, thanks!

simahawk (approved):

eLBati:
Added ```website_sale_charge_payment_fee_quote``` to make ```website_sale_charge_payment_fee``` work with ```website_quote```, but travis, about this like https://github.com/eLBati/e-commerce/blob/d3d2c3ebebe823e8da865bfabd1149d87c41ef85/website_sale_charge_payment_fee/static/src/js/website_sale_fee.js#L4 , says ```u'TypeError: undefined is not a function (evaluating \'window.location.pathname.startsWith("/shop/payment")\')```
Does anybody know why?

simahawk:
@eLBati not supported maybe? AFAIK it comes w/ ES6
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
Was it working before or is a new thing?

eLBati:
@simahawk right. Replaced with ```_.str.startsWith``` and it is green now

eLBati:
@OCA/e-commerce-maintainers mergeable?

PR #258, opened 2018-11-19
[FIX] website_sale_charge_payment_fee: Do not blow up when there is no acquirer

SimoRubi:
If there are no acquirers, gets
```
Error to render compiling AST
TypeError: 'NoneType' object has no attribute '__getitem__'
Template: 929
Path: /templates/t/t/div/div[1]/div/div[2]/input
Node: <input type="hidden" t-att-value="selected_acquirer.id if selected_acquirer else acquirers[0].id" name="selected_acquirer_id" data-oe-id="1278" data-oe-xpath="/data/xpath[3]/input" data-oe-model="ir.ui.view" data-oe-field="arch"/>
```

### DIFF
--- a/website_sale_charge_payment_fee/__manifest__.py
+++ b/website_sale_charge_payment_fee/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "eCommerce: charge payment fee",
     "summary": "Payment fee charged to customer",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
`or False`?

SimoRubi:
Yes @pedrobaeza thanks, I was feeling that something was missing.. I also upgraded the version

pedrobaeza (approved):

tafaRU (approved):

PR #260, opened 2018-11-30
[IMP] hide only categories that are "recursively empty" and collapsible category list support

francoisdelarbre:
Hello,
This PR solves the issue OCA/website#416 . 
It does 2 things:
- the website_sale_hide_empty_category also works with "Collapsible Category List" enabled. 
- the website_sale_hide_empty_category is more intuitive, a category is now hidden only if has no products and its child categories are hidden.

I think it's more intuitive like that and will be useful to lots of people.

if you, I can split this PR in two, one for "Collapsible Category List" support and the other for recursive category hiding.

It's my first contribution to OCA so if you have any remarks, don't hesitate.


website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,3 +16,17 @@ class ProductPublicCategory(models.Model):
         column1='product_public_category_id',
         column2='product_template_id',
     )
+
+    has_products_recurcive = fields.Boolean(
```
chienandalu: s/recurcive/recursive

Also better use in singular: `has_product_recursive`

website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,3 +16,17 @@ class ProductPublicCategory(models.Model):
...
+        string='This catégory or one of its children has product',
+        compute='_has_products_recurcive'
```
chienandalu: Follow the guidelines for compute methods: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst, it would be `_compute_has_product_recursive`

website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,3 +16,17 @@ class ProductPublicCategory(models.Model):
...
+    )
+
+    @api.depends('product_ids', 'child_id.has_products_recurcive')
+    def _has_products_recurcive(self):
+        for category in self:
+            category.has_products_recurcive = \
+                bool(category.product_ids or any((child.has_products_recurcive
+                                                  for child in
+                                                  category.child_id)))
+
```
chienandalu: Seems that you added an extra line at the end that makes LINT fail

chienandalu (changes requested):

francoisdelarbre:
Hi,
@chienandalu, I did some modifications from your remarks. Could you tell me what you of the last version?
Do you you think we should change the version of the module in the manifest? If so what should the version number?

houssine78:
@remytms could you make a review of this PR when you get time ?

remytms (approved):
Looks good to me. :)

remytms:
@chienandalu , does it seems ok to you too ?

website_sale_hide_empty_category/tests/__init__.py
```diff
@@ -0,0 +1 @@
+from . import test_has_product_recurcive
```
chienandalu, line 1: You missed this **recurcive** :wink:

remytms: Hi @francoisdelarbre,
Could you rename your file `website_sale_hide_empty_category/tests/test_has_product_recurcive.py` to `website_sale_hide_empty_category/tests/test_has_product_recursive.py` and update the `__init__.py` file in consequence ?

After that, I think your PR will be accepted. :)

Rad0van:
This seems to be stuck and I would like to use exactly that. @francoisdelarbre I know it's almost 2 years but are you planning to finish this?
@pedrobaeza / @chienandalu / @remytms if @francoisdelarbre does not finish it can I pick this up and redo it in another PR? Or what would be the correct approach?

pedrobaeza:
Yes, you can resume it starting with the current contribution (but squashing all commits into one), and create a new PR.

Rad0van:
> Yes, you can resume it starting with the current contribution (but squashing all commits into one), and create a new PR.

I'm on it. Was scratching my head with why it doesn't filter empty categories until I found this...

Rad0van:
> Yes, you can resume it starting with the current contribution (but squashing all commits into one), and create a new PR.

@pedrobaeza sorry to bother but I am stuck with this. The original code mostly works. However only one of the templates gets into the customization menu. My findings seem to suggest that it happens because the `website_sale.option_collapse_categories_recursive` template is not directly used - it's only used by replacing call to the `website_sale.categories_recursive` here:
https://github.com/odoo/odoo/blob/77ffd6343be21ef188046f138493e1e807090c48/addons/website_sale/views/templates.xml#L515-L522
So that's why the customization menu always contains "Hide Empty Category list" which is tied to `website_sale.option_collapse_categories_recursive` and the other one is not put into the menu.
Is there a way to create a template with customize_show that would modify 2 other templates? In that way I would extend both templates. Or activate a template depending on whether active flag is set for another customize_show template? I.e. when collapsing is enabled activate one otherwise the other. Hope it makes sense.

Rad0van:
Solved. Will publish new PR.

PR #264, opened 2019-01-17
[10.0][FIX] website_sale_charge_payment_fee: Use a browse record if there a…

SimoRubi:
…re no acquirers

### DIFF
--- a/website_sale_charge_payment_fee/controllers/website_sale.py
+++ b/website_sale_charge_payment_fee/controllers/website_sale.py
@@ -25,7 +25,9 @@ class WebsiteSaleFee(WebsiteSale):
                     'payment.acquirer'].browse(int(payment_fee_id))
                 values['selected_acquirer'] = selected_acquirer
             else:
-                selected_acquirer = values['acquirers'][0]
+                selected_acquirer = request.env['payment.acquirer'].browse()
+                if values['acquirers']:
+                    selected_acquirer = values['acquirers'][0]
             order.sudo().update_fee_line(selected_acquirer.sudo())
             return request.render("website_sale.payment", values)
         return res
--- a/website_sale_charge_payment_fee/__manifest__.py
+++ b/website_sale_charge_payment_fee/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "eCommerce: charge payment fee",
     "summary": "Payment fee charged to customer",
-    "version": "10.0.1.0.1",
+    "version": "10.0.1.0.2",
     "development_status": "Beta",
     "category": "Website",
     "website": "https://github.com/OCA/e-commerce",


SimoRubi:
@OCA/e-commerce-maintainers can someone please take a look? Thanks

SimoRubi:
@OCA/e-commerce-maintainers can someone please take a look?
How can I help this PR to be merged?
Thanks

website_sale_charge_payment_fee/controllers/website_sale.py
```diff
@@ -25,7 +25,9 @@ def payment(self, **post):
                     'payment.acquirer'].browse(int(payment_fee_id))
                 values['selected_acquirer'] = selected_acquirer
             else:
-                selected_acquirer = values['acquirers'][0]
+                selected_acquirer = request.env['payment.acquirer'].browse()
+                if values['acquirers']:
+                    selected_acquirer = values['acquirers'][0]
```
simahawk, line 30: I don't know this module but I'm a bit sceptic about this change. The original module provides a list of rendered HTML, hence what you would have here is not an ID or a recordset and the template does seem to take this into account.
Am I wrong?

See https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L747

simahawk (changes requested):

website_sale_charge_payment_fee/controllers/website_sale.py
SimoRubi: Hi, thanks for checking!
Actually the original module provides a list of records found with `search` https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L744-L746
and then appended one by one to `value['acquirers]` in https://github.com/odoo/odoo/blob/10.0/addons/website_sale/controllers/main.py#L760; the rendered HTML is just assigned to the `button` field of every acquirer in the list.

simahawk: hmm seems you are right. Sorry, I probably checked too quickly :sweat: . Then I don't see any blockers...

simahawk (approved):
reminder: version update is not needed any more ;)

SimoRubi:
@OCA/e-commerce-maintainers how can I help this PR to be merged? Thanks

simahawk:
@SimoRubi tests are failing there, can you check?

SimoRubi:
@simahawk rebased on `10.0`.
Tests are failing because they are already failing in `10.0`:
image
(from https://travis-ci.org/OCA/e-commerce/branches)

SimoRubi:
@simahawk does this PR have to wait for the fix of `10.0`?
If no-one is tackling it, I can at least try.

simahawk:
@SimoRubi 10.0 is green https://github.com/OCA/e-commerce/commits/10.0

image

What you see there is an old build. The green one is pretty old and might fail today.

In your build I see `raise UserError(_("Unable to find Wkhtmltopdf on this system. The PDF can not be created."))`.

You can rebase on 10.0 and see if it happens again.

SimoRubi:
@simahawk the last commit is _Unverified_ because it has [ci skip]:
image
That means that Travis did not run on that commit, you can verify that by clicking on the :heavy_check_mark: :
image

The last _verified_ commit is the previous one, that is red:
image
and its travis build corresponds to the latest build of branch `10.0` that is https://travis-ci.org/OCA/e-commerce/builds/564785979 and is failed.

Anyway I already rebased like I said in https://github.com/OCA/e-commerce/pull/264#issuecomment-583364617 and from that moment on, tests began failing in this PR too (they were green before, when you first tried to merge).

simahawk:
@SimoRubi oh yeah, sorry (repeat with me: don't do 3 things at the same time when answering to ppl :sweat_smile: )
Then... you know what to do: try to fix 10.0 :)

SimoRubi:
> @SimoRubi oh yeah, sorry (repeat with me: don't do 3 things at the same time when answering to ppl  )
> Then... you know what to do: try to fix 10.0 :)

@simahawk I've fixed `10.0` in https://github.com/OCA/e-commerce/pull/366 and rebased, can you try again to merge?
Thanks!

PR #267, opened 2019-02-26
[12.0][MIG] website_sale_require_legal: Migration to v12

cristinamartinrod:
cc @Tecnativa


website_sale_require_legal/oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+website
```
yajo: This file goes in the repo root, not in the addon folder.

yajo (changes requested):

website_sale_require_legal/oca_dependencies.txt
cristinamartinrod: Right! 😅

yajo (changes requested):
Code looks OK, but just a couple of details:

captura de pantalla de 2019-03-01 11-56-58

Apart from that, clicking any of them while unauthenticated directs me to a 404 error:

captura de pantalla de 2019-03-01 11-57-11

cristinamartinrod:
@Yajo This is something from v12. It's because there are times that the cart can be very long and this a more visible option in the part on the right. And I suppose that the use of different colors is to only focus attention in one of them, that is usually something normal.

The 404 error seems to be something of runbot because if you uninstall the module it continues to happen and it doesn't seem to have any additional associated permission in the module. On local I didn't detected that error. Could it be that? Is it crazy?

yajo:
> This is something from v12. It's because there are times that the cart can be very long and this a more visible option in the part on the right. And I suppose that the use of different colors is to only focus attention in one of them, that is usually something normal.

True, I didn't see it before!

OK then.



> The 404 error seems to be something of runbot because

Possibly you need to enable signup in the website. I think it's done from its settings.

I don't know if it's done now via `ir.config_parameter` for all websites, or is it done per website. Check it out and:

* If it's global, enable it automatically when installing the addon.
* If it's per website, add the instructions to the configuration section of the README, and do not display the button in websites that have it disabled.

pedrobaeza:
Shouldn't we activate it when installing this module?

yajo:
Yes, if it's global, but if it's per website, then IMHO we should leave that option untouched.

pedrobaeza:
If not, it should be properly documented

Yajo:
The tour is failing. You'll have to update some step probably.

yajo (changes requested):
The tour is failing. You'll have to update some step probably.

website_sale_require_legal/static/src/js/tour.js
```diff
@@ -8,7 +8,23 @@
 
     var steps = [
         {
-            trigger: "a:contains('iPod')",
+            content: "select conference chair",
```
yajo: :fire: You don't need the `content` key if it's a [tour-only test](https://github.com/OCA/e-commerce/pull/267/commits/c292c82a9fab4a6a9a95a44f2bf625b60433703f#diff-45d95ed91b5ba60838fc08079f91087aR65).

website_sale_require_legal/static/src/js/tour.js
```diff
@@ -8,7 +8,23 @@
...
+            trigger: '.oe_product_cart:first a:contains("Conference Chair")',
+        },
+        {
+            content: "select Conference Chair Aluminium",
```
yajo: :fire:

website_sale_require_legal/static/src/js/tour.js
```diff
@@ -8,7 +8,23 @@
...
+            extra_trigger: '#product_detail',
+            trigger: 'label:contains(Aluminium) input',
+        },
+        {
+            content: "select Conference Chair Steel",
```
yajo: :fire:

website_sale_require_legal/static/src/js/tour.js
```diff
@@ -8,7 +8,23 @@
...
+            extra_trigger: '#product_detail',
+            trigger: 'label:contains(Steel) input',
+        },
+        {
+            content: "click on add to cart",
```
yajo: :fire:

website_sale_require_legal/views/website_sale.xml
```diff
@@ -1,23 +1,25 @@
...
 
     <template id="accept_input">
-        <div t-attf-class="form-group col-md-12 #{'has-error' if error and error.get('accepted_legal_terms') else ''}">
-            <label for="accepted_legal_terms" class="control-label">
-                <input
-                    type="checkbox"
-                    name="accepted_legal_terms"
-                    id="accepted_legal_terms"
-                    required="required"/>
+        <div t-attf-class="form-group col-lg-12">
```
yajo: If this has no changes, there's no need to use `t-attf-`. Remove that part please.

yajo (approved):
Tests fixed

pedrobaeza:
I'm afraid not

yajo:
I restarted https://travis-ci.org/OCA/e-commerce/builds/437949375 because I suspect it's the main branch that's broken.

pedrobaeza:
Yes, main branch is also red. Please check and fix it.

yajo:
Fix in #281.

pedrobaeza:
Please rebase

yajo:
Done, let's see travis.

pedrobaeza (approved):

PR #269, opened 2019-03-04
[ADD] website_no_price_on_main

gfcapalbo:
Identify a website (or multiple websites)  as "main", all prices will not be shown on these website(s).

### DIFF
--- a/website_no_price_on_main/README.rst
+++ b/website_no_price_on_main/README.rst
@@ -1,0 +1,61 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+========================
+No Price on main website
+========================
+
+* Identify a website as "main", all prices will not be shown on this website.
+* The "main" websites will have another flow for carts, other than review > shipping & billing > payment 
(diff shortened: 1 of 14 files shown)


pedrobaeza:
And why not letting this be selected on `website` model?

gfcapalbo:
> And why not letting this be selected on `website` model?

website.config.settings seems naturally the place. obviously, is_main is defined on website.
in v9 website.config.settings is more accessible for the website manager than website.
seemed like a natural fit.

pedrobaeza:
But why it should be only on main? If I want to hide the prices in 2 websites, this module has all the base, but limit artificially the possibility.

gfcapalbo:
The module would work with multiple websites set as "is_main" . this module derived from a user request who needed 1 main website with no prices. The variable name "is_main" is not expressive of  that usecase, but it will work if you set 2-3 ..n websites as "is_main". the is_main flag is set per-website. so it will will work.

pedrobaeza:
But then call it "Hide prices" instead of "Is main"...

Any way, there's a module https://github.com/OCA/e-commerce/tree/11.0/website_sale_hide_price that performs the same. You should expand that one for being selectable at website level.

gfcapalbo:
@pedrobaeza  agreed with extension , closing MR. will do.

PR #270, opened 2019-03-04
[ADD] webshop_product_filter

gfcapalbo:
Add a column in webshop for advanced search on specific category - attributes. All products in that category will have these attributes to filter upon. Amazon-style.


pedrobaeza:
Please call it following core nomenclature: `website_sale_product_filter`

PR #275, opened 2019-03-12
[11.0][WIP] website_sale_secondary_unit: New module

sergio-teruel:
cc @Tecnativa


chienandalu (commented):
Add needed repos for dependecies :slightly_smiling_face:

chienandalu (commented):
Tested in runbot :+1: 

Anyway, your panthomjs tests seem to be failing in presence of another module of the report :slightly_frowning_face:

rafaelbn:
Tested functionally :+1: 

Please review travis in :red_circle:

rafaelbn (approved):
Functionally tested! :+1: 

Review please travis in :red_circle: Sergio

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
```diff
@@ -0,0 +1,40 @@
...
+                ],
+                'allow_uom_sell': True,
+                'website_published': True,
+            })
+
+    def test_ui_website(self):
+        """Test frontend tour."""
+        tour = (
+            "odoo.__DEBUG__.services['web_tour.tour']",
+            "website_sale_secondary_unit.tour",
```
yajo: You have to search the tour by [the name used to register it](https://github.com/OCA/e-commerce/pull/275/files#diff-33dc1ed7fbbae44004358753bd9f7874R50), not by the JS module name.

Change this for: `"website_sale_secondary_unit"`.

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+        self.phantom_js(
+            url_path="/",
+            code="%s.run('%s')" % tour,
+            ready="%s.tours['%s'].ready" % tour,
+            login="admin",
+            timeout=60)
```
yajo: Possibly the default timeout should be enough once you fix the above  points.

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
```diff
@@ -0,0 +1,40 @@
...
+        """Test frontend tour."""
+        tour = (
+            "odoo.__DEBUG__.services['web_tour.tour']",
+            "website_sale_secondary_unit.tour",
+        )
+        self.phantom_js(
+            url_path="/",
+            code="%s.run('%s')" % tour,
+            ready="%s.tours['%s'].ready" % tour,
+            login="admin",
```
yajo: I think you are supposed to be a public user instead, since you're buying in the e-commerce. If so, leave blank.

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
```diff
@@ -0,0 +1,40 @@
...
+                'website_published': True,
+            })
+
+    def test_ui_website(self):
+        """Test frontend tour."""
+        tour = (
+            "odoo.__DEBUG__.services['web_tour.tour']",
+            "website_sale_secondary_unit.tour",
+        )
+        self.phantom_js(
```
yajo, line 14: use `self.browser_js` instead (this is an alias only, I guess it's deprecated or will be in the future).

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2019 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests.common import HttpCase
+
+
+class WebsiteSaleHttpCase(HttpCase):
+    def setUp(self):
+        super(WebsiteSaleHttpCase, self).setUp()
+        with self.cursor() as cr:
+            env = self.env(cr)
```
yajo: Do not use this `with` trick anymore. In v12 it's not needed and in v10-11 it's broken now.

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
```diff
@@ -0,0 +1,40 @@
...
+                'secondary_uom_ids': [
+                    (0, 0, {
+                        'name': 'box',
+                        'uom_id': self.product_ipod.uom_id.id,
+                        'factor': 10,
+                    }),
+                    (0, 0, {
+                        'name': 'box',
+                        'uom_id': self.product_ipod.uom_id.id,
+                        'factor': 20,
```
yajo: You can load this as demo data instead if you wish. Tours are an exception to the "rule" of not using demo data in tests, because to design them you need to have clickable viewable things, and without demo data it's very hard to fix a broken tour because you cannot run it visibly in the browser.

yajo (changes requested):

website_sale_secondary_unit/models/product_secondary_unit.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2019 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import models, fields
+
+
+class ProductTemplate(models.Model):
```
martintorresvm: Think this should be the CamelCase transcription of `_name` so maybe call it `ProductSecondaryUnit`

website_sale_secondary_unit/models/sale_order.py
```diff
@@ -0,0 +1,97 @@
...
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def create(self, vals):
+        SecondaryUom = self.env['product.secondary.unit']
+        secondary_uom = SecondaryUom.browse(
+            vals.get('secondary_uom_id', False))
+        product_uom = self.env['product.uom'].browse(vals['product_uom'])
+        if secondary_uom:
+            factor = secondary_uom.factor * product_uom.factor
```
martintorresvm, line 78: It looks that from line 78 to 81 is the same as 92 to 95 (which is not very much) so if need change one place then you'll need change the other so better split on a function?

martintorresvm (changes requested):
About Travis not sure is can be fixed by add `travis_wait`

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
yajo: Oops, we're in v11; forget this.

website_sale_secondary_unit/models/sale_order.py
```diff
@@ -0,0 +1,97 @@
...
+        return super(SaleOrderLine, self).create(vals)
+
+    def write(self, vals):
+        SecondaryUom = self.env['product.secondary.unit']
+        for line in self:
+            secondary_uom = ('secondary_uom_id' in vals and
+                             SecondaryUom.browse(vals['secondary_uom_id']) or
+                             line.secondary_uom_id)
+            if 'product_uom_qty' in vals and secondary_uom:
+                factor = secondary_uom.factor * line.product_uom.factor
```
pedrobaeza:
```suggestion
                factor = secondary_uom.factor * vals.get('product_uom', line.product_uom.factor)
```

website_sale_secondary_unit/models/sale_order.py
pedrobaeza: Some code is shared, but the need of treating differently write and create is needed, so I think is not possible to extract it to a function without being more complex.

website_sale_secondary_unit/models/sale_order.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    def create(self, vals):
+        SecondaryUom = self.env['product.secondary.unit']
+        secondary_uom = SecondaryUom.browse(
+            vals.get('secondary_uom_id', False))
+        product_uom = self.env['product.uom'].browse(vals['product_uom'])
```
pedrobaeza, line 76: This line raises an error in conjunction with other module that has been fixed in https://github.com/OCA/delivery-carrier/commit/6b242987b29ec528994e25dca808bde96be892c9, but I wonder why `product_uom` wasn't needed in that case, as it seems to be required.

website_sale_secondary_unit/models/sale_order.py
martintorresvm: Agree with you

website_sale_secondary_unit/models/sale_order.py
pedrobaeza: Any thought about this?

website_sale_secondary_unit/models/product_secondary_unit.py
pedrobaeza: This is pending

website_sale_secondary_unit/models/sale_order.py
sergio-teruel: I don't know

sergio-teruel:
@victormmtorres @pedrobaeza changes done

pedrobaeza (approved):
Squashing everything on merge.

sergio-teruel:
Changes done

pedrobaeza:
@Yajo can you update your review?

chienandalu (approved):

yajo (approved):

PR #278, opened 2019-04-11
[FIX] website_sale_hide_price: compatible with website_multi_theme

rubenseijas:
When trying to install the module in a multi company and multi website system with the **website_multi_ theme** module installed it finds errors with the xpath of the templates because there is an xpath that does not inherit correctly

### DIFF
--- a/website_sale_hide_price/views/website_sale_template.xml
+++ b/website_sale_hide_price/views/website_sale_template.xml
@@ -16,20 +16,16 @@
         </template>
 
         <!-- Hide Add To Cart Button if not website_show_price -->
-        <template id="product_add" inherit_id="website_sale.product">
+        <template id="website_sale_product" inherit_id="website_sale.product" name="Product Hide Add to Cart">
             <xpath expr="//a[@id='add_to_cart']" position="attributes">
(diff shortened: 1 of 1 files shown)


rafaelbn:
@yelizariev could please take a look here?

yelizariev:
I guess, the error may happen even without multi-theme, when ``Select Quantity`` option is disabled. 
If I'm right, commit message could be updated

sergio-teruel:
This module can crash other modules tests from this repository because "hide" the button "Add to cart" and other modules run the tours with this button.
Also, crash website_sale form Odoo core, see here: https://github.com/odoo/odoo/blob/f6839b01d6bc5a89e859c9c2a05fc7cefbc2b215/addons/website_sale/static/src/js/website_sale_tour_buy.js#L44

pedrobaeza:
@sergio-teruel then what to do? Is there any mechanism that can prevent to hide the button if doing tours for other modules?

rubenseijas:
First at all, apologies by lazy answer. I just already done the last commit who must prevent tour crashes. Can someone check it, please?

PR #279, opened 2019-04-12
[11.0][IMP] website_sale_product_attribute_filter_visibility: Add hook to publish all existing attributes after install this module

sergio-teruel:
cc @Tecnativa


pedrobaeza:
Isn't simpler and better to just add `default=True` to the field?

```python
_inherit = [..., 'website.published.mixin']

website_published = fields.Boolean(default=True)
```

sergio-teruel:
I know!! but for all models wich inherit from website.published.mixin by default not is published in website, so for this model I don't want change this.
Ufff travis now fails due to database cursor.... :sob:

pedrobaeza:
You are changing it only for this model with that. The problem I see with this is that new attributes will be also non visible by default, which I don't consider appropriate for this case. It's legit to do this in the inherited model (nor for the general mixin).

sergio-teruel:
Ok. Changes done...

website_sale_product_attribute_filter_visibility/tests/test_attribute_filter_visibility.py
```diff
@@ -9,16 +9,18 @@ def setUp(self):
         super(WebsiteSaleHttpCase, self).setUp()
         with self.cursor() as cr:
             env = self.env(cr)
+            self.all_attributes = env['product.attribute'].search([])
+            self.all_attributes.write({'website_published': False})
```
pedrobaeza: Why unpublish all?

sergio-teruel: For not change the test

pedrobaeza: OK, but IMO, you should test both existence and not existence of one attribute, not only one side. Or are you doing that already (I have not followed full tour)?

sergio-teruel: in the next lines I create two attributes, one published and other unpublished

pedrobaeza: OK, got it! I saw the assignment line removed, but I was not counting with default value. Thanks for your patience.

pedrobaeza: Ey, wait, this is still applying... if you remove the line with `'website_published': True,`, and all attributes have been previously disabled, then no attribute is published!

website_sale_product_attribute_filter_visibility/tests/test_attribute_filter_visibility.py
```diff
@@ -9,16 +9,18 @@ def setUp(self):
...
                 'name': 'test_color',
                 'create_variant': False,
-                'website_published': True,
                 'type': 'radio',
             })
             self.attribute_size = env['product.attribute'].create({
                 'name': 'test_size',
                 'create_variant': False,
                 'type': 'radio',
+                'website_published': False,
```
pedrobaeza: This one is not needed according previous code.

website_sale_product_attribute_filter_visibility/tests/test_attribute_filter_visibility.py
```diff
@@ -9,16 +9,18 @@ def setUp(self):
         super(WebsiteSaleHttpCase, self).setUp()
         with self.cursor() as cr:
             env = self.env(cr)
+            self.all_attributes = env['product.attribute'].search([])
+            self.all_attributes.write({'website_published': False})
             self.attribute_color = env['product.attribute'].create({
                 'name': 'test_color',
                 'create_variant': False,
-                'website_published': True,
```
pedrobaeza, line 15: And this one is needed.

pedrobaeza: OK, not needed as you are creating.

website_sale_product_attribute_filter_visibility/tests/test_attribute_filter_visibility.py
pedrobaeza: But it's needed on creation, right.

website_sale_product_attribute_filter_visibility/tests/test_attribute_filter_visibility.py
pedrobaeza: OK, forget my comments. It's Friday afternoon and I'm telling nonsense.

pedrobaeza (approved):

sergio-teruel:
@pedrobaeza Ready for merge, I change test to js tour

yajo (approved):

PR #281, opened 2019-04-18
[FIX] website_sale_require_login: Fix public user test

yajo:
This PR should render the 12.0 branch :heavy_check_mark: in Travis again.

This test was expecting sign up to be disabled because `auth_signup` is not in its module graph, but since that addon is autoinstallable, it could be installed without our knowledge.

Just make sure signup is disabled before running the public user test, to make it unitary.

Also, moved to at-install mode, since the post-install one adds no value.

@Tecnativa

### DIFF
--- a/website_sale_require_login/tests/test_checkout.py
+++ b/website_sale_require_login/tests/test_checkout.py
@@ -2,8 +2,6 @@
 import odoo.tests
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Fast-tracking this

PR #285, opened 2019-05-10
12.0 mig website_sale_default_country

ernestotejeda:
* readme by fragments

Cc @Tecnativa


pedrobaeza:
Seeing this code, I consider this module as not needed:

https://github.com/odoo/odoo/blob/2a99a28004f8fcec96c73f847cbb6f3939fec4ab/addons/website_sale/controllers/main.py#L638-L646

pedrobaeza:
What do you think, @Tardo?

Tardo:
This module is usefull if you want have a country on all situations... without this module country_id can be "False" (Users without country_id defined)

pedrobaeza:
@Tardo not really, as you are using the public user associated to the website, so it's your duty to assign to that user the country.

chienandalu (commented):
I tested the flow setting a default country at the Public User in Odoo standard and I agree with @pedrobaeza that this seems to be uneeded from this version

ernestotejeda:
It's true, I tested the flow setting to default country at the Public User and it works fine. I even tried it locally in v11 and v10 and it works fine as well, so I think in v11 and v10 it's not necessary, either :) :)

Anyway, I'm going to test now how compatible this is with multiwebsite

ernestotejeda:
Yes, it works well for multi-website in odoo 12 with the public user. Looking a bit at the code, I have seen that each website is related to a user and that user is the user that is related to the company to which the website belongs

chienandalu:
I tested also with multicompany and is covered in standard as well

yajo:
Well, closing then as unneeded. It'd be good psosibly to have an OpenUpgrade migration script where this is merged into `website_sale`.

pedrobaeza:
You need to merge the module in OpenUpgrade.

PR #286, opened 2019-05-13
[12.0][MIG] website_sale_checkout_country_vat

Tardo:
Normal migration... Odoo 12.0 doesn't have "vat" as required field.


Tardo:
Runbot fails because missing dependencies.... (are in other repo and not merged into 12.0)

pedrobaeza (changes requested):
Trying on runbot, I find these problems:

- When not logged in, pressing "Sign up and checkout" leads to `404` page
- When logged in by other means, if I edit the address, I get `500: Internal Server Error`.

pedrobaeza:
About Travis tests, can be due to this module.

pedrobaeza:
I still get 404 page when I click in "Sign up and checkout". Is it due to this module? I have been able however to see the country dropdown when editing the address once logged-in.

Tardo:
Dropped oca_dependencies changes.

website_sale_checkout_country_vat/static/src/js/dropdown.js
```diff
@@ -1,17 +1,19 @@
...
-        selector: ".oe_website_sale:has(.js_country_dropdown, \
-                                        select[name=country_id])",
+    var CheckoutCountryVatDropdown = animation.Class.extend({
+        selector: ".oe_website_sale:has(.js_country_dropdown, " +
+                  "select[name=country_id])",
         start: function (editable_mode) {
             var result = this._super(editable_mode);
-            this.$address_country = this.$("select[name=country_id]");
-            this.$vat_no_country_field = this.$("#no_country_field");
+            this.$address_country = this.$target.find(
```
yajo: I'm curious: why are you removing the `this.$` shorthand? It's available in all subclasses of `Widget`...

yajo (changes requested):

website_sale_checkout_country_vat/static/src/js/dropdown.js
Tardo: Umm to ensure that doesn't have conflicts if the snippet it's used more than one time in the same page...

yajo: Don't worry, it doesn't happen.

If you were to use `$(selector)`, yes, it's raw jquery and selects in the whole page, but `this.$(selector)` is different and only searches inside the widget.

Tardo:
Reverted "$target" changes

Tardo:
@pedrobaeza The problem is that "website_sale_suggest_create_account" module  fails on runbot... I installed it on local and works :/

pedrobaeza:
Please check with @Yajo what can be the problem.

pedrobaeza:
Although I'm now seeing that this PR is green.

pedrobaeza (approved):

yajo (approved):

PR #290, opened 2019-05-21
[MIG] website_sale_show_company_data: Migration to 12.0

martintorresvm:
cc @Tecnativa


tbaden (approved):
thanks for the work 👍 
I did a code review: LGTM

chienandalu (commented):
Testing on runbot I got this weird behaviour:

vat-issue

pedrobaeza:
This seems something with website_sale_checkout_country_vat more than with this module. cc @Tardo

Tardo:
I think that conflicts come from here: https://github.com/odoo/odoo/blob/12.0/addons/website_sale/views/templates.xml#L1283
And others for use t-esc with empty values...

pedrobaeza:
So it's reproducible without the module?

Tardo:
@pedrobaeza Yes, without 'website_sale_checkout_country_vat' gets the same behavior

pedrobaeza:
OK, thanks for the confirmation. @victormmtorres then you should find another way to inject the fields for not conflicting that way.

martintorresvm:
@chienandalu Please have a check

chienandalu:
Ok, once runbot is up I'll check it. Did you get to find out if there was an Odoo core bug?

martintorresvm:
>Ok, once runbot is up I'll check it. Did you get to find out if there was an Odoo core bug?

You mean about the VAT no being filled from parent partner? Not sure if it's reported but for sure I could not see this field propagated as soon you edited the parent partner

martintorresvm:
@chienandalu https://github.com/odoo/odoo/pull/33695

martintorresvm:
@chienandalu please have a test when have time

website_sale_show_company_data/readme/CONTRIBUTORS.rst
```diff
@@ -1 +1,3 @@
-* David Vidal <david.vidal@tecnativa.com>
+* `Tecnativa <https://www.tecnativa.com>`_:
+  * David Vidal
```
chienandalu, line 3: You need a blank line in between or it won't be properly formated (you can make a quick test with http://rst.ninjs.org)

chienandalu (approved):
Looks ok now. The other bug is on Odoo's roof now. Just a tiny thing:

martintorresvm:
Changes done @chienandalu

PR #300, opened 2019-07-08
[ADD] Adds website_sale_payment_term module

daramousk:


lfreeke:
@NL66278 can you have a look at this?

website_sale_payment_term/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,8 @@
+
+Contributors
+------------
+
+* George Daramouskas <gdaramouskas@therp.nl>
```
NL66278, line 1: This is the only line needed in this file

website_sale_payment_term/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+Configuration
+=============
+
+No configuration can take place
```
NL66278: Only this line needed. Just say: no specific configuration is needed for this module.

website_sale_payment_term/readme/CREDITS.rst
```diff
@@ -0,0 +1,9 @@
+
```
NL66278: Leave out this file altogether. There are np special credits.

website_sale_payment_term/readme/INSTALL.rst
```diff
@@ -0,0 +1,6 @@
+Installation
+============
+
+To install this module, you need to:
+
+Follow the usual installation procedure for your version.
```
NL66278, line 1: Limit to this line.

website_sale_payment_term/readme/ROADMAP.rst
```diff
@@ -0,0 +1,5 @@
+
```
NL66278: Leave out this file altogether

website_sale_payment_term/readme/USAGE.rst
```diff
@@ -0,0 +1,12 @@
+Usage
+=====
+
+To use this module, you need to simply install it.
```
NL66278, line 1: This and the next three lines suffice

website_sale_payment_term/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>
```
NL66278, line 2: No more copyright in __init__.py files

website_sale_payment_term/models/sale_order.py
```diff
@@ -0,0 +1,26 @@
...
+    def write(self, vals):
+        vals = self._set_payment_term_id(vals)
+        return super(SaleOrder, self).write(vals)
+
+    def _set_payment_term_id(self, vals):
+        payment_acquirer_id = vals.get('payment_acquirer_id')
+        if not vals.get('payment_term_id') and payment_acquirer_id:
+            payment_acquirer_id = self.env['payment.acquirer'].browse(
+                payment_acquirer_id)
+            vals['payment_term_id'] = payment_acquirer_id.payment_term_id.id
```
NL66278: What if it is not set on the payment_acquirer? Then still should default to that of partner like usual...

NL66278 (changes requested):
Changes needed. Also look at red tests.

daramousk:
@NL66278 Not really sure about those failures, I am not sure how all these `wkhtmltopdf` missing and the phantom js errors can be related to this PR.

website_sale_payment_term/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+{
+    "name": "website_sale_payment_term",
+    "version": "10.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Website",
```
NL66278, line 9: Missing the 'website' key (value should be: https://github.com/OCA/e-commerce).

website_sale_payment_term/models/sale_order.py
```diff
@@ -0,0 +1,26 @@
...
+        return super(SaleOrder, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        vals = self._set_payment_term_id(vals)
+        return super(SaleOrder, self).write(vals)
+
+    def _set_payment_term_id(self, vals):
+        payment_acquirer_id = vals.get('payment_acquirer_id')
+        payment_term_id = self.env['payment.acquirer'].browse(
```
NL66278, line 22: Will fail if there is no payment_acquirer_id in vals.

Just start with:
```
if vals.get('payment_term_id') or not vals.get('payment_acquirer_id'):
    return vals
```
Then you know you have a payment acquirer and are not overwriting an explicitly requested payment_term_id.

NL66278 (changes requested):
See comment

website_sale_payment_term/models/sale_order.py
daramousk: I am not sure how this will fail.

If there is no `payment_acquirer_id` in vals then that whole `browse(None).field` will evaluate to `False`. And then the condition will evaluate to `False` again.

NL66278: True it does not fail, did not expect browse(False) to just return an empty record set. Still an unneeded and easily avoidable function call...

PR #325, opened 2020-01-13
[MIG] website_sale_secondary_unit: Migration script to v12.0

sergio-teruel:
cc @Tecnativa
TT21362
@Tardo @chienandalu Review please


pedrobaeza (approved):
I'm fast-tracking this because it's better to mitigate this bug ASAP

PR #326, opened 2020-01-13
[FIX] website_sale_product_minimal_price: combination does not exist message error

sergio-teruel:
cc @Tecnativa
TT21362

### DIFF
--- a/website_sale_product_minimal_price/__manifest__.py
+++ b/website_sale_product_minimal_price/__manifest__.py
@@ -3,7 +3,7 @@
 {
     'name': 'Website Sale Product Minimal Price',
     'summary': 'Display minimal price for products that has variants',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
     'development_status': 'Production/Stable',
     'maintainers': ['sergio-teruel'],
     'category': 'Website',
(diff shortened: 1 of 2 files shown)


website_sale_product_minimal_price/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     'name': 'Website Sale Product Minimal Price',
     'summary': 'Display minimal price for products that has variants',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
```
pedrobaeza, line 6: Remember that this is done by the bot.

pedrobaeza (approved):

website_sale_product_minimal_price/__manifest__.py
sergio-teruel: Ok, But should i delete it?

pedrobaeza: Well, not this time, unless you plan to put other PR over the same module.

sergio-teruel: ok

Tardo (approved):
:+1: LGTM

PR #330, opened 2020-01-27
[12.0][ADD] add website_sale_product_display_mode

foutoucour:
The module adds the field display mode on product template, in e-commerce tab.

display_mode


flotho:
Hi,

Sorry for this probably stupid question but what's the use of the new field you introduced there?

Regards

ValentinVinagre:
Hi @foutoucour ,
I agree with @flotho , I don't see any changes in Odoo, only one field was added in eCommerce tab, but this field doesn't change anything, commits contain all the code?

PR #333, opened 2020-01-30
[12.0][FIX] website_sale_checkout_skip_payment: Change error template reference

Tardo:
In some point the module was changed their name, but the reference to the error template not. This commit improves the error template (adding a padding) and fixes the reference problem.

cc @tecnativa


sergio-teruel (approved):

MeetKD (approved):
Can you please update the module version ?

PR #343, opened 2020-03-25
new module abandoned_carts to delete abandoned carts, orders, contacts

jans23:
This module allows to delete Quotations and related partner records when those are older than a definable retention period.

Features:
* Configure a retention period of x hours (default: 48 hours)
* Specify a maximum of abandoned items to be deleted at a time (default: 2000 items) in order to avoid unresponsive server
* A log for reviewing purposes


rousseldenis:
@jans23 Have you ever seen the sale_automatic_workflow module ?

https://github.com/OCA/sale-workflow/tree/12.0/sale_automatic_workflow

This one could be extended to delete such orders (as you can define a domain to select orders and then do an action on them).

In another hand, you can have a look at https://github.com/shopinvader/odoo-shopinvader/tree/12.0/shopinvader_cart_expiry/

That if you manage e-commerce platform. Maybe you could switch to shopinvader :smiley:

PR #346, opened 2020-04-05
[12.0] FIX website_sale_product_brand tanslation error on update with --i18n-overwrite

eLBati:
Steps:

 - Install website_sale_product_brand
 - Install Italian language
 - Update with `-u  website_sale_product_brand  --i18n-overwrite`

Get

```
bad query:  INSERT INTO ir_translation(name, lang, res_id, src, type, value, module, state, comments)
                           SELECT name, lang, res_id, src, type, value, module, state, comments
                           FROM tmp_ir_translation_import
                           WHERE type = 'model'
                           AND noupdate IS NOT TRUE
                           ON CONFLICT (type, lang, name, res_id) WHERE type = 'model'
                            DO UPDATE SET (name, lang, res_id, src, type, value, module, state, comments) = (EXCLUDED.name, EXCLUDED.lang, EXCLUDED.res_id, EXCLUDED.src, EXCLUDED.type, EXCLUDED.value, EXCLUDED.module, EXCLUDED.state, EXCLUDED.comments)
                            WHERE EXCLUDED.value IS NOT NULL AND EXCLUDED.value != '';
                       
ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time
HINT:  Ensure that no rows proposed for insertion within the same command have duplicate constrained values
```

### DIFF
--- a/website_sale_product_brand/i18n/it.po
+++ b/website_sale_product_brand/i18n/it.po
@@ -1,42 +1,37 @@
 # Translation of Odoo Server.
 # This file contains the translation of the following modules:
-# * website_sale_product_brand
+#	* website_sale_product_brand
 #
-# Translators:
-# OCA Transbot <transbot@odoo-community.org>, 2016
-# Paolo Valier <paolo.valier@hotmail.it>, 2017
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 9.0c\n"
+"Project-Id-Version: Odoo Server 12.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-03-28 02:55+0000\n"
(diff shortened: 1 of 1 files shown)


eLBati:
@sbidoul I don't know if this can be fixed in weblate directly though

sbidoul (changes requested):
Actually the `.pot` is bad in the first place. And the reason for that is there is no test job that has MAKEPOT=1 in .travis.yml. Fixing that will have travis regenerate the .pot, which weblate should pickup and msgmerge it in the .po files.

pedrobaeza:
I have added MAKEPOT in e8376801d978ff017ab9e58a7eee6fa767d21c2d, so it should be fixed when Travis finishes.

eLBati:
Thanks!

PR #350, opened 2020-04-28
[ADD] website_sale_exception: new module

zaoral:
This module allows to pass a website sale from draft state to sent if there are exeptions that do not let to continue to done state. This is required to close the website sale order after the sale order has been confirmed and the payment has been receive. This way the user can create a new sale order from website without overwritten the last one and without losing the payment information.


jjscarafia (approved):
:+1: technical and functional review

jjscarafia:
Anyone using sale_exception consider this interesting? 
@ivantodorovich @rvalyi @pedrobaeza @florian-dacosta @sebalix @simahawk 
Thanks!

website_sale_exception/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020 ADHOC
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Website Sale Exception',
+    'summary': 'support for sale exceptions on website',
+    'version': '12.0.1.1.0',
```
pedrobaeza:
```suggestion
    'version': '12.0.1.0.0',
```

website_sale_exception/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows to pass a website sale from draft state to sent if there are exeptions that do not let to continue to done state. This is required to close the website sale order after the sale order has been confirmed and the payment has been receive. This way the user can create a new sale order from website without overwritten the last one and without losing the payment information.
```
ivantodorovich:
```suggestion
This module allows to pass a website sale from draft state to sent if there are exeptions that do not let to continue to done state. This is required to close the website sale order after the sale order has been confirmed and the payment has been received. This way the user can create a new sale order from the website without overwriting the last one and without losing the payment information.
```

website_sale_exception/__manifest__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020 ADHOC
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Website Sale Exception',
+    'summary': 'support for sale exceptions on website',
```
ivantodorovich:
```suggestion
    'summary': 'Support for sale exceptions on website',
```

ivantodorovich (approved):
Minor comment, LGTM :+1:

zaoral:
thanks for the review @pedrobaeza @ivantodorovich

pedrobaeza:
Can you please squash commits?

zaoral:
Hi @pedrobaeza 
Ready, the commits has been squashed

website_sale_exception/models/sale.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = 'sale.order'
+
+    @api.multi
+    def action_confirm(self):
+        if self.detect_exceptions():
```
pedrobaeza, line 11: Isn't this modifying also the behavior of regular sales orders?

ivantodorovich: it's filtering on orders with website_id right after that, so I would say it's not

pedrobaeza: OK, I see, but shouldn't this be done after calling super then?

zaoral: I did it for some reason but at this moment do not remember, I will try to do it after the super() and test it. Let you know the result

zaoral: @pedrobaeza 

I have check the behavior: If we do the super first, then the website sale orders that has detect exceptions will be pass to sent state but they will not reflect the exception info in the sale order, we will never be inform that an exception occurs.

image

The way we are doing it now let us to detect exceptions, pass those sale orders to sent state, and then run the normal exception workflow so this way we will have a sent sale order but we will have the info about the exception that occurs trying to validate the order.

image

pedrobaeza: OK, thanks for the explanations!

pedrobaeza (approved):

PR #351, opened 2020-05-04
[12.0][FIX] website_sale_checkout_country_vat: inherit the correct view

yajo:
This module happened to be working accidentally because [it was inheriting from a view that had another inheriting view which defined the necessary elements][1]; but that view can be disabled and then this module breaks.

Fixed by inheriting from the correct view.

Also benefits from OCA/website#720 by using readonly mode if that module is updated.

@Tecnativa TT17694

[1]: https://github.com/odoo/odoo/blob/6e99a89574d11811c03fe3cdc61922352a9610ea/addons/website_sale/views/templates.xml#L1296

### DIFF
--- a/website_sale_checkout_country_vat/views/templates.xml
+++ b/website_sale_checkout_country_vat/views/templates.xml
@@ -5,7 +5,7 @@
      License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl-3). -->
 <odoo>
 
-<template id="address" inherit_id="website_sale.address">
+<template id="address" inherit_id="website_sale.address_b2b">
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

Tardo (approved):

pedrobaeza:
@Yajo can we merge this?

yajo:
We could, but this fix is incomplete. :man_shrugging:

pedrobaeza:
OK, then please finish it

yajo:
Finished.

yajo:
We need https://github.com/OCA/website/pull/720 to fully benefit from this patch.

PR #354, opened 2020-05-07
[12.0][IMP] website_sale_stock_available_display: Allow overwrite invisible attrs in view to avoid crash with other modules

sergio-teruel:
cc @Tecnativa
TT23780
ping @chienandalu @Tardo

### DIFF
--- a/website_sale_stock_available_display/models/product_template.py
+++ b/website_sale_stock_available_display/models/product_template.py
@@ -1,6 +1,5 @@
 # Copyright 2019 Tecnativa - Sergio Teruel
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
-
 from odoo import fields, models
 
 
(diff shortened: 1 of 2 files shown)


website_sale_stock_available_display/models/product_template.py
```diff
@@ -11,3 +13,36 @@ class ProductTemplate(models.Model):
...
+        res = super().fields_view_get(
+            view_id=view_id, view_type=view_type, toolbar=toolbar,
+            submenu=submenu,
+        )
+        if view_type == 'form':
+            view_xml = etree.XML(res['arch'])
+            xml_field = view_xml.xpath("//field[@name='custom_message']")
+            if xml_field:
+                field = xml_field[0]
+                attribs = safe_eval(field.attrib.get("attrs", "{}"))
```
pedrobaeza: Don't `safe_eval` this as it can contain unevaluable things (like a field name). You have to inject it through strings:

https://github.com/OCA/commission/blob/05a05f3060c2085f014ec231296e51659e675b59/sale_commission/models/sale_order.py#L28-L48

pedrobaeza (changes requested):

pedrobaeza (approved):

chienandalu (approved):

PR #355, opened 2020-05-12
[12.0][FIX] website_sale_stock_available: immediately_usable_qty in_cart_update

ernestotejeda:
Fix: take into account here (see image)  'Available to promise' instead of 'Forecast Quantity' To restrict add to cart

website_fix

Cc @Tecnativa TT21547

### DIFF
--- a/website_sale_stock_available/models/__init__.py
+++ b/website_sale_stock_available/models/__init__.py
@@ -1,3 +1,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+from . import product_product
 from . import product_template
+from . import sale_order
--- a/website_sale_stock_available/models/product_product.py
+++ b/website_sale_stock_available/models/product_product.py
@@ -1,0 +1,19 @@
+# Copyright 2020 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class Product(models.Model):
+    _inherit = 'product.product'
+
+    def _compute_quantities_dict(self, lot_id, owner_id, package_id,
+                                 from_date=False, to_date=False):
+        res = super()._compute_quantities_dict(
+            lot_id, owner_id, package_id, from_date, to_date)
+        if self.env.context.get('website_sale_stock_available'):
+            for product in self.with_context(
+                    website_sale_stock_available=False):
+                immediately = product.immediately_usable_qty
+                res[product.id]['virtual_available'] = immediately
+        return res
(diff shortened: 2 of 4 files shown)


chienandalu (commented):
Should the "Add to cart" button in the category view be disabled the same way the same way it is in the product view?

pedrobaeza:
> Should the "Add to cart" button in the category view be disabled the same way the same way it is in the product view?

That's not possible due to performance and Odoo standard doesn't do it. Only when you try to add it, you receive the warning.

chienandalu (approved):
Then it's ok

website_sale_stock_available/models/product.py
```diff
@@ -0,0 +1,33 @@
...
+    def _get_combination_info(self, combination=False, product_id=False,
+                              add_qty=1, pricelist=False,
+                              parent_combination=False, only_template=False):
+        template = self.with_context(website_sale_stock_available=True)
+        return super(ProductTemplate, template)._get_combination_info(
+            combination, product_id, add_qty, pricelist, parent_combination,
+            only_template)
+
+
+class Product(models.Model):
```
pedrobaeza: Don't join both models in the same file. I prefer 2 separate files.

website_sale_stock_available/models/product.py
```diff
@@ -0,0 +1,33 @@
...
+    def _compute_quantities_dict(self, lot_id, owner_id, package_id,
+                                 from_date=False, to_date=False):
+        res = super()._compute_quantities_dict(
+            lot_id, owner_id, package_id, from_date, to_date)
+        if 'website_sale_stock_available' in self.env.context:
+            new_dict = dict(self.env.context)
+            new_dict.pop("website_sale_stock_available", None)
+            for product in self.with_context(new_dict):
+                immediately = product.immediately_usable_qty
+                res[product.id]['virtual_available'] = immediately
```
yajo: This should be the same:

```suggestion
        for product in self.with_context(website_sale_stock_available=False):
            immediately = product.immediately_usable_qty
            res[product.id]['virtual_available'] = immediately
```

yajo (changes requested):

ernestotejeda:
Changes done. @pedrobaeza @Yajo

website_sale_stock_available/models/product.py
ernestotejeda: I passed website_sale_stock_available=False to context instead of removing it from context, as you stated, but I didn't remove the 'if' statement. I don't want that code block inside the 'if' to be executed if the context doesn't have that variable or it is False.

yajo (approved):

sergio-teruel (approved):
Tested in customer environment

PR #356, opened 2020-05-13
[ADD][12.0] website_sale_product_sort: New module

chienandalu:
This module allows to set a default sorting criteria for the e-commerce product
list.

It provides the standard criterias, but allows to extend the options via third
modules.

cc @Tecnativa TT20974


website_sale_product_sort/models/website.py
```diff
@@ -0,0 +1,26 @@
...
+            ('list_price asc', _('Catalog price: Low to High')),
+            ('name asc', _('Name - A to Z')),
+            ('name desc', _('Name - Z to A')),
+        ]
+
+    default_product_sort_criteria = fields.Selection(
+        selection='_get_product_sort_criterias',
+        string="Sort products by",
+        help="Default criteria for sorting products in the shop",
+        default='website_sequence asc',
```
sergio-teruel: This default value not exists, so this module can not be installed

sergio-teruel (changes requested):

website_sale_product_sort/models/website.py
chienandalu: Sure. I changed the options in the last force push and forgot about the default. Thanks

sergio-teruel (approved):

rafaelbn (approved):
:+1:

ValentinVinagre:
The `'development_status': 'Beta'` it's correct?

pedrobaeza:
Well, it can fit stable level:

https://odoo-community.org/page/module-maturity-levels

but we can let it as is if we want.

ValentinVinagre:
> Well, it can fit stable level:
> 
> https://odoo-community.org/page/module-maturity-levels
> 
> but we can let it as is if we want.

okay.

PR #357, opened 2020-05-15
[ADD] website_sale_category_description

yajo:
With this module, you'll be able to add a description individually for each of your website product categories.

@Tecnativa TT23930


website_sale_category_description/models/product_public_category.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2020 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class ProductPublicCategory(models.Model):
+    _inherit = "product.public.category"
+
+    website_description_1 = fields.Html(
```
pedrobaeza: Why not `website_description_top` (and `_bottom`)?

website_sale_category_description/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Jairo Llopis <jairo.llopis@tecnativa.com> (`Tecnativa <https://www.tecnativa.com/>`__)
```
pedrobaeza: We use to put company and indented, the person (without the email for avoiding "spam").

website_sale_category_description/models/product_public_category.py
yajo: I thought it but felt like not anchoring the field to a place would leave more room for theme designs, if some wanted to place these different.

But I'm not reluctant to name them like that if you think it's better.

pedrobaeza: Well, their intention is clear, and designs can vary the exact place on top or on bottom, but to misuse them for putting for example both on top, or one in the middle, has no sense on the context of this module. If they do that, it can be still with the same fields. On the contrary, having `_1` and `_2`, you can have the doubt of which one represents the top part.

chienandalu (approved):
Cool :+1:

website_sale_category_description/README.rst
```diff
@@ -0,0 +1,105 @@
...
+=====
+
+To use this module, you need to:
+
+#. Go to `your eCommerce </shop>`__.
+#. Click on *Customize*.
+#. Enable *eCommerce Categories* if it was disabled.
+#. Navigate to any category.
+#. Click on *Edit*.
+#. You will see 2 block drop areas at the top, and 2 on the bottom:
```
rafaelbn: I don't see 4 block drom areas, I see 2

website_sale_category_description/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+   * 1st top block is a description that will appear in all your eCommerce pages.
```
rafaelbn: I have tested and this block doesn't exists

website_sale_category_description/README.rst
```diff
@@ -0,0 +1,105 @@
...
+   * 2nd top block is the same, but specific for this category.
```
rafaelbn: :+1:

website_sale_category_description/README.rst
```diff
@@ -0,0 +1,105 @@
...
+   * 1st bottom block is also a specific description for this category.
```
rafaelbn: :+1:

website_sale_category_description/README.rst
```diff
@@ -0,0 +1,105 @@
...
+   * 2st bottom block is another description that will appear in all your eCommerce pages.
```
rafaelbn: I have tested and this block doesn't exists

rafaelbn (dismissed):
Functional reviewed, check my comments please.

yajo:
@pedrobaeza @rafaelbn all fixed.

yajo:
Merge?

rafaelbn:
@Yajo I'm not able to save any kind of block in the botton, see this screenshot:

yajo:
@rafaelbn That issue is a bug in odoo framework fixed in https://github.com/odoo/odoo/pull/51885 that was affecting this module in https://github.com/OCA/e-commerce/commit/3fd89a2502480526be1f24df3c1bbe6ad033c07d.

I force-pushed it to reduce the feature scope and make it match what odoo v13 upstream does. Now it's not affected by that bug, as a side effect.

Here's how it looks like now:

pedrobaeza (approved):

PR #359, opened 2020-05-25
[13.0] [MIG] website_sale_attribute_filter_category: Migration to v13

CarlosRoca13:
cc @Tecnativa TT24057


website_sale_attribute_filter_category/__manifest__.py
```diff
@@ -3,8 +3,7 @@
 {
     "name": "Website Sale Attribute Filter Category",
     "summary": "Allow group attributes in shop by categories",
-    "version": "12.0.1.0.0",
-    "development_status": "Beta",
```
pedrobaeza: Should this be elevated to `Production/Stable`?

website_sale_attribute_filter_category/tests/test_website_sale_attribute_filter_category.py
```diff
@@ -15,7 +15,6 @@ def setUp(self):
         self.product_attribute = ProductAttribute.create(
             {
                 "name": "Test",
-                "website_published": True,
```
pedrobaeza: Why not published?

CarlosRoca13: The field is_published has the default value True

pedrobaeza: In this case I prefer to be `explicit better than implicit`. If your test depends on this value, don't rely on default values, that other modules or versions can change it.

pedrobaeza (approved):
Code review

website_sale_attribute_filter_category/tests/test_website_sale_attribute_filter_category.py
CarlosRoca13: Ok, so I have to see it better, because this field is giving me problems on test...

sergio-teruel:
@CarlosRoca13 @chienandalu I ported commit form v12.0... 
https://github.com/OCA/e-commerce/pull/387

sergio-teruel (approved):

PR #360, opened 2020-05-25
[12.0][ADD] website_snippet_product_category

Tardo:
Adds a new snippet to show the e-commerce product categories.

categories

cc @tecnativa TT20965


website_snippet_product_category/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Usage
+=====
+
+#. Go to website (frontend)
+#. Press 'Edit' button
+#. Drag the 'Product Category' snippet to the place of the website that you want
+#. Save changes
+
+You can select what categories can be shown by the snippet and an image for the category:
```
pedrobaeza: This should go to `CONFIGURE.rst`.

website_snippet_product_category/README.rst
```diff
@@ -0,0 +1,105 @@
...
+#. Go to website (backend) > Configuration > Products > eCommerce Categories
+#. Create or Edit one
+#. You can see two new options "Website Published" and "Category Image"
+
+Known issues / Roadmap
+======================
+
+The snippet shows a maximum of 4 levels by design.
```
pedrobaeza: This is a point also

website_snippet_product_category/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+#. Adds a option to configure the initial level
```
pedrobaeza: Don't use numbered list here. Use `*`

website_snippet_product_category/models/product_public_category.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2020 Tecnativa - Alexandre Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProductPublicCategory(models.Model):
+    _inherit = ["product.public.category"]
+
+    website_published = fields.Boolean("Website Published", default=True)
```
pedrobaeza: Please use `website.published.mixin` instead

pedrobaeza (commented):
Please also indicate in the README in a section inside `USAGE.rst` with `~~~~` heading, the relevant classes for theming this snippet.

Tardo:
Changes done

website_snippet_product_category/models/product_public_category.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2020 Tecnativa - Alexandre Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProductPublicCategory(models.Model):
+    _inherit = ["product.public.category", "website.published.mixin"]
```
pedrobaeza, line 7: When inheriting from multiple models, you have to explicitly put `_name`
```suggestion
    _inherit = ["product.public.category", "website.published.mixin"]
    _name = "product.public.category"
```

Tardo:
omg... fixed.

website_snippet_product_category/README.rst
```diff
@@ -0,0 +1,124 @@
...
+#. Go to website (frontend)
+#. Press 'Edit' button
+#. Drag the 'Product Category' snippet to the place of the website that you want
+#. Save changes
+
+
+Theming
+~~~~~~~
+
+* `.categ_container` > The container per master category
```
pedrobaeza: Say that these are `CSS classes`

website_snippet_product_category/README.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Also you can replace the entire template replacing the 'data-template' attribute value 
+in the `s_product_category_options` view. You only need know that the name 
+of the variable that have all the categories is called `object`.
+
+Known issues / Roadmap
+======================
+
+* The snippet shows a maximum of 4 levels by design.
+* Adds a option to configure the initial level
```
pedrobaeza:
```suggestion
* Add an option to configure the initial level.
```

website_snippet_product_category/README.rst
```diff
@@ -0,0 +1,124 @@
...
+* Adds a option to select how many levels can be represented
```
pedrobaeza:
```suggestion
* Add an option to select how many levels will be shown.
```

website_snippet_product_category/README.rst
```diff
@@ -0,0 +1,124 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+You can select what categories can be shown by the snippet and an image for the category:
+#. Go to website (backend) > Configuration > Products > eCommerce Categories
```
pedrobaeza, line 40: You have to put an empty line here for proper rendering

website_snippet_product_category/static/src/js/frontend.js
```diff
@@ -0,0 +1,81 @@
...
+// License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+odoo.define("website_snippet_product_category.s_product_category", function (require) {
+    "use strict";
+
+    var core = require("web.core");
+    var sAnimation = require("website.content.snippets.animation");
+
+    var _t = core._t;
+
+    sAnimation.registry.js_product_category = sAnimation.Class.extend({
```
pedrobaeza, line 12: Is this for lazy/asynchronous rendering? If so, please indicate it as comment in the corresponding method. Remember, always docstring of the why, not the what (of the code).

website_snippet_product_category/views/product_public_category.xml
```diff
@@ -0,0 +1,23 @@
...
+                name="inherit_id"
+                ref="website_sale.product_public_category_form_view"
+            />
+            <field name="arch" type="xml">
+                <field name="sequence" position="after">
+                    <field name="website_published" />
+                    <field
+                        name="website_image"
+                        widget="image"
+                        options="{'size': [32,32]}"
```
pedrobaeza: Is this limiting the uploaded image to 32x32?

website_snippet_product_category/views/assets.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
pedrobaeza: Shouldn't this go `/templates/` instead?

website_snippet_product_category/views/product_public_category.xml
```diff
@@ -0,0 +1,23 @@
...
+        <record id="product_public_category_form_view" model="ir.ui.view">
+            <field name="name">public.category.form</field>
+            <field name="model">product.public.category</field>
+            <field
+                name="inherit_id"
+                ref="website_sale.product_public_category_form_view"
+            />
+            <field name="arch" type="xml">
+                <field name="sequence" position="after">
+                    <field name="website_published" />
```
pedrobaeza: Why not putting this as smart-button? Any way, the default label for the field is not very descriptive for this use case, so I think you need to override that string.

pedrobaeza (changes requested):
Functional and code review. Additional to the inline comments, you have to:

- Add demo data for putting some demo website categories as visible in the website, and put also an image in one of them. The data set you put on the screenshot was great.
- Add demo page with the snippet inserted (is that easily done)? I think you have other examples like country dropdown demo.

website_snippet_product_category/controllers/website.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+    def render_product_category(self, template, **kwargs):
+        categories = request.env["product.public.category"].search(
+            [("parent_id", "=", False), ("website_published", "=", True)],
+            order="child_id,sequence DESC",
+        )
+        keep = QueryURL("/shop", category=0)
+        return request.website.viewref(template).render(
+            {"object": categories, "keep": keep}
+        )
```
yajo: This `viewref` is interesting indeed. I didn't know of that method.

However, digging a little bit, it's better to stick with the good old `request.render`, which renders views lazily by default (thus making it more extendable and performant):

```suggestion
        return request.render(
            template,
            {"object": categories, "keep": keep}
        )
```

website_snippet_product_category/readme/ROADMAP.rst
```diff
@@ -0,0 +1,3 @@
+* The snippet shows a maximum of 4 levels by design.
+* Adds a option to configure the initial level
+* Adds a option to select how many levels can be represented
```
yajo:
```suggestion
* Add a option to configure the initial level
* Add a option to select how many levels can be represented
```

website_snippet_product_category/readme/USAGE.rst
```diff
@@ -0,0 +1,21 @@
...
+
+
+Theming
+~~~~~~~
+
+* `.categ_container` > The container per master category
+* `.categ_tree_level` > The category row (image + text)
+* `.main_tree_level` > The first category row level displayed
+* `.categ_link` > The link of the category row
+* `.categ_img` > The container of the icon of the category row
```
yajo: Remember this is RST, not MD. Single-backtick is a link; double is code:

```suggestion
* `.categ_container` > The container per master category
* ``.categ_tree_level`` > The category row (image + text)
* ``.main_tree_level`` > The first category row level displayed
* ``.categ_link`` > The link of the category row
* ``.categ_img`` > The container of the icon of the category row
```

BTW thanks for documenting this! :raised_hands:

website_snippet_product_category/readme/USAGE.rst
```diff
@@ -0,0 +1,21 @@
...
+
+You can use 'data-tree-level' attribute to select the category row of a selected
```
yajo:
```suggestion
You can use ``data-tree-level`` attribute to select the category row of a selected 
```

website_snippet_product_category/readme/USAGE.rst
```diff
@@ -0,0 +1,21 @@
...
+* `.main_tree_level` > The first category row level displayed
+* `.categ_link` > The link of the category row
+* `.categ_img` > The container of the icon of the category row
+
+You can use 'data-tree-level' attribute to select the category row of a selected 
+level (1-4 by design).
+
+Also you can replace the entire template replacing the 'data-template' attribute value 
+in the `s_product_category_options` view. You only need know that the name 
+of the variable that have all the categories is called `object`.
```
yajo:
```suggestion
Also you can replace the entire template replacing the ``data-template`` attribute value 
in the ``s_product_category_options`` view. You only need know that the name 
of the variable that have all the categories is called ``object``.
```

website_snippet_product_category/static/src/js/frontend.js
yajo, line 12: If I'm not wrong, the correct class to extend is `web_editor.snippets.options.SnippetOption`, which includes a `cleanForSave()` method that should empty this widget before hitting "Save". It would solve this problem:

1. User edits website and drops this snippet.
1. It downloads current categories.
1. User saves.
1. Those categories are saved as part of the page.
1. eCommerce categories change, but this page has a hardcoded collection in this snippet.
1. A visitor browses this page. He sees briefly the old categories before the rpc call is completed.

This would possibly imply a refactor that maybe is not worth the effort, so let's leave this as it is. I'd label this module as Alpha category though.

website_snippet_product_category/static/src/scss/snippet.scss
```diff
@@ -0,0 +1,42 @@
...
+    border-width: 0 1px 0 0;
+
+    .categ_tree_level {
+        &.main_tree_level {
+            color: $primary;
+            text-transform: uppercase;
+        }
+
+        &[data-tree-level="2"] {
+            font-size: 0.9em;
```
yajo, line 16: :no_entry_sign: Always use `rem` units: https://getbootstrap.com/docs/4.0/migration/#global-changes

website_snippet_product_category/static/src/scss/snippet.scss
```diff
@@ -0,0 +1,42 @@
...
+            .categ_link {
+                text-transform: capitalize;
+                color: $gray-800;
+            }
+            .categ_img img {
+                width: 80%;
```
yajo: Would it make more sense to use BS classes such as `col-10` and so on? (the same for all `width` declarations)

website_snippet_product_category/views/snippets.xml
```diff
@@ -0,0 +1,81 @@
...
+                    <div
+                        t-if="category.website_image"
+                        t-field="category.website_image"
+                        t-attf-class="categ_img d-inline-block pr-1"
+                        t-options="{'widget': 'image', 'class': 'img img-responsive rounded','resize': '24x24' }"
+                    />
+                    <a
+                        t-att-href="keep('/shop/category/' + slug(category), category=0)"
+                        class="categ_link"
+                        t-field="category.name"
```
yajo: This does not work as a normal t-field and you cannot edit category names from the widget.

I'm not sure if that's expected or just an accident. It's not like a big deal.

yajo (changes requested):
Am I supposed to be able to do this? :dizzy_face: 

Peek 27-05-2020 09-46

website_snippet_product_category/views/product_public_category.xml
Tardo: Changed, now uses existing "image" fields (image, image_medium, image_small) that resizes the image automatically.

website_snippet_product_category/views/snippets.xml
Tardo: The content of the snippet is not editable with the web editor by design (only can use the snippet options to change the style)

website_snippet_product_category/static/src/js/frontend.js
Tardo: web_editor.snippets.options.SnippetOption is used for snippet options
sAnimation is used for snippet initialization (and other website elements)

This snippet doesn't allow to modify their content. The only way to change it is on the 'backend' or using the snippet options.

website_snippet_product_category/static/src/scss/snippet.scss
Tardo: Umm in this place i prefer use "em" instead of "rem"... if you change the font-size of 'categ-container' you can see how all the other texts are "adapted" to this "parent" rule.

https://www.w3.org/Style/Examples/007/units.en.html#font-size

chienandalu (commented):
Could we get a better responsiveness than this?
image

website_snippet_product_category/static/src/scss/snippet.scss
Tardo: Ummm.... for me try to use BS for this behaviour adds more complexity to the xml file and for alignment operations too

Tardo:
Thanks! changes done

website_snippet_product_category/readme/USAGE.rst
Tardo: :sob:

chienandalu (commented):
image

server error traceback in console
<details>

```
web.assets_common.js:3556 Server application error
 Error code: 200 
 Error message: Odoo Server Error 
 Error data message:
 'JsonRequest' object has no attribute 'render' 
 Error data debug:
 Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/http.py", line 656, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 314, in _handle_exception
    raise pycompat.reraise(type(exception), exception, sys.exc_info()[2])
  File "/.repo_requirements/odoo/odoo/tools/pycompat.py", line 87, in reraise
    raise value
  File "/.repo_requirements/odoo/odoo/http.py", line 698, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 346, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 98, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 941, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/build/OCA/e-commerce/website_snippet_product_category/controllers/website.py", line 22, in render_product_category
    return request.render(template, {
  File "/.repo_requirements/virtualenv/python3.5/lib/python3.5/site-packages/werkzeug/local.py", line 343, in __getattr__
    return getattr(self._get_current_object(), name)
AttributeError: 'JsonRequest' object has no attribute 'render'
```

</details>

website_snippet_product_category/controllers/website.py
Tardo: Can't use 'render' in json request.

Tardo:
Updated! problems using "render" in a json request.

yajo:
Can you check travis please?

Tardo:
Travis green!

chienandalu:
> Travis green!

It'd be cool to have a bot that tried to fix CI exceptions when invoked :rofl: but it's still red. Small LINT error: https://travis-ci.org/github/OCA/e-commerce/jobs/693887622#L314

pedrobaeza:
It'd be cool to have a bot that programs the feature we want :stuck_out_tongue_closed_eyes:

Tardo:
Green Green Green!

rafaelbn:
- All mega menu is open by default so there is not enough space in the screen. Should be only 1st level and you should be able to open and close levels easily.
- The categories don't collapse / un-collapse  bidirectionaly.
- First you must see main categories (parents), after children and children (4 levels).
- You should be able to navigate between level easily and quickly
- And you should be able to open each individually

Like a real mega menu of categories.

Check image, I've use only 2 levels! (of 4 levels) and I have to scroll up and down to navigate just for parents categories :fearful: 



---------------------------
The expected functionality by the customer is this in normal screen:



----------------------------
**And this in mobile**



----------------------------

chienandalu:
> * All mega menu is open by default so there is not enough space in the screen. Should be only 1st level and you should be able to open and close levels easily.
> * The categories don't collapse / un-collapse  bidirectionaly.
> * First you must see main categories (parents), after children and children (4 levels).
> * You should be able to navigate between level easily and quickly
> * And you should be able to open each individually

Those should be theming decisions, in my opinion.

> Check image, I've use only 2 levels! (of 4 levels) and I have to scroll up and down to navigate just for parents categories 

Maybe a subcategory limit setting could be useful here (with an extra item of "More...")

> The expected functionality by the customer is this in normal screen:
> **And this in mobile**

I don't see those very usable, to be honest. In desktop, it's up to the website owner to set up their public categories conveniently. Something that could be interesting is to be able to place snippets with different root categories, so you can have diferent trees with a more convenient structure (I think it's in the roadmap)

A good mobile experience for me would something like this:
megamenu-mobile

pedrobaeza:
That category layout is mixing brands with types. That's not a very well thought category layout. It should be rationalized.

Tardo:
Some changes was made... it's difficult handle this in a good way for everyone :/ i think that have a clean base to extend with particular use cases is fine.

pedrobaeza:
Can you please update the animation with latest style and mobile view?

Tardo:
First comment updated

chienandalu (approved):
Ok, let's go for it  :)

pedrobaeza (approved):

pedrobaeza:
@Yajo can you change your request changes review?

HaraldPanten (approved):

website_snippet_product_category/static/src/js/frontend.js
```diff
@@ -0,0 +1,88 @@
...
+                            .val();
+                        if (!count) {
+                            self.$target.append(
+                                $("<div/>").append(
+                                    $("<div/>", {
+                                        class:
+                                            "alert alert-warning" +
+                                            " alert-dismissible text-center",
+                                        text: _t(
+                                            "No categories was found. Make" +
```
yajo:
```suggestion
                                            "No categories were found. Make" +
```

website_snippet_product_category/static/src/scss/snippet.scss
```diff
@@ -0,0 +1,37 @@
...
+
+    .categ_img img {
+        width: 1.4em;
+    }
+    .categ_tree_level {
+        &[data-tree-level="2"] {
+            font-size: 0.9em;
+            .categ_link {
+                text-transform: capitalize;
+                color: $gray-800;
```
yajo, line 19: Have you considered to use [BS text classes](https://getbootstrap.com/docs/4.5/utilities/colors/#color) instead? I think that being standard BS4 classes will make it easier that their colors fit with any BS theme out there.

yajo (changes requested):
I'm afraid it is still a little bit clunky:

Peek 09-06-2020 13-56

website_snippet_product_category/static/src/scss/snippet.scss
Tardo: I prefer use the variables because i want to the gray-scale not colours... but if you have a good combination, please tell me and i will change it.
"color-secondary" changes dramatically theme by theme :/

Tardo:
Changes done

yajo (approved):

PR #362, opened 2020-05-27
[IMP] website_sale_stock_list_preview: hide the add to cart button when no stock available

CarlosRoca13:
cc @Tecnativa TT22731

### DIFF
--- a/website_sale_stock_list_preview/static/src/js/tour.js
+++ b/website_sale_stock_list_preview/static/src/js/tour.js
@@ -8,50 +8,116 @@ odoo.define("website_sale_stock_list_preview.tour", function(require) {
 
     var steps = [
         {
-            trigger: "a:contains('Test Product 1')",
-            extra_trigger: ".text-success:contains('30  available')",
+            trigger:
+                ".o_wsale_product_information:has(.text-success:contains('30 Units available')) a:contains('Test Product 1')",
         },
         {
             trigger: "a[href='/shop']",
+            extra_trigger:
+                "#product_details .text-success:contains('30 Units available')",
         },
         {
-            trigger: "a:contains('Test Product 2')",
-            extra_trigger: ".text-success:contains('In stock')",
-        },
-        {
-            trigger: "a[href='/shop']",
-        },
-        {
-            trigger: "a:contains('Test Product 3')",
-            extra_trigger: ".text-warning:contains('5  available')",
+            trigger:
+                ".o_wsale_product_information:not(:has(.btn-secondary.disabled)) a:contains('Test Product 1')",
         },
         {
             trigger: "a[href='/shop']",
+            extra_trigger: "#product_details a:not(#add_to_cart.disabled)",
         },
         {
-            trigger: "a:contains('Test Product 4')",
-            extra_trigger: ".text-success:contains('test message')",
-        },
-        {
-            trigger: "a[href='/shop']",
-        },
-        {
-            trigger: "a:contains('Test Product 5')",
-            //            Extra_trigger: ".text-success:contains('30  available')",
+            trigger:
+                ".o_wsale_product_information:has(.text-success:contains('In stock')) a:contains('Test Product 2')",
         },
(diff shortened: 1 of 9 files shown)


CarlosRoca13:
I'm working for test improvement, wait for this changes.

website_sale_stock_list_preview/static/src/js/tour.js
```diff
@@ -53,6 +52,9 @@ odoo.define("website_sale_stock_list_preview.tour", function(require) {
             trigger: "a:contains('Test Product 7')",
             extra_trigger: ".text-danger:contains(' Temporarily out of stock')",
         },
+        {
+            trigger: "a[href='/shop']",
```
yajo, line 114: This is probably not doing what you expect. It will get triggered before the previous step succeeds.

Add an `extra_trigger` that checks one element exists before hitting this link.

website_sale_stock_list_preview/static/src/js/website_sale_stock_list_preview.js
```diff
@@ -47,6 +47,16 @@ odoo.define("website_sale_stock_list_preview.shop_stock", function(require) {
                                 )
                             ).get(0)
                         );
+                    if (
+                        product.virtual_available == 0 &&
+                        (product.inventory_availability == "always" ||
+                            product.inventory_availability == "threshold")
+                    ) {
+                        $(product_dic[product.id])
+                            .find(".fa-shopping-cart")
```
yajo, line 71: Is there a client qweb template that displays this? Can you add a `t-if` in that template?

sergio-teruel: @Yajo What about of load performance....¿?¿?¿??¿

website_sale_stock_list_preview/static/src/js/tour.js
CarlosRoca13: This is just for return to the grid page. Have I to check an extra_trigger at all?

website_sale_stock_list_preview/static/src/js/website_sale_stock_list_preview.js
yajo: It depends on the specific case. May you point me to where the shopping cart icon is rendered? This way I can see if there's a better way to do this.

In any case, consider this: if the availability is being lazily-loaded as I see here (which is OK if that takes long), there's a possible race condition where a user could click the cart icon before it gets hidden. Wouldn't it be better to hide it before, and just show it later here if there's stock? :thinking:

website_sale_stock_list_preview/static/src/js/tour.js
yajo: Yes, actually looking at the logs, the tour is not working fine: https://travis-ci.org/github/OCA/e-commerce/jobs/691706160#L1606-L1620

It should display page reloads between steps, but there are none.

Imagine, let's describe how the previous step and this one should behave:

1. Wait until there's a `.text-danger` that contains "Temporarily out of stock".
1. Click on an `<a>` that contains "Test Product 7".
1. **Wait until the page for Test Product 7 is loaded.** (this is missing).
1. Click on the shop menu item.

As it isn't instructed to wait, it clicks the shop menu before the product page is loaded, so the tour isn't actually imitating what the user would do.

The same happens with most steps.

HaraldPanten:
Hi @CarlosRoca13 is it ready for reviews? If it is, please, remove [WIP] 😉

CarlosRoca13:
@HaraldPanten I'm working on tests

HaraldPanten:
> @HaraldPanten I'm working on tests

OK, thanks!

rafaelbn (approved):
Tested in runbot :+1:

HaraldPanten (approved):
I think it's OK. Depending on configuration/stock, button is faded. 👍

yajo (approved):

PR #369, opened 2020-06-02
[12.0][ADD] website_snippet_carousel_product

Tardo:
carousel

cc @tecnativa TT23634


HaraldPanten (approved):
GJ @Tardo . Amazing module. Functional review tested on runbot 👍

chienandalu (commented):
Looks like the domain doesn't work

domain-error

Looks like some json mismatch

pedrobaeza:
About the domain, it's evaled through `pyeval`?

website_snippet_carousel_product/controllers/website.py
```diff
@@ -0,0 +1,46 @@
...
+    @http.route(
+        ["/website/render_carousel_product"],
+        type="json",
+        auth="public",
+        website=True,
+        csrf=False,
+        cache=30,
+    )
+    def render_carousel_product(
+        self, template, domain=False, limit=12, pps=4, **kwargs
```
yajo: What is `pps`? Would you mind to use a more descriptive name?

website_snippet_carousel_product/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+#. Go to website (frontend)
+#. Press 'Edit'
+#. Drag & Drop the 'Carousel Products' snippet
```
yajo:
```suggestion
#. Drag & Drop the 'Products Carousel' snippet
```

website_snippet_carousel_product/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+#. Write the domain or leave it empty to show all published products
+#. Save
+
+Theming
+~~~~~~~
+This snippet uses 'native' bootstrap carousel and the 'website_sale.products_item' qweb template
```
yajo:
```suggestion

This snippet uses 'native' bootstrap carousel and the ``website_sale.products_item`` qweb template
```

website_snippet_carousel_product/static/src/js/s_carousel_product_frontend.js
```diff
@@ -0,0 +1,101 @@
...
+            var template =
+                this.$target.attr("data-template") ||
+                "website_snippet_carousel_product.s_carousel_product_items";
+
+            // Compatibility with db that saved content inside by mistake
+            this.$target.empty();
+            // Prevent user edition
+            this.$target.attr("contenteditable", "False");
+
+            var def = $.Deferred();
```
yajo: Aren't we supposed to use promises in v12? :thinking: 

```suggestion
            var def = new Promise();
```

website_snippet_carousel_product/static/src/js/snippet.options.js
```diff
@@ -0,0 +1,114 @@
...
+    options.registry.js_carousel_product = options.Class.extend({
+        popup_template_id: "editor_new_carousel_product_domain",
+        popup_title: _t("Add a Carousel Products"),
+
+        /**
+         * @override
+         */
+        _onLinkClick: function (ev) {
+            var self = this;
+            var $elm = $(ev.target);
```
yajo, line 21: Use `this.$target` instead:

```suggestion
            var $elm = this.$target;
```

website_snippet_carousel_product/static/src/js/snippet.options.js
```diff
@@ -0,0 +1,114 @@
...
+        /**
+         * Reset target snippet
+         */
+        _refreshTargetSnippet: function () {
+            this.trigger_up("animation_start_demand", {
+                editableMode: true,
+                $target: this.$target,
+            });
+        },
+    });
```
yajo, line 137: I'm missing the `cleanForSave` method.

You probably don't want the products to be saved in the template code, since they are supposed to change dynamically.

IMHO  you should clean all the code and leave just the container that will trigger the animation.

website_snippet_carousel_product/static/src/scss/s_carousel_product.scss
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 Tecnativa - Alexandre Díaz
+// License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+.s_product_carousel {
+    .oe_product {
+        border-width: 0;
+
+        .oe_product_image {
+            position: unset;
+            padding: 15px;
```
yajo: Can't you use normal bootstrap classes instead? things like a `.col-*` inside a `.row`, etc.

website_snippet_carousel_product/static/src/scss/s_carousel_product.scss
```diff
@@ -0,0 +1,29 @@
...
+        .carousel-control-prev-icon {
+            background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23000000' viewBox='0 0 8 8'%3E%3Cpath d='M5.25 0l-4 4 4 4 1.5-1.5-2.5-2.5 2.5-2.5-1.5-1.5z'/%3E%3C/svg%3E");
+        }
+    }
+    .carousel-control-next {
+        left: 100%;
+        .carousel-control-next-icon {
+            background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23000000' viewBox='0 0 8 8'%3E%3Cpath d='M2.75 0l-1.5 1.5 2.5 2.5-2.5 2.5 1.5 1.5 4-4-4-4z'/%3E%3C/svg%3E");
+        }
+    }
```
yajo: This is [in standard bootstrap code](https://github.com/odoo/odoo/blob/d4255e01d67154a60cd0c31070ac62e581f50340/addons/web/static/lib/bootstrap/scss/_carousel.scss#L149-L150).

Are you redefining these here for some specific reason?

website_snippet_carousel_product/templates/snippet.xml
```diff
@@ -0,0 +1,141 @@
...
+            position="inside"
+        >
+            <t
+                t-snippet="website_snippet_carousel_product.s_carousel_product"
+                t-thumbnail="/website_snippet_carousel_product/static/src/img/snippet_carousel_product.png"
+            />
+        </xpath>
+    </template>
+    <!-- Snippet Carousel Product -->
+    <template id="s_carousel_product" name="Carousel Products">
```
yajo:
```suggestion
    <template id="s_carousel_product" name="Products Carousel">
```

website_snippet_carousel_product/templates/snippet.xml
```diff
@@ -0,0 +1,141 @@
...
+            <div
+                t-att-id="uniqueId"
+                class="carousel slide o_not_editable"
+                data-ride="carousel"
+            >
+                <div class="carousel-inner">
+                    <t t-set="products_index" t-value="0" />
+                    <t t-foreach="objects" t-as="products">
+                        <div
+                            t-attf-class="carousel-item #{not products_index and 'active' or ''}"
```
yajo: This is the same, but easier to read. Please take this as a rule of thumb:

```suggestion
                            t-attf-class="carousel-item #{'' if products_index else 'active'}"
```

website_snippet_carousel_product/templates/snippet.xml
```diff
@@ -0,0 +1,141 @@
...
+                                <t t-foreach="products" t-as="product">
+                                    <div t-attf-class="col-md-#{int(12 / pps)}">
+                                        <div t-attf-class="oe_product #{' '.join(x.html_class for x in product.website_style_ids if x.html_class)}" >
+                                            <t t-call="website_sale.products_item" />
+                                        </div>
+                                    </div>
+                                </t>
+                            </div>
+                        </div>
+                        <t t-set="products_index" t-value="products_index + 1" />
```
yajo: Also unneeded.

```suggestion
```

website_snippet_carousel_product/templates/snippet.xml
```diff
@@ -0,0 +1,141 @@
...
+                t-if="objects"
+                t-att-value="len(objects)"
+            />
+            <div
+                t-att-id="uniqueId"
+                class="carousel slide o_not_editable"
+                data-ride="carousel"
+            >
+                <div class="carousel-inner">
+                    <t t-set="products_index" t-value="0" />
```
yajo: You don't need this.

Qweb itself declares this variable automatically: https://www.odoo.com/documentation/12.0/reference/qweb.html#loops

```suggestion
```

yajo (changes requested):
Some changes in code.

Also:

1. If you move `/static/src/img/snippet_carousel_product.png` to `/static/description/icon.png`, you can use the same icon for the snippet and for the module itself. It will look better for free.
1. I guess you'll understand that the domain selector is pretty unusable for a normal human. At least, add that to the roadmap/known issues please.

website_snippet_carousel_product/controllers/website.py
Tardo: 1. Ok
2. Yes, but Odoo doesn't offer any domain selector for the frontend (or i can't see it)... For me the domain selector is outside the scope of this module.

PPS = Products Per Slide (Similar to the PPG (products per page) from Odoo)

website_snippet_carousel_product/static/src/scss/s_carousel_product.scss
Tardo: This part is a workarround to show the tiles... i don't like the idea of change the 'website_sale.products_item' template.

website_snippet_carousel_product/static/src/scss/s_carousel_product.scss
Tardo: Yes, change the color from white to black.

website_snippet_carousel_product/templates/snippet.xml
Tardo: I know the linter rule.. but here is a different case: short-circuit evaluation VS ternary operator

I can read all the alternatives without any difficult... but ok, i will change it to use ternary operator without negations.

website_snippet_carousel_product/templates/snippet.xml
Tardo: Why? without this you will get all slides as "active".

website_snippet_carousel_product/static/src/scss/s_carousel_product.scss
Tardo: Bootstrap carousel is really used for images... and puts the controls inside the carousel.
The snippet moves these controls to the outside of the carousel (because blocks mouse interaction with the product tile) and change the colour to black.

website_snippet_carousel_product/static/src/js/snippet.options.js
Tardo: Is not the same:
- this.$target refers to the snippet
- $elm refers to the menu option

Tardo:
Some suggestions done, waiting replies to resolve all of them.
@pedrobaeza no, pyeval is not present "by default" in the website. So, you need use json notation to write the domain (a little bit hacked here to replace simple commas by double commas to avoid annoying parser errors)
@chienandalu fixed! a problem replacing simple commas.

website_snippet_carousel_product/readme/CONFIGURATION.rst
```diff
@@ -0,0 +1,7 @@
+#. Go to website (frontend)
+#. Press 'Edit'
+#. Select the snippet that you want configure
+#. Press 'Customize' dropdown > Here You can see the implemented options for the widget:
+    * ``Set Domain`` > Sets a new domain to use
+    * ``Limit`` > Sets how many products can be getted
+    * ``Show`` > Sets how many products can be show per slide
```
pedrobaeza:
```suggestion
    * ``Show`` > Sets how many products can be shown per slide
```

website_snippet_carousel_product/readme/CONFIGURATION.rst
```diff
@@ -0,0 +1,7 @@
+#. Go to website (frontend)
+#. Press 'Edit'
+#. Select the snippet that you want configure
+#. Press 'Customize' dropdown > Here You can see the implemented options for the widget:
+    * ``Set Domain`` > Sets a new domain to use
+    * ``Limit`` > Sets how many products can be getted
```
pedrobaeza:
```suggestion
    * ``Limit`` > Sets how many products will be fetched at maximum following default product order
```

website_snippet_carousel_product/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 Tecnativa - Alexandre D. Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Website Snippet Carousel Product",
+    "version": "12.0.1.0.0",
+    "category": "Website",
+    "author": "Tecnativa, " "Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/e-commerce",
+    "license": "AGPL-3",
```
pedrobaeza: For website modules, we use LGPL-3 for not forcing themes over this module to be released. Please review all your website modules according this.

```suggestion
    "license": "LGPL-3",
```

website_snippet_carousel_product/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Adds a new snippet to insert a products carousel with a custom domain
```
pedrobaeza: This is user documentation

```suggestion
Adds a new snippet to insert a products carousel with a custom filter
(which is set through a technical domain).
```

website_snippet_carousel_product/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+#. Go to website (frontend)
+#. Press 'Edit'
+#. Drag & Drop the 'Products Carousel' snippet
```
pedrobaeza: In UI the used word is block, isn't it?

```suggestion
#. Drag & Drop the 'Products Carousel' block
```

website_snippet_carousel_product/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
...
+#. Write the domain or leave it empty to show all published products
+#. Save
+
+Theming
```
pedrobaeza: I would put this more in the configuration section.

website_snippet_carousel_product/controllers/website.py
```diff
@@ -0,0 +1,46 @@
...
+    ):
+        records = request.env["product.template"].search(domain or [], limit=limit)
+
+        records_grouped = []
+        records_list = []
+        for index, record in enumerate(records):
+            records_list.append(record)
+            if (index + 1) % products_per_slide == 0:
+                records_grouped.append(records_list.copy())
+                records_list.clear()
```
pedrobaeza: Instead of copying and clearing, which is heavier, simply reassign:
```suggestion
                records_grouped.append(record_list)
                record_list = []
```

website_snippet_carousel_product/controllers/website.py
```diff
@@ -0,0 +1,46 @@
...
+        csrf=False,
+        cache=30,
+    )
+    def render_product_carousel(
+        self, template, domain=False, limit=12, products_per_slide=4, **kwargs
+    ):
+        records = request.env["product.template"].search(domain or [], limit=limit)
+
+        records_grouped = []
+        records_list = []
```
pedrobaeza: Avoid plurals in the middle of the word that are very prone to typos (and `list` semantically already indicates the plural):
```suggestion
        record_list = []
```

website_snippet_carousel_product/controllers/website.py
```diff
@@ -0,0 +1,46 @@
...
+    @http.route(
+        ["/website/render_product_carousel"],
+        type="json",
+        auth="public",
+        website=True,
+        csrf=False,
+        cache=30,
+    )
+    def render_product_carousel(
+        self, template, domain=False, limit=12, products_per_slide=4, **kwargs
```
pedrobaeza: Why passing the template as an argument? I think we should it that for avoiding exploits or information leak.

website_snippet_carousel_product/static/src/scss/s_product_carousel.scss
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 Tecnativa - Alexandre Díaz
+// License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+.s_product_carousel {
```
pedrobaeza, line 3: This class is not mentioned in the theming section.

pedrobaeza (changes requested):

chienandalu:
The domain works but now the other settings don't:

prod-car-cfg

Tardo:
Changes done

chienandalu (approved):
Cool! :+1: 

(would it be very hard to have a setting to avoid the coroussel animation?)

Tardo:
@chienandalu thanks! added a new option to customize the slide swapping interval

chienandalu:
> @chienandalu thanks! added a new option to customize the slide swapping interval

Great!

website_snippet_carousel_product/static/src/scss/s_carousel_product.scss
yajo: OK, but then try using rem units, as it's the standard in BS4.

website_snippet_carousel_product/static/src/js/s_product_carousel_frontend.js
```diff
@@ -0,0 +1,104 @@
...
+            });
+        },
+
+        /**
+         * @override
+         */
+        start: function () {
+            var self = this;
+            var limit = Number(this.$target.attr("data-products-limit")) || 12;
+            var domain = this.$target.attr("data-domain") || false;
```
yajo:
```suggestion
            var domain = this.$target.attr("data-domain") || "[]";
```

Otherwise the `JSON.parse()` call you do later will fail.

website_snippet_carousel_product/static/src/js/s_product_carousel_frontend.js
```diff
@@ -0,0 +1,104 @@
...
+            var products_per_slide =
+                Number(this.$target.attr("data-products-per-slide")) || 4;
+            var interval = this.$target.attr("data-interval");
+            if (typeof interval === 'undefined') {
+                interval = 5000;
+            } else {
+                interval = Number(interval);
+            }
```
yajo: I'd do it like this:

```suggestion
            var interval = Number(this.$target.attr("data-interval"));
            if (_.isNaN(interval)) {
                interval = 5000;
            }
```

website_snippet_carousel_product/static/src/js/s_product_carousel_frontend.js
```diff
@@ -0,0 +1,104 @@
...
+                                        ),
+                                    })
+                                )
+                            );
+                            return;
+                        }
+
+                        self.$target.html($object_html);
+                        self.$target.find('.carousel').carousel({
+                            interval: interval,
```
yajo, line 73: I'm not sure this is the best way for setting the interval.

The common pattern is to use snippet options and just store [the `data-interval` attribute](https://getbootstrap.com/docs/4.5/components/carousel/#options) with the expected value, just like the upstream carousel does:

Peek 09-06-2020 08-26

You'd probably be able to reuse upstream snippet options if you make your snippet match [this selector](https://github.com/odoo/odoo/blob/b88900187b87b2ca8c18d712292a367671a1fff5/addons/website/views/snippets.xml#L1077), or if you extend it to match your snippet.

website_snippet_carousel_product/static/src/js/snippet.options.js
```diff
@@ -0,0 +1,139 @@
...
+
+    var core = require("web.core");
+    var options = require("web_editor.snippets.options");
+    var wUtils = require("website.utils");
+
+    var _t = core._t;
+
+    options.registry.js_product_carousel = options.Class.extend({
+        popup_template_id: "editor_new_product_carousel_domain",
+        popup_title: _t("Add a Carousel Products"),
```
yajo:
```suggestion
        popup_title: _t("Add a products carousel"),
```

website_snippet_carousel_product/static/src/js/snippet.options.js
```diff
@@ -0,0 +1,139 @@
...
+
+        /**
+         * @override
+         */
+        _onLinkClick: function (ev) {
+            var self = this;
+            var $elm = $(ev.target);
+            if ($elm.is(".s_carousel_set_domain")) {
+                this.select_domain().fail(function () {
+                    self.getParent()._onRemoveClick($.Event("click"));
```
yajo: This is an old programming pattern that was being used in Odoo v8-v10.

The common pattern for v11+ in JS is never to search for parents, but [bubble up events instead](https://www.odoo.com/documentation/12.0/reference/javascript_reference.html#extended-event-system).

website_snippet_carousel_product/static/src/js/snippet.options.js
yajo: Then it's `this.$el`, [check the docs](https://github.com/odoo/odoo/blob/b88900187b87b2ca8c18d712292a367671a1fff5/addons/web_editor/static/src/js/editor/snippets.options.js#L34-L37) (a little bit hidden as always :laughing:).

website_snippet_carousel_product/static/src/js/snippet.options.js
```diff
@@ -0,0 +1,139 @@
...
+        },
+
+        /**
+         * @override
+         */
+        onBuilt: function () {
+            var self = this;
+            this._super();
+            this.select_domain().fail(function () {
+                self.getParent()._onRemoveClick($.Event("click"));
```
yajo: Also `trigger_up` if possible.

website_snippet_carousel_product/static/src/js/s_product_carousel_frontend.js
```diff
@@ -0,0 +1,104 @@
...
+                Number(this.$target.attr("data-products-per-slide")) || 4;
+            var interval = this.$target.attr("data-interval");
+            if (typeof interval === 'undefined') {
+                interval = 5000;
+            } else {
+                interval = Number(interval);
+            }
+
+            // Compatibility with db that saved content inside by mistake
+            this.$target.empty();
```
yajo: How about adding `t-ignore="True"` to the snippet, to avoid the server saving whatever is inside that block?

Also I see you have a `cleanForSave` below, so is this really gonna happen at any point?

website_snippet_carousel_product/static/src/js/s_product_carousel_frontend.js
```diff
@@ -0,0 +1,104 @@
...
+            // Prevent user edition
+            this.$target.attr("contenteditable", "False");
```
yajo, line 29: I think it makes more sense to make the whole snippet have a `o_not_editable` class to avoid user editions.

website_snippet_carousel_product/static/src/scss/s_product_carousel.scss
```diff
@@ -0,0 +1,31 @@
+// Copyright 2020 Tecnativa - Alexandre Díaz
+// License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+.s_product_carousel {
+    .oe_product {
+        border-width: 0;
+
+        .oe_product_image {
+            position: unset;
+            padding: 15px;
```
yajo: I wonder if you can replace this for something similar to:

```suggestion
            padding: $grid-gutter-width / 2;
```

Or otherwise use rem units as the rest of BS4 please.

website_snippet_carousel_product/static/src/scss/s_product_carousel.scss
```diff
@@ -0,0 +1,31 @@
...
+                position: unset;
+            }
+        }
+    }
+
+    .carousel-control-prev {
+        width: 5%;
+        left: -5%;
+        .carousel-control-prev-icon {
+            background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23000000' viewBox='0 0 8 8'%3E%3Cpath d='M5.25 0l-4 4 4 4 1.5-1.5-2.5-2.5 2.5-2.5-1.5-1.5z'/%3E%3C/svg%3E");
```
yajo: I admit I don't really like to maintain this. Also the hardcoded color can fight some themes stylings.

Can't you just put a `<i class="text-secondary fa fa-chevron-left"/>` inside the carousel control? Look at these changes I made to the structure, it looks exactly the same, but it follows the theme colors and we don't need to maintain this:

chevrons

website_snippet_carousel_product/static/src/scss/s_product_carousel.scss
```diff
@@ -0,0 +1,31 @@
...
+
+            img {
+                position: unset;
+            }
+        }
+    }
+
+    .carousel-control-prev {
+        width: 5%;
+        left: -5%;
```
yajo: You can use standard Bootstrap markup to solve these problems in a way that is more maintainable and readable IMHO, as explained [here](https://stackoverflow.com/a/59377509), and make it look pretty much the same:

layout

website_snippet_carousel_product/static/src/scss/s_product_carousel.scss
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+    .carousel-control-next {
+        width: 5%;
+        left: 100%;
+        .carousel-control-next-icon {
+            background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' fill='%23000000' viewBox='0 0 8 8'%3E%3Cpath d='M2.75 0l-1.5 1.5 2.5 2.5-2.5 2.5 1.5 1.5 4-4-4-4z'/%3E%3C/svg%3E");
+        }
+    }
+}
```
yajo: Same as above, to be removed.

yajo (changes requested):

website_snippet_carousel_product/static/src/js/s_product_carousel_frontend.js
Tardo: I see more problems than advantages trying to use the attribute instead of js (which in the end is what bootstrap will do).
But, now the interval options works too similar.

website_snippet_carousel_product/static/src/js/snippet.options.js
Tardo: Thanks, this is completely refactored.

website_snippet_carousel_product/static/src/scss/s_product_carousel.scss
Tardo: Refactored

website_snippet_carousel_product/static/src/scss/s_product_carousel.scss
Tardo: I prefer user "d-flex" that works better with small screens

website_snippet_carousel_product/static/src/js/s_product_carousel_frontend.js
Tardo: It doesn't seem to work in snippets

Tardo:
Changes done
carousel

website_snippet_carousel_product/templates/snippet.xml
```diff
@@ -0,0 +1,196 @@
...
+                    <span class="sr-only">Previous</span>
+                </a>
+                <div class="carousel-inner">
+                    <t t-foreach="objects" t-as="products">
+                        <div
+                            t-attf-class="carousel-item #{'' if products_index else 'active'}"
+                        >
+                            <div class="row">
+                                <t t-foreach="products" t-as="product">
+                                    <div t-attf-class="col-#{int(12 / products_per_slide)}">
```
yajo: The snippet lost its responsivity:

Captura de pantalla de 2020-06-11 13-22-36

I suggest you to use this instead, or something similar:

```suggestion
                                    <div t-attf-class="col-12 col-sm-6 col-md-#{int(12 / products_per_slide)}">
```

yajo (changes requested):

Tardo:
Changes done

yajo (approved):

pedrobaeza (approved):

PR #372, opened 2020-06-03
[12.0][ADD] website_sale_attribute_filter_price

Tardo:
pricefilter

cc @tecnativa TT23634


website_sale_attribute_filter_price/controllers/website_sale.py
```diff
@@ -0,0 +1,59 @@
...
+        try:
+            custom_max_price = float(post.get("max_price"))
+        except (ValueError, TypeError):
+            custom_max_price = max_price
+        if custom_min_price > custom_max_price:
+            custom_max_price, custom_max_price = custom_min_price, custom_max_price
+        # Call Super with context
+        request.env.context = dict(request.env.context,
+            price_vals = [custom_min_price, custom_max_price]
+        )
```
chienandalu: The lint error comes from here
```suggestion
        request.env.context = dict(
            request.env.context,
            price_vals=[custom_min_price, custom_max_price])
```

chienandalu (commented):
* Check travis
* There's some issue loading the lib assets:
image

pedrobaeza:
Isn't there a slider in existing resources for not requiring an extra library?

pedrobaeza (changes requested):
Two functional problems:

- Incorrect limits:

Selección_009

The same may happen when you search for a text, and also applies to the minimum value.

- No price filter:

Selección_010

And check also that Travis is red

chienandalu (commented):
A pair of usability problems I found

price-filter

Tardo:
@chienandalu thanks. Improved the usability... but the problem with the 'Sort By' is a 'core' issue. This happens because the inputs can't use 'keep' method and the form doesn't have the 'order' value (as "input type='hidden'")

chienandalu:
Great. Being picky, one would expect to re-apply the filter once you click on the reset button
erase-filter

Tardo:
@chienandalu changes made, removed all custom css and compatible with currency symbol position

pricefilter

chienandalu (approved):
Great!

pedrobaeza (approved):

PR #377, opened 2020-06-11
[ADD] website_sale_product_attachment: add attachments to products in website

yajo:
This module lets you publish downloadable attachments in a product page.

This is useful if you want to publish firmwares, manuals, specs, warranties,
or whatever document related to the product.

@Tecnativa TT20984 TT23657


chienandalu (approved):
Awesome! :+1:

website_sale_product_attachment/README.rst
```diff
@@ -0,0 +1,137 @@
...
+=============
+
+Instead of configuring attachments per product, to configure products per attachment:
+
+#. You need permissions to access system settings.
+#. Go to *Settings > Dashboard > Activate the developer mode*.
+#. Go to *Settings > Technical > Database Structure > Attachments*.
+#. Pick or create one.
+#. Use the field *Attached in product templates* to choose the products.
+
```
HaraldPanten, line 48: Functional review 👍 

Just a minor change. I would add a little explanation that user should check "Is public document".

HaraldPanten (approved):

website_sale_product_attachment/README.rst
pedrobaeza: Is that really needed? I don't think so.

HaraldPanten: I tried both cases. If the checkbox is set, user can see the attachments. If the chechbox is not set, attachments are not appearing.

Did I miss something?

I think that the "document downloading" functionality depends on that checkbox. Doesn't it?

yajo: Oh yes, good point.

ValentinVinagre:
LGTM 👍 but... why doesn't it have tests?

website_sale_product_attachment/README.rst
pedrobaeza: We have talked internally and the constraint is needed for avoiding leaks, so we are going to document it, but keep it.

pedrobaeza:
@ValentinVinagre there are no backend things to test. Only a tour makes a bit sense, but we are over the budget, so no tests for now.

@Yajo please put it on the roadmap.

ValentinVinagre:
> @ValentinVinagre there are no backend things to test. Only a tour makes a bit sense, but we are over the budget, so no tests for now.
> 
> @Yajo please put it on the roadmap.

Okay, no problem, I was curious.

yajo:
All attended.

HaraldPanten:
> All attended.

IMO, it's ready to merge

pedrobaeza (approved):

PR #381, opened 2020-06-15
[FIX] website_sale_product_detail_attribute_image: rename file

wtaferner:
It depends on the used filesystem, so basically colon is destroying the use of this repo in incompatible filessystems -> https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations (reserved characters)

exFAT, NTFS, HFS, HFS+ to name the most prominent filesystems

Info: @wt-io-it

Closes https://github.com/OCA/e-commerce/issues/379

PR #382, opened 2020-06-15
[FIX] website_sale_product_detail_attribute_image: rename file

wtaferner:
It depends on the used filesystem, so basically colon is destroying the use of this repo in incompatible filessystems -> https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations (reserved characters)

exFAT, NTFS, HFS, HFS+ to name the most prominent filesystems

Info: @wt-io-it

Forward port https://github.com/OCA/e-commerce/pull/381


HaraldPanten (approved):

HaraldPanten:
I understand it's urgent like the other FIX

PR #391, opened 2020-06-25
[MIG] website_sale_product_reference_displayed: Migration to 12.0

chienandalu:
- Standard migration
- Adapt to BS4

cc @Tecnativa TT23639


website_sale_product_reference_displayed/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+Just install and the products will be shown with their full display name.
```
yajo, line 1: You could explain how to toggle this behavior per website.

website_sale_product_reference_displayed/views/website_sale_views.xml
```diff
@@ -0,0 +1,46 @@
...
+<odoo>
+
+    <template id="products_item" name="Display product reference" inherit_id="website_sale.products_item" customize_show="True">
+        <!-- Product title in list -->
+        <xpath expr="//a[@t-field='product.name']" position="attributes">
+            <attribute name="class" add="d-none" separator=" "/>
+        </xpath>
+        <xpath expr="//a[@t-field='product.name']" position="after">
+            <a itemprop="name" t-att-href="product_href" t-att-content="product.name" t-field="product.name"/>
+        </xpath>
```
yajo: Wouldn't it be easier to just alter those attributes in the 1st matched element? :thinking:

website_sale_product_reference_displayed/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,2 @@
+This module extends the ``website_sale`` views to display the product's full
+display name, with its product reference included.
```
yajo, line 2: A screenshot would be appreciated.

yajo (approved):
Looks good. Some non-requirement suggestions.

It seems the bot is unhappy. No travis here?

website_sale_product_reference_displayed/views/website_sale_views.xml
chienandalu: There was a mistake here (solved).

Anyway, the attribute I want to change is the field value itself. So the approach is:

- hide the original field (product.name)
- show a copy of the same tag but with the other (product.display_name)

website_sale_product_reference_displayed/readme/DESCRIPTION.rst
chienandalu: Added

website_sale_product_reference_displayed/readme/USAGE.rst
chienandalu: Done

HaraldPanten (approved):
Functional review 👍 

Just a non-blocking question:

Does it make sense to have the "Display product reference" option available in website products form? 🤔

chienandalu:
> Functional review 
> 
> Just a non-blocking question:
> 
> Does it make sense to have the "Display product reference" option available in website products form?

For our customer it does :sweat_smile: Their point is to have de model reference (wich is a relevant info in their case) in the page content.

HaraldPanten:
> > Functional review
> > Just a non-blocking question:
> > Does it make sense to have the "Display product reference" option available in website products form?
> 
> For our customer it does 😅 Their point is to have de model reference (wich is a relevant info in their case) in the page content.

Hi @chienandalu I think I didn't explain my question very well. I'm not casting doubts on the module usability; I think it's an interesting feature.

I mean if this option:

Captura de pantalla 2020-06-26 a las 17 35 11

Is needed in the website products form, or it should be only available to be activated in the products grid. Do you know what I mean?

chienandalu:
> Is needed in the website products form, or it should be only available to be activated in the products grid. Do you know what I mean?

Yes, and in alternative products as well

PR #392, opened 2020-06-29
[13.0][IMP] website_sale_attribute_filter_category: Use sass instead of les…

mamcode:
…s for styles.

Sass seems to be the most used now for css styles. I didn't make any changes to the code just change the file extension and directory name to scss.


pedrobaeza:
It's not only the most used, but the one that avoids lessc dependency.

CarlosRoca13 (approved):
I test it and continue working! :+1: LGTM

website_sale_attribute_filter_category/views/assets.xml
```diff
@@ -9,7 +9,7 @@
             <link
                 type="text/less"
```
Tardo:
```suggestion
                type="text/css"
```

Tardo (changes requested):
Looks good! thanks, please update the mimetype too

mamcode:
> Looks good! thanks, please update the mimetype too

Thanks :) Done.

HaraldPanten (approved):

Tardo (approved):

PR #394, opened 2020-06-30
[FIX][12.0] website_sale_product_available_display: Take into account negative units to display danger style message

sergio-teruel:
cc @Tecnativa TT24520
@chienandalu @Tardo Can you review, please!!

PR #395, opened 2020-07-01
[IMP] website_sale_stock_available: always change virtual_available in website

ernestotejeda:
In the website, always change virtual_available by immediately_usable_qty

cc @Tecnativa TT24522


pedrobaeza:
Let's close this for now as this can have side effects. Fix moved to #394

PR #407, opened 2020-07-16
[NEW] website_sale_tax_toggle: Test more resilient to CI

sergio-teruel:
cc @Tecnativa 
Removed post_install, with l10n_es account charts the created tax is removed and test fails.


pedrobaeza:
Fast-tracking this:

/ocabot merge patch

PR #408, opened 2020-07-17
product_template_multi_link: fix link type data

simahawk:
@lmignon this bug passed under radars :sweat_smile:


lmignon (approved):
Thank you @simahawk  This one should be fast tracked....

PR #422, opened 2020-08-09
[12.0][ADD] add website_sale_phone_validation

celm1990:


website_sale_phone_validation/controllers/main.py
```diff
@@ -0,0 +1,29 @@
...
+        partner_model = request.env["res.partner"].sudo()
+        error, error_message = super(WebsiteSale, self).checkout_form_validate(
+            mode, all_form_values, data
+        )
+        for field_name in self._get_phone_fields_to_validate_sale():
+            if not data.get(field_name):
+                continue
+            try:
+                data.update(
+                    {field_name: partner_model.phone_format(data.get(field_name))}
```
mamcode: I think that there is a problem with this line, `res.partner` has no the method `phone_format()`, I tried use it and I get an error, please check it in the following recording:
Shop _ My Website
 
Can you please describe how to use the module in readme/USAGE.rst ? Also check the travis logs, something is breaking.

mamcode (changes requested):
Hi Carlos please review my comments...

website_sale_phone_validation/controllers/main.py
celm1990: Hi, yes, method `phone_format `is on model `phone.validation.mixin`, i update code for fix the problem, please try again

PR #420, opened 2020-08-07
[ADD] website_sale_b2x_alt_price: display B2{B,C} complementary price

yajo:
- If your website is B2B, display a smaller B2C price next to the main one. And viceversa.
- Toggle feature independently in list and product templates.
- Toggle feature per website.
- Supports live price recomputing.

@Tecnativa TT24410

PR #423, opened 2020-08-12
[FIX] website_tax_toggle: make tour effective

yajo:
This tour was not effective because:

- Steps did not have proper wait conditions.
- That made them run fast without actually testing the desired effect.
- That left many remaining requests that could made tests to fail randomly, as you can see in https://github.com/OCA/e-commerce/pull/420#issuecomment-672026443.

Now it's shorter but more effective:

- It skips going to `/shop` at the start because it starts already there.
- It moves tour code to demo data.
- It adds proper checks to each trigger, to make sure it doesn't run before checking the result is OK.

@Tecnativa TT24410

### DIFF
--- a/website_sale_tax_toggle/__manifest__.py
+++ b/website_sale_tax_toggle/__manifest__.py
@@ -18,4 +18,5 @@
         'views/assets.xml',
         'views/templates.xml',
     ],
+    "demo": ["demo/assets.xml"],
 }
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):

pedrobaeza:
Fast-tracking this for fixing the branch if everything goes OK:

/ocabot merge minor

HaraldPanten (approved):
Seems to be OK for fixing merge conflicts in https://github.com/OCA/e-commerce/pull/420

chienandalu (approved):

PR #424, opened 2020-08-14
[FIX] website_sale_b2x_alt_price: l10n resilient tests

yajo:
When testing on an integration environment which had a different l10n package enabled, the tours failed due to currency rate conversion.

Create instead a test-specific pricelist and reset currency conversions before testing, to avoid false negative tests. That's not the feature being tested here.

@Tecnativa TT24410

### DIFF
--- a/website_sale_b2x_alt_price/tests/test_ui.py
+++ b/website_sale_b2x_alt_price/tests/test_ui.py
@@ -7,6 +7,15 @@ from odoo.tests.common import HttpCase, Form
 class UICase(HttpCase):
     def setUp(self):
         super().setUp()
+        # Create and select a pricelist
+        # to make tests pass no matter what l10n package is enabled
+        website = self.env["website"].get_current_website()
+        pricelist = self.env["product.pricelist"].create({
+            "name": "website_sale_b2x_alt_price public",
+            "currency_id": website.user_id.company_id.currency_id.id,
+            "selectable": True,
+        })
+        website.user_id.property_product_pricelist = pricelist
         # Create some demo taxes
         self.tax_group_22 = self.env["account.tax.group"].create(
             {"name": "Tax group 22%"}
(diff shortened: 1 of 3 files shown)


website_sale_b2x_alt_price/tests/test_ui.py
```diff
@@ -7,6 +7,16 @@
...
         super().setUp()
+        # Create and select a pricelist and reset currency conversion rates
+        # to make tests pass no matter what l10n package is enabled
+        website = self.env["website"].get_current_website()
+        pricelist = self.env["product.pricelist"].create({
+            "name": "website_sale_b2x_alt_price public",
+            "currency_id": website.user_id.company_id.currency_id.id,
+            "selectable": True,
+        })
+        website.user_id.property_product_pricelist = pricelist
```
pedrobaeza, line 18: I think this line and the following are not needed

yajo: Then how do you specify which pricelist should the tour start with?
Regarding the other lines, they seem to fix the build in the integration environment, so I think I'd prefer to keep them... :roll_eyes:

pedrobaeza: Well, AFAIK, you are selecting the pricelist in the tour, but OK if you don't do for "the default". About currency rate conversion, it shouldn't be needed if you are using the same currency than the company.

yajo: > AFAIK, you are selecting the pricelist in the tour

No, it expects to have one, then switch it to other, then back to the 1st.



> About currency rate conversion, it shouldn't be needed if you are using the same currency than the company

Ok, I'll remove them and retry downstream.

yajo: You're right, the rate conversion stuff was not required. Ready to merge then. Thanks!

pedrobaeza:
Fast-tracking then for having these tests resilient:

/ocabot merge patch

PR #427, opened 2020-09-08
[12.0][ADD] website_sale_attribute_filter_category_collapsable

Tardo:
Extends website_sale_attribute_filter_category with collapsable sections for the attributes. The website_sale_attribute_filter_category module only collapse the category.

attribute_filter_collapse

cc @tecnativa TT25329


pedrobaeza:
Why that huge indentation?

Tardo:
@pedrobaeza it happens because of this rule: https://github.com/odoo/odoo/blob/ab3605081801bc2772c7bf2987096a9b6034c0c8/addons/website_sale/static/src/scss/website_sale.scss#L9

pedrobaeza:
But is it something general without this module? If so, let's keep it that way and fix it on themes.

chienandalu (changes requested):
Thanks @Tardo 

- Check travis for sytax and lint errors
- Isn't it possible to get the same visually compact result as in collapsable categories?:
image

Tardo:
@chienandalu this module doesn't add any CSS rule or "ul" elements... maybe can be done in base 'website_sale_attribute_filter_category' but i thinkg that this must be handle by a custom theme.

chienandalu:
> @chienandalu this module doesn't add any CSS rule or "ul" elements... maybe can be done in base 'website_sale_attribute_filter_category' but i thinkg that this must be handle by a custom theme.

Well, I meant using the same bootstrap classes as in the standard collapsable categories

Tardo:
If we want change the visual behaviour, we need change 'website_sale_attribute_filter_category' module first. This implementation of collapsable elements is better because uses unicode character inyected in css... but yeah, breaks the visual style with Odoo implementation.

website_sale_attribute_filter_category and website_sale_attribute_filter_category_collapsable both modules adds collapsable zones.

@sergio-teruel ping!

sergio-teruel:
> If we want change the visual behaviour, we need change 'website_sale_attribute_filter_category' module first

Ok... do it if necessary.
Why haven't you expanded the website_sale_attribute_filter_category module instead of create a new one?
But for me Ok for the module...

sergio-teruel (approved):
Only if you can review css

pedrobaeza (changes requested):
Let's go for the inclusion of this in the other module as a customize_show option.

PR #429, opened 2020-09-16
[13.0] Add product_variant_multi_link

simahawk:
Extends `product_template_multi_link` to link product variants.

NOTE: on v10 there's [product_multi_link](https://github.com/OCA/e-commerce/blob/10.0/product_multi_link) but is not based on `product_template_multi_link` which is more advanced after refactoring by @lmignon

### DIFF
--- a/product_template_multi_link/models/product_template_link.py
+++ b/product_template_multi_link/models/product_template_link.py
@@ -20,21 +20,18 @@ class ProductTemplateLink(models.Model):
         required=True,
         ondelete="cascade",
     )
-
     right_product_tmpl_id = fields.Many2one(
         string="Linked Product",
         comodel_name="product.template",
         required=True,
         ondelete="cascade",
     )
-
     type_id = fields.Many2one(
         string="Link type",
         comodel_name="product.template.link.type",
         required=True,
         ondelete="restrict",
     )
-
     link_type_name = fields.Char(related="type_id.name")  # left to right
     link_type_inverse_name = fields.Char(
         related="type_id.inverse_name"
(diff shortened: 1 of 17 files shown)


lmignon (approved):
Thank you @simahawk. Cristal clear as always :smirk: (Code review only) . Unfortunately travis is :red_circle:

TDu (approved):

simahawk:
@lmignon @TDu thanks for reviewing. Would you mind giving a look at my last commit https://github.com/OCA/e-commerce/pull/429/commits/0095267e1af4e469cdc436bcf46d989f87c9b00c ?

Right now if you install this module it blocks testing the base module in isolation. You cannot run them a second time (on travis they run once and the template one runs before the variant module is installed.
The change allows to test them independently even if you have installed them at the same time (eg: local testing where you run tests more than once, maybe using pytest).

lmignon (approved):

lmignon:
@simahawk For me its normal that the tests of product_template_multi_link could break once a child addon is installed. I'm not very happy with your last commit. Tests of one addon must only work in a context where only this addon is installed with all its dependencies. We must avoid to add into a parent addon specific code required by child addon. Parent addon must be agnostic of all its children.  When I run tests of an addon, I always create an empty DB where I only install the addon to test.

simahawk:
> @simahawk For me its normal that the tests of product_template_multi_link could break once a child addon is installed. I'm not very happy with your last commit. Tests of one addon must only work in a context where only this addon is installed with all its dependencies. We must avoid to add into a parent addon specific code required by child addon. Parent addon must be agnostic of all its children. When I run tests of an addon, I always create an empty DB where I only install the addon to test.

I got your 2nd approval as an approval for that change :wink: 
Anyway, me too I'm against inter-dependency but recreating a DB every time seems to me a waste of time.
I tried to find an agnostic way to make this work but I couldn't.
A ctx key - which is not required - sounds like a good compromise IMO and does not make the base module tightly coupled to the depending one. Anyway, I can drop it if nobody likes it :man_shrugging:

lmignon:
> Anyway, me too I'm against inter-dependency but recreating a DB every time seems to me a waste of time.

In this case, I often have several DBs at the same time :smirk:  I wasn't conscientious enough during my last review. I'm not going to ask you to change here but the problem is that what sometimes seems to be a small compromise in one case may become a basic rule for others and we'll soon end up with spaghetti code.(Anyway, I'm sure this will be taken as an argument to justify code not well designed in other addons).

PR #430, opened 2020-09-16
[12.0][IMP+FIX] website_sale_attribute_filter_category: collapsible

Tardo:
Adds new customize view to make all collapsible.

After some tries i think that the visual changes must be done in a new version. Changes will break themes, and perhaps other modules that have this dependency.

categ

cc @tecnativa TT25329


Tardo:
@chienandalu @sergio-teruel ping!

Tardo:
Ready to review

pedrobaeza:
The initial GIF is the final state?

chienandalu (commented):
> The initial GIF is the final state?

I just tested an indeed it is.

But why can't we get a similar aspect to the one in website categories?

sergio-teruel (commented):
What about extra space fix??

Tardo:
@sergio-teruel I think it's better to leave it for a new version... now requires changes that can break the visualization on themes (changes in nodes).

Tardo:
New gif in first comment.

pedrobaeza:
I still see a lot of indentation from initial collapsable title to the second one.

Tardo:
@pedrobaeza yes... fix it requires modify nodes. This change will break themes, etc... I think it is better to leave it for a major version change.
What do you think?

pedrobaeza:
I don't think so. Only play with the visibility, not with modifying nodes.

Tardo:
Changes done, gif updated.

sergio-teruel (approved):

sergio-teruel (changes requested):
Waiting fix travis...

Tardo:
Travis Green

sergio-teruel (approved):

PR #431, opened 2020-09-17
[12.0][FIX+IMP] website_sale_product_attribute_value_image: Move test tour JS to demo + Add 'alt' and 'title'

Tardo:
Add 'alt' and 'title' html attributes

cc  @tecnativa TT25329

### DIFF
--- a/website_sale_product_detail_attribute_value_image/templates/shop_product.xml
+++ b/website_sale_product_detail_attribute_value_image/templates/shop_product.xml
@@ -21,11 +21,12 @@
         </xpath>
         <!-- Add the attribute value images -->
         <xpath expr="//hr[@id='hr_product_attributes_simple']" position="after">
-            <hr t-if="sum([(1 if any(l.value_ids) else 0) for l in product._get_valid_product_template_attribute_lines()])" id="attr_value_image_separator" />
+            <hr t-if="attrs_published" id="attr_value_image_separator" />
         </xpath>
         <xpath expr="//t[@t-as='variant_id']" position="inside">
-            <div class="d-inline-flex flex-wrap mb-0 product-detail-attributes-values">
-                <t t-foreach="variant_id.value_ids.filtered(lambda x: x.website_product_detail_image_published)" t-as="attr_value">
+            <t t-set="attrs_published" t-value="variant_id.value_ids.filtered(lambda x: x.website_product_detail_image_published)" />
+            <div t-if="any(attrs_published)" class="d-inline-flex flex-wrap mb-0 product-detail-attributes-values">
+                <t t-foreach="attrs_published" t-as="attr_value">
                     <div class="flex-fill px-1">
                         <div class="text-center">
                             <img class="img" t-attf-src="/web/image?model=product.attribute.value&amp;field=website_product_detail_image&amp;id=#{attr_value.id}" t-att-title="attr_value.website_name or attr_value.name" t-att-alt="attr_value.website_name or attr_value.name"/>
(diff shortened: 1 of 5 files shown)


website_sale_product_detail_attribute_value_image/templates/shop_product.xml
```diff
@@ -28,10 +28,7 @@
                 <t t-foreach="variant_id.value_ids.filtered(lambda x: x.website_product_detail_image_published)" t-as="attr_value">
                     <div class="flex-fill px-1">
                         <div class="text-center">
-                            <span
-                                t-field="attr_value.website_product_detail_image"
-                                t-options="{'widget': 'image', 'resize': '40x40'}"
-                            />
+                            <img class="img" height="40" t-attf-src="/web/image?model=product.attribute.value&amp;field=website_product_detail_image&amp;id=#{attr_value.id}" t-att-title="attr_value.website_name or attr_value.name" t-att-alt="attr_value.website_name or attr_value.name"/>
```
pedrobaeza: why not max-height and max-width?

Tardo: You don't want the resize behaviour if the image height is less than 40px?

pedrobaeza: Resize? The problem I see with this is that an image of 1200x40 will break the layout.

Tardo: If only sets the "height" or "width" the image preserve the aspect ratio... but yeah, maybe better force the 40x40px to work as before.

pedrobaeza: That's why I told about max-widht:40 and max-height:40. With that, it adapts to the highest one, and the other is resized in browser preserving aspect ratio.

Tardo: I saw that if the same value is used "max-width" wins... so only need use "max-width" in this case. Thanks!

Tardo:
Changes done

pedrobaeza (approved):

website_sale_product_detail_attribute_value_image/templates/shop_product.xml
```diff
@@ -28,10 +28,7 @@
                 <t t-foreach="variant_id.value_ids.filtered(lambda x: x.website_product_detail_image_published)" t-as="attr_value">
                     <div class="flex-fill px-1">
                         <div class="text-center">
-                            <span
-                                t-field="attr_value.website_product_detail_image"
-                                t-options="{'widget': 'image', 'resize': '40x40'}"
-                            />
+                            <img class="img" t-attf-src="/web/image?model=product.attribute.value&amp;field=website_product_detail_image&amp;id=#{attr_value.id}" t-att-title="attr_value.website_name or attr_value.name" t-att-alt="attr_value.website_name or attr_value.name"/>
```
chienandalu, line 32: Why the website name? The normal would be the attribute name itself as first option, IMO

Tardo: This follows the behavior of 'website_sale_product_attribute_image' which also includes a "website_name" field. Normal is to only use the "name" field, but if you want to change the way they are displayed on the website, you can use "website_name"; in this case, the value of this field wins.

chienandalu: Sorry, you're right. It's ok!

chienandalu (approved):

PR #433, opened 2020-09-22
[12.0][ADD] website_sale_stock_provisioning_date: new module.

ernestotejeda:
Show the future provisioning date closest to the current date for a product in the eCommerce.

cc @Tecnativa TT24985


website_sale_stock_provisioning_date/models/product_product.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = 'product.product'
+
+    def _get_next_provisioning_date(self):
+        domain = [
+            ("company_id", "=", self.env.user.company_id.id),
```
pedrobaeza, line 11: Is this the same way as Odoo uses in the stock info popup for sales?

website_sale_stock_provisioning_date/models/product_product.py
```diff
@@ -0,0 +1,23 @@
...
+            ("location_id.usage", "=", "supplier"),
+            ("location_dest_id.usage", "=", "internal"),
+            ("date_expected", ">=", fields.Datetime.now()),
+        ]
+        move = (
+            self.env["stock.move"]
+                .sudo()
+                .search(domain, order="date_expected", limit=1)
+        )
+        return move.date_expected
```
pedrobaeza: Show only date. Time part is mostly irrelevant.
```suggestion
        return move.date_expected.date()
```

pedrobaeza (commented):
Tested on runbot. Seems OK except the comments.

website_sale_stock_provisioning_date/models/product_product.py
ernestotejeda: In the popup, the date that is shown is the expected date to deliver the products defined in the sale order.

pedrobaeza: Uhm, how about the expected one, how is computed?

ernestotejeda: But the expected date is information related to the sales order and what we want is information related to a product. In this case, the nearest future date the product is expected to be received. In the same sale order. In the same sales order we could have several products with different expected reception dates.

pedrobaeza: Yeah, but this computation is done taking the minimum expected date for each product. How that expected date is computed? We want the same algorithm (removing the customer lead time).

ernestotejeda: That is in v13 and as I see in the code it is almos the same as the date of the sale order.
In v12 there is effective_date, but what it calculates is the minimum date of the done pickings related to the sale_order.
 I don't think those fields are useful for what we want.

pedrobaeza: OK

pedrobaeza (approved):

website_sale_stock_provisioning_date/models/product_product.py
```diff
@@ -0,0 +1,23 @@
...
+            ("location_id.usage", "=", "supplier"),
+            ("location_dest_id.usage", "=", "internal"),
+            ("date_expected", ">=", fields.Datetime.now()),
+        ]
+        move = (
+            self.env["stock.move"]
+                .sudo()
+                .search(domain, order="date_expected", limit=1)
+        )
+        return move and move.date_expected.date() or False
```
pedrobaeza, line 23:
```suggestion
        return move.date_expected and move.date_expected.date() or False
```

website_sale_stock_provisioning_date/README.rst
```diff
@@ -0,0 +1,93 @@
...
+To use this module, you need to:
+
+#. Go to *Sales > Products > Products*, create a new product or edit an
+   existing one.
+#. Check 'Show Next Provisioning Date' field under eCommerce tab.
+#. Go to Website Shop.
+#. Click on a product that have 'Show Next Provisioning Date' checked.
+#. You will see the future provisioning date closest to the current date
+   for this product under the 'Add to Cart' button.
+
```
HaraldPanten: Just a picky comment: I would just add that it doesn't work with all of the "Availability" options that user can set in Ecommerce tab (in product form). It will work only with the ones that show the product quantity.

HaraldPanten (approved):
Functional review 👍

website_sale_stock_provisioning_date/README.rst
ernestotejeda: I also think that it would not make much sense to show that date if the stock availability is not also shown. But we don't want to force this functionality to be used only for specific availability options. I have modified the field help so that whoever is going to use this field takes this into account.

PR #443, opened 2020-10-07
[13.0][MIG] website_sale_require_legal: Migration to 13.0

ernestotejeda:
cc @Tecnativa TT25963


chienandalu (changes requested):
Tours are failing

website_sale_require_legal/controllers/main.py
```diff
@@ -0,0 +1,51 @@
...
+        return result
+
+    def checkout_form_validate(self, mode, all_form_values, data):
+        """Require accepting legal terms to validate form."""
+        # Patch context
+        old_context = request.context
+        request.context = dict(request.context, needs_legal=True)
+        result = super().checkout_form_validate(mode, all_form_values, data)
+        # Unpatch context
+        request.context = old_context
```
Tardo, line 34: I think that here you can use:
```suggestion
        result = super(WebsiteSale, self.with_context(needs_legal=True)).checkout_form_validate(mode, all_form_values, data)
```

website_sale_require_legal/views/assets.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
Tardo, line 1: Move tour to demo

Tardo (changes requested):

website_sale_require_legal/views/assets.xml
```diff
@@ -0,0 +1,13 @@
...
+<!-- Copyright 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="assets_frontend" inherit_id="website.assets_frontend">
```
Tardo:
```suggestion
    <template id="assets_tests" inherit_id="website.assets_tests">
```

Tardo (changes requested):

website_sale_require_legal/views/assets.xml
```diff
@@ -0,0 +1,13 @@
...
+        <xpath expr=".">
+            <script
+                type="text/javascript"
+                src="/website_sale_require_legal/static/src/js/tour.js"
```
Tardo: Move tour.js to a new location: /website_sale_require_legal/static/tests/tours/tour.js

Tardo (changes requested):

ernestotejeda:
@pedrobaeza , I have applied CSS to the label of the checkbox to be in red if it is not checked. I think that with that, we don't have to modify website_legal_page (https://github.com/OCA/website/pull/782) so the translations of this module will not be affected.

pedrobaeza:
Shouldn't the CSS be applied on parent module?

ernestotejeda:
I don't think so, In this case the selector of the css rule is based on the input that is created in this module

sergio-teruel:
Is needed this module??
You can customize this option in odoo core checkout step
https://github.com/odoo/odoo/blob/905fe5af69094051b9f7bf433a9c83cc0953fde1/addons/website_sale/views/templates.xml#L1489

ernestotejeda:
Yes, maybe this module is not necessary. Ping @pedrobaeza . Confirm this to me to do the corresponding changes in OpenUpgrade
Peek 2020-10-15 10-54

pedrobaeza:
Please show me screenshots of the existing feature and the module to compare

ernestotejeda:
This is how this module works:
Peek 2020-10-15 11-04

sergio-teruel:
I think that with the core function it's ok

chienandalu (commented):
The core function looks good, but we should check if metadata is still being registered (https://github.com/OCA/e-commerce/blob/fdec6c10e00803dabd7062ae0bdc527fdaf3f512/website_sale_require_legal/controllers/main.py#L37-L51) as it's a legal requirement in some countries.

sergio-teruel:
@chienandalu What do you propose?

chienandalu:
To check if it's covered. Otherwise, we should do it (maybe in other module)

ernestotejeda:
@chienandalu , I have reviewed and as I have seen, with the Odoo core functionality, this metadata is not sent either when the address user form is modified or when the payment is done, no matter if the checkbox of the terms and terms is checked.

pedrobaeza:
I think you haven't showed how standard works (or the video was for the standard?). And if so, how the other works? I want to compare? Anyway, that metadata should be saved, so this module can remain for at least that.

ernestotejeda:
- @pedrobaeza standard works: https://github.com/OCA/e-commerce/pull/443#issuecomment-709383153
- And this is how this module works: https://github.com/OCA/e-commerce/pull/443#issuecomment-709390063 , I just didn't show in the video that the metadata is recorded in the chatter of the partner associated with the user.

pedrobaeza:
Ok, thanks. Then let's keep this module for recording that metadata.

ernestotejeda:
@sergio-teruel @chienandalu @pedrobaeza Ready to review.

website_sale_require_legal/README.rst
```diff
@@ -0,0 +1,112 @@
...
+
+To use this module, you need to:
+
+* Buy something from your website.
+
+Known issues / Roadmap
+======================
+
+* Shopping terms and conditions are accepted only on user registration or
+  address edition. So if those terms change after the user signed up, a
```
sergio-teruel, line 85: With the new behavior, the checkbox is displayed in checkout step.

sergio-teruel (changes requested):
I have a doubt... 
We are going to use the core check but recording the metadata... Is that so?

pedrobaeza:
@ernestotejeda please check this

website_sale_require_legal/README.rst
ernestotejeda: @sergio-teruel , the Odoo (core) option shows the check box always just before paying, but what this point of the readme refers to is to show that box only if the terms have changed (perhaps to be shown before paying as well). I think they are two different things. And it seems to me that to cover that point (if we needed), we must add the functionality in this module, because the odoo option is not very configurable since it is a template designed to be activated and deactivated by the UI.

ernestotejeda: What do you think?

ernestotejeda:
@pedrobaeza @sergio-teruel @chienandalu 
What are we going to do with this PR?

HaraldPanten:
Hi,

Any news about this module? Finally we're going to keep it in OCA, right? @pedrobaeza

houssine78:
> 
> 
> Yes, maybe this module is not necessary. Ping @pedrobaeza . Confirm this to me to do the corresponding changes in OpenUpgrade
> Peek 2020-10-15 10-54

Hi, 

i was looking for this feature in v14.0 and it seems that it not standard... From my research the feature seems to come from this module https://apps.odoo.com/apps/modules/13.0/e_shop_term_and_condition/. I may be wrong but the printscreen of the screens under the description of the module tend to describe the exact feature.

Could we clarify this before concluding that it's in standard community ?

Thanks !

yajo:
Last commit applies all that has been discussed and tests it. Also, rebased. Please review again. Thanks!

yajo (approved):

pedrobaeza:
Travis is red

yajo:
Thanks, it should be fixed now.

chienandalu (approved):
:+1: Tested

PR #447, opened 2020-10-15
[FIX] website_sale_hide_price: Hide price from all combination info requests

Tardo:
Avoid show the price in other unexpected places on the website.

Now because we can't know the template where the price is displayed... this change will overwrite it to 0.00 (so, not really hidden)

@chienandalu @sergio-teruel 

cc @tecnativa TT26263

### DIFF
--- a/website_sale_hide_price/models/product_template.py
+++ b/website_sale_hide_price/models/product_template.py
@@ -1,0 +1,34 @@
+# Copyright 2020 Tecnativa - Alexandre D. Díaz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    @api.multi
+    def _get_combination_info(
+        self,
+        combination=False,
+        product_id=False,
+        add_qty=1,
+        pricelist=False,
+        parent_combination=False,
+        only_template=False
+    ):
+        """Avoid show prices in other website places.
+        This will show 0.00 value"""
+        ret = super()._get_combination_info(
+            combination=combination,
+            product_id=product_id,
+            add_qty=add_qty,
+            pricelist=pricelist,
+            parent_combination=parent_combination,
+            only_template=only_template)
+        if not self.env.user.partner_id.website_show_price:
+            ret.update(
+                list_price=0,
+                price=0,
+            )
+        return ret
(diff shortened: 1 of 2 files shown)


chienandalu (changes requested):
Tiny lint fix

website_sale_hide_price/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+        combination=False,
+        product_id=False,
+        add_qty=1,
+        pricelist=False,
+        parent_combination=False,
+        only_template=False
+    ):
+        """Avoid show prices in other website places.
+        This will show 0.00 value"""
+        ret = super()._get_combination_info(
```
sergio-teruel, line 22: Out of curiosity, why ret?

sergio-teruel (approved):

website_sale_hide_price/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+            product_id=product_id,
+            add_qty=add_qty,
+            pricelist=pricelist,
+            parent_combination=parent_combination,
+            only_template=only_template)
+        if not self.env.user.partner_id.website_show_price:
+            ret.update(
+                list_price=False,
+                price=False,
+            )
```
sergio-teruel, line 33: Why False instead of 0.00??

sergio-teruel (approved):

website_sale_hide_price/models/product_template.py
Tardo: I try to difference it from a legit 0.00 value.

website_sale_hide_price/models/product_template.py
Tardo: ret = return ... 
maybe better "res = result"?

sergio-teruel: ok ok ... thanks... I imagined that

victoralmau (approved):

website_sale_hide_price/models/product_template.py
```diff
@@ -0,0 +1,34 @@
...
+        ret = super()._get_combination_info(
+            combination=combination,
+            product_id=product_id,
+            add_qty=add_qty,
+            pricelist=pricelist,
+            parent_combination=parent_combination,
+            only_template=only_template)
+        if not self.env.user.partner_id.website_show_price:
+            ret.update(
+                list_price=False,
```
pedrobaeza: Never change the expected type for a variable. This can provoke later problems. At all effects, if a product has 0 price, it's the same as not be set. Change it to 0 both.

pedrobaeza (changes requested):

Tardo:
Changes done

pedrobaeza (approved):

PR #450, opened 2020-10-20
[MIG] website_sale_hide_price: Migration to 13.0

maitaoriana:


maitaoriana:
Hi @pedrobaeza, the migration is ready, please could you review it? Thanks

pedrobaeza:
I have added the PR to the migration issue, but I'm not a qualified reviewer.

tonygalmiche:
I failed to do this on runbot :
`If you wanted to have the prices hidden by default when no user is logged in you should go to Public User's partner and set Show prices on website off.`
I did this, but the prices are still displayed

tonygalmiche (approved):
LGTM : Functional review only with runbot

image

chienandalu (approved):
Looks like the compare views still show the price
image

pedrobaeza:
@chienandalu then it's not an approval?

chienandalu (changes requested):
Oh, sorry

maitaoriana:
@chienandalu Thanks by you review

I think that to hide the price on the comparator is not the responsibility of this module because it would need to depend on website_sale_comparison. Besides on v 12 this module doesn't hide the prices on the comparator. [Video](https://drive.google.com/file/d/1STlgTo0N3fY1xMuymn09r7q_2rqy7M1V/view)

To add this new feature would be good to create a new module that depends on this module and website_sale_comparison module.

chienandalu (approved):
Ok, great :+1:

PR #454, opened 2020-10-26
[12.0][MIG] website_sale_order_type: Migration to 12.0 

joao-p-marques:
@Tecnativa
TT26399

ping @pedrobaeza @rafaelbn


website_sale_order_type/models/sale_order.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    @api.multi
+    def _cart_update(
```
chienandalu, line 11: I'd say it would be better to use this prepare https://github.com/odoo/odoo/blob/12.0/addons/website_sale/models/website.py#L193-L219

Otherwise, your triggering the onchange on every cart update, which doesn't seem very optimum and could change some prefrences of the checkout like the shipping address...

joao-p-marques: Ok, thank you. I hadn't thought of the repercussions of using this onchange method. The only change I want to make is to add the SO type, and I didn't consider that all the other fields would be updated as well.

I pushed an update following your suggestion. From what I have tested, the `_prepare_sale_order_values` method is not updating the values on the sale order as with the `_cart_update` method. What else would I need to do?

chienandalu: Mmm, but the tests go fine, aren't they?

joao-p-marques: With the current method, yes, not with `_prepare_sale_order_values`

website_sale_order_type/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "category": "Website",
+    "website": "https://github.com/OCA/e-commerce/",
+    "maintainers": ["joao-p-marques"],
+    "author": "Agile Business Group, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": [
+        "website_sale",
+        "sale_order_type"
+    ],
+    "data": [
```
yajo:
```suggestion
    "demo": [
```

At least until v13, this improves performance.

joao-p-marques: Thank you, changes made!

victoralmau:
Please, check Travis

joao-p-marques:
Travis is green now @victoralmau Thank you :+1:

victoralmau (approved):

website_sale_order_type/views/assets.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <template id="assets_frontend" inherit_id="website_sale.assets_frontend">
+        <xpath expr=".">
+            <script type="text/javascript"
+                    src="/website_sale_order_type/static/src/js/website_sale_order_type_tour.js"/>
```
pedrobaeza: Why is this regular data instead of demo one? Put also a note for being moved to `assets_tests` + data on v13.

joao-p-marques: Thanks. Changes made! Was that what you meant?

pedrobaeza: It's still loaded as data instead of demo. In manifest, put the XML in demo key and move the file to demo folder.

joao-p-marques: I think it was already like that: https://github.com/OCA/e-commerce/pull/454/files#diff-c638094d29e2fd858515ae4b86befdf3258a8b235bfad3842845768b76d1519cR18-R20
Right?

joao-p-marques: I just need to move it to the `demo` folder

joao-p-marques: Change done @pedrobaeza

pedrobaeza (approved):

PR #456, opened 2020-10-29
[14.0][MIG] website_sale_attribute_filter_category > website_sale_product_attribute_filter_category

ivantodorovich:
Standard migration + the following changes:

1. https://github.com/OCA/e-commerce/commit/b57b9b6ac6bec1f3461010f5fb2fc4194163bb6d Rename module from `website_sale_attribute_filter_category` to `website_sale_product_attribute_filter_category`.

This is because there are many modules related to the attribute filters in this repository, and they don't share the same naming convention. REF: https://github.com/OCA/e-commerce/issues/441#issuecomment-706189979

2. https://github.com/OCA/e-commerce/commit/b5a6da604dd80e9a5b1c2533cb79168aa2873803 Add customize_show to the view so that in a multi-website deployment the user can activate/deactivate this feature on each website

3. https://github.com/OCA/e-commerce/pull/456/commits/e732bcc921d79048adb2c4a1c4cd4a650c81d67e Use a style similar to the eCommerce Categories list. Picture below

filters_by_category

ping @sergio-teruel @CarlosRoca13


ivantodorovich:
@OCA/community-maintainers any idea why pre-commit is failing in travis here? It's passing on my local

lfreeke (approved):
LGTM :+1: Functional test

claudiagn (approved):
LGTM! functional review

CarlosRoca13:
Please, squash the administrative commits, and set your migration into 2 commits:

1. [IMP] black, isort, prettier : pre-commit run -a
2. [MIG] website_sale_attribute_filter_category: Migration to 14.0

The other commits are part of the migration and just create noise

CarlosRoca13 (changes requested):
Please, squash the administrative commits, and set your migration into 2 commits:

1. [IMP] black, isort, prettier : pre-commit run -a
2. [MIG] website_sale_attribute_filter_category: Migration to 14.0

The other commits are part of the migration and just create noise

ivantodorovich:
> Please, squash the administrative commits, and set your migration into 2 commits:
> 
> 1. [IMP] black, isort, prettier : pre-commit run -a
> 2. [MIG] website_sale_attribute_filter_category: Migration to 14.0
> 
> The other commits are part of the migration and just create noise

It's done.

chienandalu (approved):

website_sale_product_attribute_filter_category/i18n/fi.po
```diff
@@ -0,0 +1,55 @@
...
+#~ msgstr "Testi"
+
+#~ msgid "Test Category"
+#~ msgstr "Testi Kategoria"
+
+#~ msgid "Test v1"
+#~ msgstr "Testi v1"
+
+#~ msgid "Test v2"
+#~ msgstr "Testi v2"
```
CarlosRoca13, line 55: Delete this
```suggestion
```

CarlosRoca13 (approved):
LGTM :smile: Just minor changes on translations (non blocking)

pedrobaeza:
You didn't rename the old commits, and thus, the commit history was lost. I have just launched the merge with the old name, but I'm aborting. I'm going to check what to do.

ivantodorovich:
Uhm. Sorry about that.

I didn't know we had to rename old commits, nor understand exactly what you mean by that TBH 😓 

It seems a bit weird having to alter old commits, though. The history wouldn't be real.

I think the real issue here is the fact that I ended up squashing the migration commits (https://github.com/OCA/e-commerce/pull/456#issuecomment-811071741). So, since the commit diff is more than just a rename (it contains other changes within them), git treats them as deleted/new files, and thus the history per file is lost. Otherwise, if the rename would have been kept as a separate commit, git would just recognize it as such and keep the history.

Something to consider for the future I guess. Sorry about this.

pedrobaeza:
OK, don't worry. Fixed in #727. You alter in a similar way things when fetching the commit history, as the resulting commits are not the original ones with the same hash (having a different parent commit). We apply this command to avoid losing part of these commits in the next iteration:

```
git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
```

PR #458, opened 2020-10-29
[14.0][ADD] website_sale_product_attribute_filter_count

ivantodorovich:
This module extends the functionality of website sale module to show the number of products next to each attribute filter.

count

It also adds an option to hide unavailable attribute values from the list, which is a feature that's also offered by `13.0` module [website_sale_product_attribute_value_filter_existing](https://github.com/OCA/e-commerce/tree/13.0/website_sale_product_attribute_value_filter_existing). I believe this module can superseed that one, that's why I added previous module contributors to `CONTRIBUTORS.rst`.

config

ping @sergio-teruel @CarlosRoca13


ioans73:
Functional review ok

ioans73 (approved):

CarlosRoca13:
Thanks for the attribution, but I think that with this fusion we are mixing two different things on a module... 

In my opinion the best way is to maintain two different modules, each one with its functionality.

pedrobaeza:
I agree they don't fit together.

ivantodorovich:
From a technical point of view, filtering on existing attr values means simply filterting values where their product count is `> 0`.

So, provided that the counts are computed, implementing the `value_filter_existing` feature requires only a very simple qweb view: https://github.com/OCA/e-commerce/pull/458/files#diff-a71cc82148595fc019de42dc35b0433a40a30962d8c8466c501c6389f10b48ddR40

By merging both features into a single module, the backend code to maintain is reduced, as it's kind of redundant

Having said that, I won't of course oppose a general consensus, but I urge you to consider that reasoning

pedrobaeza:
I understand your reasoning, but not sure if your count computation is heavier than previous one done only for hiding attributes. Can you tell us?

And if similar, we can have the other existing module depending on this one, but split.

ivantodorovich:
Here's a quick comparisson on a database with:
- 24332 products
- 718 attribute values
- 11208 product template attribute lines

To compute counts on this module I'm using raw sql:

```
14-0-1661873=> select count(*) from product_template;
 count 
-------
 24322
(1 row)

14-0-1661873=> explain analyze SELECT
    ptav.product_attribute_value_id,
    count(distinct pp.product_tmpl_id)
FROM product_product pp
INNER JOIN product_variant_combination pvc
    ON pvc.product_product_id = pp.id
INNER JOIN product_template_attribute_value ptav
    ON ptav.id = pvc.product_template_attribute_value_id
GROUP BY ptav.product_attribute_value_id;
                                                                           QUERY PLAN                                                                            
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=6282.47..6656.92 rows=680 width=12) (actual time=58.905..77.506 rows=672 loops=1)
   Group Key: ptav.product_attribute_value_id
   ->  Sort  (cost=6282.47..6405.02 rows=49020 width=8) (actual time=58.873..65.183 rows=49020 loops=1)
         Sort Key: ptav.product_attribute_value_id
         Sort Method: quicksort  Memory: 3834kB
         ->  Hash Join  (cost=1498.93..2463.54 rows=49020 width=8) (actual time=14.357..46.885 rows=49020 loops=1)
               Hash Cond: (pvc.product_template_attribute_value_id = ptav.id)
               ->  Hash Join  (cost=1043.37..1879.27 rows=49020 width=8) (actual time=9.754..29.404 rows=49020 loops=1)
                     Hash Cond: (pvc.product_product_id = pp.id)
                     ->  Seq Scan on product_variant_combination pvc  (cost=0.00..707.20 rows=49020 width=8) (actual time=0.013..4.416 rows=49020 loops=1)
                     ->  Hash  (cost=682.61..682.61 rows=28861 width=8) (actual time=9.683..9.683 rows=28861 loops=1)
                           Buckets: 32768  Batches: 1  Memory Usage: 1384kB
                           ->  Seq Scan on product_product pp  (cost=0.00..682.61 rows=28861 width=8) (actual time=0.006..5.589 rows=28861 loops=1)
               ->  Hash  (cost=276.36..276.36 rows=14336 width=8) (actual time=4.571..4.571 rows=14336 loops=1)
                     Buckets: 16384  Batches: 1  Memory Usage: 688kB
                     ->  Seq Scan on product_template_attribute_value ptav  (cost=0.00..276.36 rows=14336 width=8) (actual time=0.011..2.474 rows=14336 loops=1)
 Planning time: 0.636 ms
 Execution time: 77.961 ms
```

----

To compute attribute existing, the previous module was using ORM:

```py
templates = env["product.template"].search([])                                                                                                                                                                                                                                                                                                                   

import time                                                                                                                                                                                                                                                                                                                                                      
start = time.time()

# mimick filter_existing
templates = env["product.template"].search([], limit=False)
ProductTemplateAttributeLine = env["product.template.attribute.line"]
attribute_values = ProductTemplateAttributeLine.search([("product_tmpl_id", "in", templates.ids)])
attr_values_used = attribute_values.mapped("value_ids")
end = time.time()
print('time: %s' % (end - start))
                                                
```

```
time: 0.32630491256713867
# 326 ms
```

pedrobaeza (dismissed):
Ouch, are you using plain SQL? Then definitively I don't want this. Raw SQL must be avoided almost always. It's ignoring business layer logic and ORM framework, and can lead to inconsistencies in the results. Please do it on ORM in the most optimized way.

ivantodorovich:
It was my first attempt. However the ORM is not able to compute this efficiently enough for e-commerce.
It is using raw sql but it's filtering on the product ids that where previously fetched by the ORM.
No business logic that matter are being bypassed AFAIK.

See lines:
https://github.com/OCA/e-commerce/pull/458/files#diff-bef0df12f117a94386984a78e73927f4931c9d7ab872903d1eb1bc8c3a2ee748R31
and https://github.com/OCA/e-commerce/pull/458/files#diff-bef0df12f117a94386984a78e73927f4931c9d7ab872903d1eb1bc8c3a2ee748R54

pedrobaeza:
Yes if you have record rules, or other modules on top.

ivantodorovich:
As I said, record rules are not ignored.
The sql is filtering on the product and attribute ids retrieved by the ORM.

It even plays along with modules like: https://github.com/OCA/e-commerce/tree/13.0/website_sale_attribute_filter_price

pedrobaeza:
Well, I leave you to continue if you want with that approach, but we are going to keep the other module, as in my experience, this approach is conflicting sooner than later and we have remove 100% of direct SQL queries in our modules.

PR #460, opened 2020-10-30
[12.0] [FIX] website_sale_product_minimal_price: Bug with formulas

CarlosRoca13:
Fixed a bug that compute allways the formula prices from the price. Now the price take care about the `base` field on pricelist item ids.
I add an improvement to take care about the categies.

cc @Tecnativa TT24954

@carlosdauden @chienandalu this is the solution of the bug mentioned

### DIFF
--- a/website_sale_product_minimal_price/controllers/main.py
+++ b/website_sale_product_minimal_price/controllers/main.py
@@ -23,39 +23,54 @@ class WebsiteSaleProductDetailAttributeImage(WebsiteSale):
         today = fields.Date.today()
         res = []
         pricelist = current_website.get_current_pricelist()
-        pricelist_variant_items = pricelist.item_ids.filtered(
+        pricelist_items = pricelist.item_ids.filtered(
             lambda i: i.product_id.id == product.id
             and (not i.date_start or i.date_start <= today)
             and (not i.date_end or today <= i.date_end)
             and i.min_quantity > 0)
-        if pricelist_variant_items:
-            item_min_qty = min(pricelist_variant_items, key=lambda i: i.min_quantity)
+        if pricelist_items:
+            item_min_qty = min(pricelist_items, key=lambda i: i.min_quantity)
             min_qty = item_min_qty.min_quantity
-            pricelist_tmpl_items = pricelist.item_ids.filtered(
+            pricelist_items = pricelist_items + pricelist.item_ids.filtered(
                 lambda i: i.product_tmpl_id.id == product.product_tmpl_id.id
                 and i.min_quantity < min_qty
                 and (not i.date_start or i.date_start <= today)
                 and (not i.date_end or today <= i.date_end)
                 and i.min_quantity > 0)
         else:
-            pricelist_tmpl_items = pricelist.item_ids.filtered(
+            pricelist_items = pricelist_items + pricelist.item_ids.filtered(
                 lambda i: i.product_tmpl_id.id == product.product_tmpl_id.id
                 and (not i.date_start or i.date_start <= today)
                 and (not i.date_end or today <= i.date_end)
                 and i.min_quantity > 0)
-        self._prepare_dictionary(pricelist_variant_items, product, res)
-        self._prepare_dictionary(pricelist_tmpl_items, product, res)
+        if pricelist_items:
+            item_min_qty = min(pricelist_items, key=lambda i: i.min_quantity)
+            min_qty = item_min_qty.min_quantity
+            pricelist_items = pricelist_items + pricelist.item_ids.filtered(
+                lambda i: i.categ_id.id == product.categ_id.id
+                and i.min_quantity < min_qty
+                and (not i.date_start or i.date_start <= today)
+                and (not i.date_end or today <= i.date_end)
+                and i.min_quantity > 0)
+        else:
+            pricelist_items = pricelist_items + pricelist.item_ids.filtered(
+                lambda i: i.categ_id.id == product.categ_id.id
+                and (not i.date_start or i.date_start <= today)
+                and (not i.date_end or today <= i.date_end)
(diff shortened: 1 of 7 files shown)


sergio-teruel (approved):
Check travis...
Tested on runbot

pedrobaeza:
@CarlosRoca13 @sergio-teruel check the new algorithm for the grid view prices and the selection of the combination with lowest price. I haven't touched the scaled price table though.

pedrobaeza:
Is this still WIP?

CarlosRoca13:
No sorry, ready to be merged

website_sale_product_minimal_price/controllers/main.py
```diff
@@ -23,55 +22,25 @@ def get_combination_info_pricelist_atributes(
         today = fields.Date.today()
         res = []
         pricelist = current_website.get_current_pricelist()
-        pricelist_variant_items = pricelist.item_ids.filtered(
+        pricelist_items = pricelist.item_ids.filtered(
             lambda i: i.product_id.id == product.id
+            or i.product_tmpl_id.id == product.product_tmpl_id.id
+            or i.categ_id.id == product.categ_id.id
```
pedrobaeza: On pricelist items, an upper category in the hierarchy includes all children categories, so this is not correct at all.

website_sale_product_minimal_price/controllers/main.py
```diff
@@ -23,55 +22,25 @@ def get_combination_info_pricelist_atributes(
         today = fields.Date.today()
         res = []
         pricelist = current_website.get_current_pricelist()
-        pricelist_variant_items = pricelist.item_ids.filtered(
+        pricelist_items = pricelist.item_ids.filtered(
             lambda i: i.product_id.id == product.id
```
pedrobaeza: No need to compare with `.id`. `i.product_id == product.product_id` is the same

website_sale_product_minimal_price/controllers/main.py
```diff
@@ -23,55 +22,25 @@ def get_combination_info_pricelist_atributes(
...
+            or i.product_tmpl_id.id == product.product_tmpl_id.id
```
pedrobaeza: I think the order is relevant here, as having a rule for template and for variant makes one of them invalid, and you are not taking this into account here.

pedrobaeza (changes requested):

website_sale_product_minimal_price/controllers/main.py
CarlosRoca13: I take care about that on `_prepare_dictionary` method.

pedrobaeza: I'm still not seeing any difference on this matter.

Imagine this pricelist:

- Template with min_qty = 1 - Price 10
- Template with min_qty = 30 - Price 5
- Variant with min_qty = 1 - Price 11
- Variant with min_qty = 5  - Price 10
- Variant with min_qty = 40  - Price 6

My guess is that this is going to paint

>=1 - 11 €
>=1 - 10 €
>=5 - 10 €
>=30 - 5 €
>=40 - 6 €

sergio-teruel (approved):
Ready for merge??

CarlosRoca13:
I solved what @pedrobaeza said and I think that it is ready to be merged.

website_sale_product_minimal_price/models/product_template.py
```diff
@@ -1,120 +1,81 @@
...
+        """Helper method for getting the variant with lowest price."""
+        # TODO: Cache this method for getting better performance
+        self.ensure_one()
+        context = dict(self.env.context, pricelist=pricelist.id)
+        min_price = math.inf
+        product_id = False
+        add_qty = 0
+        has_distinct_price = False
+        for product in self.product_variant_ids:
+            for qty in [1, math.inf]:
```
sergio-teruel: @CarlosRoca13 I think that the float_compare method crash with inf... @pedrobaeza 2 days ago changed inf to 99999...

website_sale_product_minimal_price/models/product_template.py
```diff
@@ -1,120 +1,96 @@
...
+            pricelist = self.env["product.pricelist"].browse(
+                context["pricelist"])
+            product_id = self._get_cheapest_info(pricelist)[0]
+            product = self.env["product.product"].browse(product_id)
+            ptavs = product.product_template_attribute_value_ids
+            variant_attributes = ptavs.mapped("attribute_id")
+            # remove returned values that are variant specific
+            res.filtered(lambda x: x.attribute_id not in variant_attributes)
+            # and inject cheapest variant ones
+            res += ptavs
```
pedrobaeza, line 96: This is not being sorted. See https://github.com/OCA/e-commerce/pull/460/files#diff-4df29720f5325f6a08d6648efb0e935de9b9cf38bc79e27087fb4c3e8f958c69L118

Why previous code was not correct (filtering if you want if pricelist is present).

chienandalu (commented):
@CarlosRoca13 what's to do here?

chienandalu:
What's left to do here. This has been in production for months isn't it?

pedrobaeza:
My comment is still unattended. The sort thing is incorrect, although it's difficult that somebody is aware of it.

CarlosRoca13:
@pedrobaeza That changes that you say, were added by you, and I'm seeing that on the product sheet the order is correct. I found a problem on products grid that don't take the minimal value, but is not produced by that fragment of code. I'm working for solve it as soon as possible.

pedrobaeza:
OK, I see. In some point, my code was not used, but it's not the case now.

CarlosRoca13:
@pedrobaeza I think that the problem that you mean is solved 🥳

pedrobaeza (approved):

PR #461, opened 2020-11-04
[12.0][IMP] website_snippet_product_category: Avoid use custom templates from javascript

Tardo:
cc @tecnativa

@Yajo

### DIFF
--- a/website_snippet_product_category/controllers/website.py
+++ b/website_snippet_product_category/controllers/website.py
@@ -13,12 +13,14 @@ class Website(http.Controller):
         auth="public",
         website=True,
     )
-    def render_product_category(self, template, **kwargs):
+    def render_product_category(self, **kwargs):
         categories = request.env["product.public.category"].search(
             [("parent_id", "=", False), ("website_published", "=", True)]
         )
         keep = QueryURL("/shop", category=0)
-        return request.website.viewref(template).render({
+        return request.website.viewref(
+            "website_snippet_product_category.s_product_category_items"
+        ).render({
             "object": categories,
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Travis is red

website_snippet_product_category/controllers/website.py
```diff
@@ -13,12 +13,14 @@ class Website(http.Controller):
...
             [("parent_id", "=", False), ("website_published", "=", True)]
         )
         keep = QueryURL("/shop", category=0)
-        return request.website.viewref(template).render({
+        return request.website.viewref(
+            "website_snippet_product_category.s_product_category_items"
+        ).render({
             "object": categories,
-            "keep": keep
+             "keep": keep
```
victoralmau, line 25:
```suggestion
            "keep": keep
```
Please, remove space

victoralmau (changes requested):

website_snippet_product_category/controllers/website.py
```diff
@@ -13,12 +13,14 @@ class Website(http.Controller):
         auth="public",
         website=True,
     )
-    def render_product_category(self, template, **kwargs):
+    def render_product_category(self, **kwargs):
```
yajo, line 16: This will break downstream extensions.

I think what we need here is a whitelist of allowed templates, and a 403 if the template isn't in the list.

yajo (changes requested):

pedrobaeza:
@Tardo is this still applicable for v13?

PR #463, opened 2020-11-06
[11.0][ADD] website_sale_payment_mode

p-tombez:
This is a glue module between `website_sale` and `account_payment_sale` in order to fill the payment mode field with the value configured on the partner when creating sale orders from the website


yvaucher (approved):

website_sale_payment_mode/models/website.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import api, models
+
+
+class Website(models.Model):
+    _inherit = 'website'
+
+    @api.multi
+    def _prepare_sale_order_values(self, partner, pricelist):
+        values = super()._prepare_sale_order_values(partner, pricelist)
+        values['payment_mode_id'] = partner.customer_payment_mode_id.id
```
chienandalu: Shouldn't this be the commercial entity payment mode?:
```suggestion
        values['payment_mode_id'] = partner.commercial_partner_id.customer_payment_mode_id.id
```

p-tombez: Yes, that would be better => Fixed

i-vyshnevska (approved):

chienandalu (approved):

PR #470, opened 2020-11-18
[FIX] website_sale_stock_force_block: prevent add to cart

chienandalu:
In core odoo there's a mechanism to prevent customers adding products to
a cart when there's no stock that relies on the `inventory_availability`
type. We must apply it as well for this new type.

cc @Tecnativa TT26832

### DIFF
--- a/website_sale_stock_force_block/models/__init__.py
+++ b/website_sale_stock_force_block/models/__init__.py
@@ -1,2 +1,3 @@
 from . import product_template
 from . import res_config_settings
+from . import sale_order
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Aren't we able to do it other cleaner way?

chienandalu:
> Aren't we able to do it other cleaner way?

If we could fool the cache somehow when this is evaluated: https://github.com/odoo/odoo/blob/12.0/addons/website_sale_stock/models/sale_order.py#L18

pedrobaeza:
What about my proposal of splitting the field in 2?

pedrobaeza (approved):

sergio-teruel:
In which case do you need to block this?

chienandalu:
> In which case do you need to block this?

Well, as you know this module adds a stock availability option that prevents sales if there isn't stock. But when the user has the item in the cart, this is the logic that constrains adding more units from there.

sergio-teruel:
oK @chienandalu

sergio-teruel (approved):

PR #415, opened 2020-07-25
Delete EN374:2003.png

cormaza:
FIX: for windows enviroments

PR #473, opened 2020-11-20
FIX: name of png file for windows so

cormaza:
Fix name EN374:2003.png


pedrobaeza:
Rename it instead of deleting it. And not sure if such image is referenced in any place. Why don't you cherry-pick the commit that does this in 12.0 instead?

cormaza:
From what I saw, there is no direct reference to the name of the file, and I do not delete it because in another pr that I did some time ago they told me not to delete it but to rename it

PR #415

PR #382 on version 12

pedrobaeza:
Yes, and this PR is again the same deleting it. You have forgotten to re-add the files to the git commit.

PR #478, opened 2020-12-10
[FIX] website_sale_stock_available: prevent payment blocking

chienandalu:
When the `virtual_available` is less than the `immediately_usable_qty`,
we'll have a blocking situation when the order is paid if we could add
more quantity to the order than the virtual one.

cc @Tecnativa TT27214


chienandalu:
@sergio-teruel @pedrobaeza please review

sergio-teruel (approved):
Code review

HaraldPanten (approved):

PR #479, opened 2020-12-16
[12.0][FIX] website_sale_attribute_filter_price: Use env context

### LINKED ISSUE #476: [12.0] website_sale_attribute_filter_price causes context overwriting
Hello,

using other addon (`website_sale_product_brand`) I've experienced the following issue: clicking on a certain brand from the frontend (`/page/product_brands`) redirects to the shop, but the brand filter is not applied.

Debugging the the code I've found that brand value is registered in the `request.env.context` (https://github.com/OCA/e-commerce/blob/12.0/website_sale_product_brand/controllers/main.py#L31), but when it's intended to be used for the domain search construction (https://github.com/OCA/e-commerce/blob/12.0/website_sale_product_brand/controllers/main.py#L14) `request.env.context` has already lost this data.

Looking for other addons that could cause this issue, I've uninstalled `website_sale_attribute_filter_price`, which I was using too, and the issue is gone. When executing step-by-step, after this line (https://github.com/OCA/e-commerce/blob/12.0/website_sale_attribute_filter_price/controllers/website_sale.py#L45) `brand_id` context value is lost. I'm newbie in e-commerce app but, are `request.context` references correct, or should be `request.env.context` used instead?

Tried to test on v12 OCA e-commerce runbot, but `admin` password for `admin` user didn't work for me today.

cc @Tardo @chienandalu

Tardo:
Fix #476 

cc @tecnativa

### DIFF
--- a/website_sale_attribute_filter_price/controllers/website_sale.py
+++ b/website_sale_attribute_filter_price/controllers/website_sale.py
@@ -10,7 +10,7 @@ from odoo.addons.website.controllers.main import QueryURL
 class WebsiteSale(WebsiteSale):
     def _get_search_domain(self, search, category, attrib_values):
         domain = super()._get_search_domain(search, category, attrib_values)
-        price_vals = request.context.get("price_vals")
+        price_vals = request.env.context.get("price_vals")
         if price_vals:
             to_add = []
             if price_vals[0] is not None:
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

dalonsod (approved):
Tested on runbot, now it works 👍

PR #483, opened 2021-01-07
[IMP] Add product links smartbutton

ajaniszewska-dev:
aim of this PR is to add button with links to product.product form view.

### DIFF
--- a/product_variant_multi_link/models/product_product.py
+++ b/product_variant_multi_link/models/product_product.py
@@ -2,7 +2,8 @@
 # Simone Orsi <simahawk@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import AccessError
 
 
 class ProductProduct(models.Model):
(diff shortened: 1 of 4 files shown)


simahawk (changes requested):
LG but please add a test. There's one already on template.

product_variant_multi_link/models/product_product.py
```diff
@@ -14,6 +15,47 @@ class ProductProduct(models.Model):
...
+
+    @api.depends("product_variant_link_ids")
+    def _compute_product_product_link_count(self):
+        link_model = self.env["product.template.link"]
+        # Set product_template_link_qty to 0 if user has no access on the model
+        try:
+            link_model.check_access_rights("read")
+        except AccessError:
+            for rec in self:
+                rec.product_product_link_count = 0
```
i-vyshnevska:
```suggestion
            self.updae({'product_product_link_count':0})
```

product_variant_multi_link/models/product_product.py
```diff
@@ -14,6 +15,47 @@ class ProductProduct(models.Model):
...
+    product_product_link_count = fields.Integer(
+        string="Product Links Count", compute="_compute_product_product_link_count"
+    )
+
+    @api.depends("product_variant_link_ids")
+    def _compute_product_product_link_count(self):
+        link_model = self.env["product.template.link"]
+        # Set product_template_link_qty to 0 if user has no access on the model
+        try:
+            link_model.check_access_rights("read")
```
i-vyshnevska, line 27: or you can add a group right on xml view so for the rest of the users button will be invisible

simahawk: then if you use the same field in another view or piece of code you won't have the same result

i-vyshnevska (approved):

product_variant_multi_link/models/product_product.py
```diff
@@ -14,6 +15,46 @@ class ProductProduct(models.Model):
         compute="_compute_product_link_ids",
     )
 
+    product_product_link_count = fields.Integer(
+        string="Product Links Count", compute="_compute_product_product_link_count"
```
sebalix: Runbot is complaining about this label:

    Two fields (product_template_link_count, product_product_link_count) of product.product() have the same label: Product Links Count.

Can you handle it? Maybe "Variant Links Count"? (I don't know the context)

ajaniszewska-dev: should be fine now.

simahawk (approved):

PR #486, opened 2021-01-19
[14.0][ADD] website_sale_attribute_filter_range

miguel-S73:
This module adds a filter range in the website sale. 
Take the attribute values included between the selected range, and applies the filter.

In product detail page, attributes of type 'range' are shown the same as attributes of type 'select'.
We have chosen this option because on this page it does not make sense for the user to choose a range
and because this option will have too many values to display in a 'radius' type. 
We opted to copy the selects functionality in the product view to avoid problems if another module modifies the 't-if'.

rangeselect


miguel-S73:
@pedrobaeza ,
¿Could you please review?

pedrobaeza (commented):
Interesting module. Thanks for the contribution. What do you think about depending on `website_sale_attribute_filter_price` for not having the same library loaded twice? Or create an intermediate module for the library and put both depending on it? This second one is not the ideal, as we are modifying the dependencies inside a stable version, but I think it's worth for the memory footprint/load time.

PR #487, opened 2021-01-20
[14.0][ADD] website_sale_attribute_filter_form_submit

miguel-S73:
This module hides all the submit elements of the attribute filter form,block submitting the form in the 'onchange' attributes values and it add a submit button at the end of the filters.

attribute_filter_submit


chienandalu (commented):
Testing in runbot, I see a problem selecting the color attributes. I think it could be achieved with bare bootstrap though. I similar case I resolved that way: https://github.com/OCA/rma/blob/12.0/rma_sale/views/sale_portal_template.xml#L45-L53

chienandalu:
Amending my comment, I tried some options:

- Removing the opacity of the checkbox input: easiest but ugly
- Adding an overlay via css that changes depending on the checked selector (something like [this](https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_custom_checkbox)): tricky and unreliable
- Styling the label (which is the one holding the color swatch style) via js: reliable and medium

So maybe we should go for the last one

miguel-S73:
Ok, I will try it

chienandalu:
> Ok, I will try it

@miguels73 take a look at #535 at last it was quite simple

chienandalu:
@miguels73 #535 is already merged can you adapt the changes from https://github.com/OCA/e-commerce/pull/535/commits/4b3f896295e0b4b1621142f644f03eedc9af5374 to your PR?

miguel-S73:
Hi @chienandalu , I have tested your changes and it works fine but I can't cherry-pick your changes because commit brings more changes. Do you know any way to do it? Or do I commit and make you a contributor?

miguel-S73:
@chienandalu , I have already applied your changes.

chienandalu (approved):
Thanks @miguels73 :)

pedrobaeza (approved):

PR #501, opened 2021-02-18
[13.0][IMP] Allow to select variant first, on product links

mmequignon:
Currently on product_variant_multi_link, variants can be selected
only when the template is set.
This PR allows to select the variant first.


simahawk (commented):
tests?

jgrandguillaume (approved):
I would go through now. Functional test are ok

PR #503, opened 2021-02-18
[13.0] [MIG] website_sale_product_sort: Migration to v13.0

CarlosRoca13:
cc @Tecnativa TT26956

please @chienandalu @victoralmau review this


chienandalu (changes requested):
Check travis

victoralmau (approved):

CarlosRoca13:
@chienandalu Travis green yet ;)

website_sale_product_sort/models/website.py
```diff
@@ -21,6 +21,6 @@ def _get_product_sort_criterias(self):
         selection="_get_product_sort_criterias",
         string="Sort products by",
         help="Default criteria for sorting products in the shop",
-        default="website_sequence desc",
+        default="website_sequence asc",
```
chienandalu, line 24: This is important and I saw it isn't considered in OpenUpgrade for reversing the current ids :thinking:

chienandalu (approved):
Thanks. A comment to take into account:

website_sale_product_sort/models/website.py
victoralmau: I think this change is related to this: https://github.com/odoo/odoo/blob/13.0/addons/website_sale/controllers/main.py#L154

chienandalu: Yes, sure, this is the commit. The migration is right (I approved :slightly_smiling_face: ). I'm just commenting that a OpenUpgrade migration script would be needed additionaly to avoid issues in migrations.

pedrobaeza:
Please check our use cases and if we don't need this module anymore then.

PR #511, opened 2021-03-04
[FIX][website_sale_product_brand] Missing ppg parameter in shop method inheritance

jctrey:
The "shop" method inheritance of the "website_sale" module is incorrect, since it omits the "ppg" parameter both in the "header" and in the call to "super".

### DIFF
--- a/website_sale_product_brand/controllers/main.py
+++ b/website_sale_product_brand/controllers/main.py
@@ -25,14 +25,14 @@ class WebsiteSale(WebsiteSale):
                 type='http',
                 auth='public',
                 website=True)
-    def shop(self, page=0, category=None, brand=None, search='', **post):
+    def shop(self, page=0, category=None, search='', ppg=False, brand=None, **post):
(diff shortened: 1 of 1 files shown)


website_sale_product_brand/__manifest__.py
```diff
@@ -8,7 +8,7 @@
               "Tecnativa, "
               "Odoo Community Association (OCA)",
     'website': 'https://github.com/OCA/e-commerce',
-    'version': '12.0.1.3.0',
+    'version': '12.0.1.3.1',
```
chienandalu: No need to bump manually, the OCA bot will do it on merge

chienandalu (commented):
Ok, just a little change:

pedrobaeza:
Please remove latest commit and let's merge it.

chienandalu (approved):

azucenatrey (approved):
LGTM

PR #513, opened 2021-03-11
[13.0][MIG] website_sale_product_sort: Migration script for change res.config.setting stored value

sergio-teruel:
cc @Tecnativa TT26956

### DIFF
--- a/website_sale_product_sort/migrations/13.0.1.0.0/post-migration.py
+++ b/website_sale_product_sort/migrations/13.0.1.0.0/post-migration.py
@@ -1,0 +1,18 @@
+# Copyright 2021 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """We avoid the error of selection option not valid changing to the current
+    default one. On OpenUpgrade itself, the field website_description is
+    swapped to change the content of the field to match the new criteria.
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """UPDATE website
+        SET default_product_sort_criteria = 'website_sequence asc'
+        WHERE default_product_sort_criteria = 'website_sequence desc'""",
+    )


pedrobaeza:
Shouldn't this be in OpenUpgrade itself?

chienandalu:
I think so as well, as this is a change introduced from Odoo themselves.

sergio-teruel:
But the field default_product_sort_criteria is created by this field...

chienandalu:
Yes, sure. The problem here is going to be more for the instances without this module. All the products are going to have the reversed intended order when an Odoo instance is migrated to v13. Then an OpenUpgrade reversing the `website_sequence`  assigned to the products should be made, but the that would affect the intended order of the instances using this module...

So if an OU migration script is eventually made then maybe another migration script should be done in this module applying the inverse website sorting rule...

pedrobaeza:
For the OU script, a similar SQL to this can be used:

https://github.com/OCA/OpenUpgrade/blob/0ccd7c0eff0b9bd5b3ca7f624380a07daf2d2323/addons/website_slides/migrations/13.0.2.0/post-migration.py#L45-L56

sergio-teruel:
what are you talking about?

chienandalu (approved):
As we talked, this patch is ok to avoid errors when entering res.config.settings

pedrobaeza:
Is there the possibility of having more than one criteria for this and website_sequence be one of them? If so, you need to do `replace('website asc', 'website desc')` instead.

Anyway, please check pre-commit.

chienandalu:
> Is there the possibility of having more than one criteria for this and website_sequence be one of them? If so, you need to do replace('website asc', 'website desc') instead.

This issue has nothing to do with that, but with that default option not existing in the module anymore, so when you go to `res.config.settings` Odoo crashes

pedrobaeza:
OK, but this change the meaning of the option, so isn't better to empty the value or look for the same one? Travis is also red.

chienandalu:
> OK, but this change the meaning of the option, so isn't better to empty the value or look for the same one? Travis is also red.

It just defaults to Odoo's only option, which has changed in this version. The change is nothing more than that and it doesn't change anything that already existed

pedrobaeza (approved):

PR #518, opened 2021-03-26
[13.0][FIX] website_sale_product_minimal_price: Fix tests

joao-p-marques:
Ensure we use a consistent currency

@Tecnativa
TT28906

ping @CarlosRoca13 @pedrobaeza

### DIFF
--- a/website_sale_product_minimal_price/tests/test_website_sale_product_minimal_price.py
+++ b/website_sale_product_minimal_price/tests/test_website_sale_product_minimal_price.py
@@ -6,13 +6,23 @@ from odoo.tests.common import HttpCase
 class WebsiteSaleProductMinimalPriceHttpCase(HttpCase):
     def setUp(self):
         super().setUp()
-
+        # Create and select a pricelist
+        # to make tests pass no matter what l10n package is enabled
+        self.website = self.env["website"].get_current_website()
+        pricelist = self.env["product.pricelist"].create(
+            {
+                "name": "website_sale_product_minimal_price public",
+                "currency_id": self.env.company.currency_id.id,
+                "selectable": True,
+                "sequence": 1,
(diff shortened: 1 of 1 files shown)


website_sale_product_minimal_price/tests/test_website_sale_product_minimal_price.py
```diff
@@ -6,13 +6,14 @@
 class WebsiteSaleProductMinimalPriceHttpCase(HttpCase):
     def setUp(self):
         super().setUp()
-
+        # Always work with dollars
+        self.company = self.env.ref("base.main_company")
+        self.company.currency_id = self.env.ref("base.USD")
```
chienandalu: Isn't this going to clash with the currency change constrains?

pedrobaeza: In fact this is forbidden in code under certain conditions, so we will have another problem in the future.

joao-p-marques: Ok, but then what is the best solution? AFAICS the problem is that the test creates a product and assigns a price (not specifying currency) but, when running the tour, the price displayed in the shop follows the default currency set. 
With only this module installed, everything is computed in dollars (default) but, when another module is installed that changes the default currency for the company, it ends up setting a price in euros, but displaying it in dollars (with different values) thus making the tour fail...

pedrobaeza: Create a pricelist with that currency and assign it.

joao-p-marques:
Changes made @chienandalu @pedrobaeza 
I think it is correct now, thanks!

pedrobaeza (approved):

PR #520, opened 2021-03-31
Website form first name and last name

Algoritmun:
Module to add firstname and lastname to webiste address management form and portal my account form.


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

PR #526, opened 2021-04-12
[13.0] [MIG] website_snippet_product_category: Migration to v13

CarlosRoca13:
cc @Tecnativa TT26961

please @Tardo @joao-p-marques review this.


CarlosRoca13:
I set it in WIP because I need some help with next situation: 

- When we are at editable mode, the columns can be edited even though the block has the attribute `contenteditable="False"`

I tried to put this attribute to all columns but I can't achieve the expected behavior

website_snippet_product_category/static/src/js/snippet.options.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Tecnativa - Alexandre Díaz
+// License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+odoo.define("website_snippet_carousel_product.snippet_options", function(require) {
```
joao-p-marques: If we define here this snippet options related to another module, shouldn't we add it as a dependency?

website_snippet_product_category/views/product_public_category.xml
```diff
@@ -0,0 +1,27 @@
...
+            <field name="arch" type="xml">
+                <field name="image_1920" position="before">
+                    <div class="oe_button_box" name="button_box">
+                        <button
+                            class="oe_stat_button"
+                            name="website_publish_button"
+                            type="object"
+                            icon="fa-globe"
+                        >
+                            <field name="is_published" widget="website_button" />
```
joao-p-marques: I can't see this button correctly in the runbot:
image

joao-p-marques (changes requested):
Code and functional review.
Just 2 minor issues, otherwise looks good :+1:

website_snippet_product_category/static/src/js/frontend.js
```diff
@@ -3,37 +3,38 @@
...
-    var sAnimation = require("website.content.snippets.animation");
+    const core = require("web.core");
+    const sAnimation = require("website.content.snippets.animation");
 
-    var _t = core._t;
+    const _t = core._t;
 
     sAnimation.registry.js_product_category = sAnimation.Class.extend({
         selector: ".js_product_category",
+        disabledInEditableMode: false,
```
chienandalu, line 13: Did you check how the widget behaves in v12? It really doesn't let you edit anything.

CarlosRoca13: Yes, I test it, but if I don't use this key we can't see the preview

CarlosRoca13:
@joao-p-marques changes done

website_snippet_product_category/static/src/js/frontend.js
chienandalu: Ok, I see :+1:

chienandalu (approved):

joao-p-marques (approved):

pedrobaeza:
@CarlosRoca13 is this still WIP?

CarlosRoca13:
No, it isn't @pedrobaeza

PR #528, opened 2021-04-15
[ADD] website_sale_resource_booking: eCommerce for resource bookings

yajo:
This module extends the functionality of ``sale_resource_booking`` to support
the eCommerce use case and to allow your visitors to buy products that produce
a resource booking, and pre-book them before buying.

You can also set a timeout for those pre-bookings to expire if unpaid.

Draft until merged:

- [x] https://github.com/OCA/calendar/pull/25
- [x] https://github.com/OCA/sale-workflow/pull/1523

@Tecnativa TT28202


pedrobaeza:
Can this be considered ready? Can you check Travis?

yajo:
Code is ready but https://github.com/OCA/sale-workflow/pull/1523 is still not merged.

website_sale_resource_booking/README.rst
```diff
@@ -0,0 +1,117 @@
...
+related modules.
+
+All products that you link to a resource booking type will allow pre-bookings
+if sold from your eCommerce. To configure those pre-bookings timeout:
+
+#. Go to the product form in the backend.
+#. Use the *Resource booking timeout* field, in the *Sales* tab.
+
+When you go to that product's eCommerce page, you'll see a little message above
+the *Add to cart* button, telling the user that he will be able to pre-book it
```
pedrobaeza: Use an impersonal form:

```suggestion
the *Add to cart* button, telling the user that they will be able to pre-book it
```

https://en.m.wikipedia.org/wiki/Singular_they

website_sale_resource_booking/controllers/main.py
```diff
@@ -0,0 +1,115 @@
...
+        if not bookings:
+            return request.redirect("/shop/checkout")
+        # Proceed to checkout if we passed the last booking
+        try:
+            booking = self._get_indexed_booking(index).with_context(
+                checkout_booking_index=index
+            )
+        except IndexError:
+            return request.redirect("/shop/checkout")
+        amount = len(bookings)
```
pedrobaeza: Amount is used for money and related things. You can use `count` instead.

website_sale_resource_booking/models/resource_booking.py
```diff
@@ -0,0 +1,97 @@
...
+                ("state", "in", ("pending", "scheduled")),
+            ]
+            + domain
+        )
+        expired.action_cancel()
+
+    def action_cancel(self):
+        "Clean personal/cron data that you will never need again."
+        self.write(
+            {"prereserved_name": False, "prereserved_email": False, "expiration": False}
```
pedrobaeza, line 98: I think names for traceability shouldn't be emptied.

website_sale_resource_booking/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Jairo Llopis <jairo.llopis@tecnativa.com> (https://www.tecnativa.com/)
```
pedrobaeza: Please use the double line format with Tecnativa and url on first line and your name without email address on the second.

website_sale_resource_booking/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2021 Tecnativa - Jairo Llopis
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Sell resource booking products in your eCommerce",
+    "summary": "Let customers book resources temporarily before buying",
+    "version": "12.0.1.0.0",
+    "development_status": "Beta",
```
pedrobaeza: As the module has tests, this can be Production/Scheduled

website_sale_resource_booking/models/resource_booking.py
yajo: Keeping that personal information after the prereservation expires is not legally preauthorized from the perspective of GDPR, so if we don't remove that information, we need to inform our users about that we are we keeping this information, why we do it, and for how long (and implement a cron that deletes it when that time expires).

For the sake of legal and implementation simplicity, the information is emptied when no longer necessary. You still retain the customer information in the sales order in case you need to contact them and ask for new reservations.

pedrobaeza: I don't think so, as they have asked for it, so this information is legitimate, but if you keep this cleaning, at least explain why in the docstring / comment, as I always ask.

yajo: OK, I modified the docstring to make it more clear.

pedrobaeza (approved):

chienandalu (approved):

pedrobaeza:
@Yajo please check Travis

yajo:
/ocabot merge nobump

It seems those were just linter problems. Fixed now.

PR #531, opened 2021-04-23
[12.0][ADD] website_sale_force_qty: Force product qty on website

joao-p-marques:
New module to force a specific quantity to be sold on website in a product on each order.

**TODO:**
- [x]  Finish tests

**Screenshots:**

@Tecnativa
TT29306

ping @Yajo


website_sale_force_qty/models/product_template.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    website_sale_force_qty = fields.Float(
+        string="Force quantity in website", default=0.0,
+        help="Limit the product to a specific quantity to be bought in the website."
+        " Set to 0 to disable"
+    )
```
chienandalu, line 13: It should be constrained to avoid negatives, don't you think?

website_sale_force_qty/models/sale_order.py
```diff
@@ -0,0 +1,17 @@
...
+class SaleOrder(models.Model):
+    _inherit = 'sale.order'
+
+    def _cart_update(self, product_id=None, line_id=None, add_qty=0, set_qty=0, **kwargs):
+        """Force quantity on product from what is defined in product.template"""
+        # Need to call super first to ensure correct default mechanism
+        res = super()._cart_update(product_id, line_id, add_qty, set_qty, **kwargs)
+        # Do nothing if we are removing the product (qty=0)
+        if res["quantity"] > 0 and product_id:
+            product = self.env["product.product"].browse(product_id)
```
chienandalu, line 13: Should we add to the roadmap that qty forcing at variant level could be considered in the future?

website_sale_force_qty/static/src/js/website_sale_force_qty_tour.js
```diff
@@ -0,0 +1,46 @@
...
+        {
+            trigger: "a:contains('Test product with fixed quantity')",
+        },
+        {
+            extra_trigger: [
+                "div.js_main_product",
+                // The quantity input box is disable and is set to 2
+                ":has(input[name=add_qty][value='2']:disabled)",
+                // There is a badge indicating the product as a fixed quantity
+                ":has(.text-info:contains('This product has a fixed quantity'))",
```
chienandalu, line 20: Maybe this text should be customizable to allow further explanations

chienandalu (commented):
Cool :+1: Some thoughts :slightly_smiling_face: :

website_sale_force_qty/models/sale_order.py
joao-p-marques: Yes, it could be a good idea. It would even help with the other point in the roadmap, as the variants have access to the quantity already in the cart, so we could use that to tweak the "Add to cart" button
@Yajo WDYT?

website_sale_force_qty/static/src/js/website_sale_force_qty_tour.js
joao-p-marques: Mmm yes, it could be, adding it to the roadmap :+1:

PR #539, opened 2021-05-17
[12.0][FIX] website_sale_product_minimal_price: Take into account all pricelist item

CarlosRoca13:
Change the get_combination_info_pricelist_atributes method to get the
correct scale of prices.

cc @Tecnativa TT29708

please @pedrobaeza @chienandalu review it

### DIFF
--- a/website_sale_product_minimal_price/controllers/main.py
+++ b/website_sale_product_minimal_price/controllers/main.py
@@ -1,7 +1,7 @@
 # Copyright 2020 Tecnativa - Carlos Roca
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo.addons.website_sale.controllers.main import WebsiteSale
-from odoo import http, fields
+from odoo import http
 from odoo.http import request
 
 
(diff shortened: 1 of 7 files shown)


pedrobaeza:
This is not `[IMP]`, but `[FIX]`, as it wasn't working previously. The commit message can be: 

```
[FIX] website_sale_product_minimal_price: Take into account all pricelist items

Previous approach only considered pricelist items in the same pricelist as the current 
one, but you can have some items that point to another pricelists, so we are changing
the approach to get all possible items, no matter the pricelist, get the minimal quantity
which they apply to, and then get the price for that quantity to see if with the price
computation, they return a different value, marking them in the scale table. With this,
we get a compromise between performance and accuracy.

NOTE: There are still left intentionally items for "All products" that can have other
scaled prices, but as this can have a lot of impact in performance and it's improbable
that someone configures pricelists this way, it's better to not include them.
```

pedrobaeza (approved):
Code and functional review, taking into account my commit message comment

CarlosRoca13:
@pedrobaeza changed

pedrobaeza:
Don't put `@Tecnativa` on the commit message. You can put the TT, but not the mention.

CarlosRoca13:
@pedrobaeza changed

website_sale_product_minimal_price/controllers/main.py
```diff
@@ -16,48 +16,33 @@ def get_combination_info_pricelist_atributes(
...
+        # price scale.
+        qty_list = request.env["product.pricelist.item"].search([
+            "|",
+            ("product_id", "=", product.id),
+            "|",
+            ("product_tmpl_id", "=", product.product_tmpl_id.id),
+            ("categ_id", "in", list(map(int, product.categ_id.parent_path.split('/')[0:-1]))),
+            ("min_quantity", ">", 0),
+        ]).mapped("min_quantity")
+        qty_list = sorted(set(qty_list))
```
chienandalu, line 32: How do you know that the pricelist is chained to the used one?

CarlosRoca13: In this case we take all the pricelist items asociated to this product to know what are the `minimal_qty`s that have to be checked, so we will know the scaled prices. This method is used just on the grid product price scale

pedrobaeza (approved):

pedrobaeza:
@chienandalu can you confirm?

chienandalu (approved):

pedrobaeza:
@CarlosRoca13 please take note for fw-porting the patch to v13 when possible.

PR #542, opened 2021-05-18
[13.0][IMP] website_sale_checkout_skip_payment: Changes on Confirm Order buttons

CarlosRoca13:
With this changes the 'Process Checkout' button change to 'Confirm' when the partner has selected the option skip_website_checkout_payment. Furthermore, the button text 'Confirm Order' has been changed to 'Confirm'.

cc @Tecnativa TT29736

please @pedrobaeza @victoralmau review this

### DIFF
--- a/website_sale_checkout_skip_payment/i18n/am.po
+++ b/website_sale_checkout_skip_payment/i18n/am.po
@@ -8,15 +8,17 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 9.0c\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2017-08-15 03:01+0000\n"
-"PO-Revision-Date: 2017-08-15 03:01+0000\n"
-"Last-Translator: OCA Transbot <transbot@odoo-community.org>, 2017\n"
-"Language-Team: Amharic (https://www.transifex.com/oca/teams/23907/am/)\n"
+"POT-Creation-Date: 2021-05-18 10:04+0000\n"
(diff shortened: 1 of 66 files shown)


pedrobaeza:
There should be something different, as the customer has the text "Pay Now". Can you check if there's any difference or other module adding that?

CarlosRoca13:
"Pay Now" appears when the partner asociated to the user has the field skip_website_checkout_payment set as True.

pedrobaeza:
It should be the contrary then

CarlosRoca13:
Why?

pedrobaeza:
If the payment is skipped (so `skip_website_checkout_payment` = True), then "Pay Now" text has no sense. For this case, "Confirm" is the needed one if I don't understand bad what you are telling to me.

CarlosRoca13:
Ups I thought that I said all the opposite, it's working as you said. When the partner has `skip_website_checkout_payment = False` the button pay now will be shown.

I will work for correct the tests

CarlosRoca13:
I add the fix on other commit.

chienandalu (approved):
Tested on runbot :+1:

pedrobaeza:
Wow, the number of changes in the PO files is huge. Maybe it's better to leave "Confirm Order" although less generic?

CarlosRoca13:
Okey!

CarlosRoca13:
BTW the changes will be there if we update the po files with the pot, so IDK if it will reduce the changes in po files

pedrobaeza:
OK, let's go with this then:

/ocabot merge minor

PR #548, opened 2021-06-07
[FIX] webiste_sale_order_type: trigger onchanges

chienandalu:
If we don't trigger the onchanges, we won't get some of the default
values that the order type has defined

cc @Tecnativa TT30235

please, review @joao-p-marques @pedrobaeza

### DIFF
--- a/website_sale_order_type/models/sale_order.py
+++ b/website_sale_order_type/models/sale_order.py
@@ -19,4 +19,5 @@ class SaleOrder(models.Model):
                      self.partner_id.commercial_partner_id.sale_type)
         if sale_type:
             self.type_id = sale_type
+            self.onchange_type_id()
         return res


pedrobaeza (approved):
Maybe this impacts too much in the performance. Should we trigger a manual call to `onchange` instead?

chienandalu:
Ok, let's just trigger the method itself then

pedrobaeza (approved):
Better that way

joao-p-marques (approved):
Seems like this got lost in the migration, sorry.

LGTM :+1:

HaraldPanten (approved):

pedrobaeza:
@chienandalu Travis is red

chienandalu:
Looks like a random issue. I'm rebuilding the only build that failed to see if it passes now

chienandalu:
Green now :slightly_smiling_face:

chienandalu:
There's also a warning in runbot, but is in the main branch. Due to website_sale_charge_payment_fee

PR #556, opened 2021-07-14
[12.0][FIX] website_sale_secondary_unit: Remove sudo access to avoid to display unpublished secondary units.

sergio-teruel:
cc @Tecnativa TT31018

ping @chienandalu @CarlosRoca13

### DIFF
--- a/website_sale_secondary_unit/__manifest__.py
+++ b/website_sale_secondary_unit/__manifest__.py
@@ -16,6 +16,7 @@
         'sale_order_secondary_unit',
     ],
     'data': [
+        'security/ir.model.access.csv',
         'security/website_sale_secondary_unit.xml',
         'views/assets.xml',
         'views/product_template_views.xml',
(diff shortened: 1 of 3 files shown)


victoralmau (approved):

pedrobaeza (changes requested):
I don't think we should change this, as we are leaking information to public. Instead, you should filter specifically on the published attribute on the controller/Qweb.

chienandalu:
I don't see any leak. The record rule is there to avoid showing more than it's published: https://github.com/OCA/e-commerce/blob/b81325fb59c09f21e3deace004dea4c9c81000b2/website_sale_secondary_unit/security/website_sale_secondary_unit.xml#L7-L8

chienandalu (approved):

pedrobaeza (approved):
OK then

sergio-teruel:
Thanks!!

chienandalu:
This should be forward ported, isn't it?

PR #564, opened 2021-08-30
[14.0][MIG] website_sale_checkout_skip_payment: Migration to V14

ferran-S73:
Standard migration from 13.0


website_sale_checkout_skip_payment/tests/test_checkout_skip_payment.py
```diff
@@ -11,9 +11,6 @@ def setUp(self):
         self.partner.with_context({"res_partner_search_mode": "customer"}).write(
             {"skip_website_checkout_payment": True}
         )
-        # Delete optional products for avoid popup window
-        product = self.env.ref("product.product_product_4_product_template")
```
leemannd: Hello, what is the reason to remove these test lines?

website_sale_checkout_skip_payment/controllers/main.py
```diff
@@ -0,0 +1,42 @@
...
+            "recall": True,
+            "message": request.website._render(
+                "website_sale_checkout_skip_payment.order_state_message"
+            ),
+        }
+
+    @http.route()
+    def payment_confirmation(self, **post):
+        if not request.website.checkout_skip_payment:
+            return super().payment_confirmation(**post)
```
leemannd, line 28: Non blocking comment: Could be great to have a test about this redirection

leemannd (approved):
Migration seems good

ajaniszewska-dev (approved):

website_sale_checkout_skip_payment/tests/test_checkout_skip_payment.py
ferran-S73: I removed them because `optional_product_ids` is no longer a field `product.template`. That's why I think those lines are not needed anymore.

leemannd: It seems to be a wrong assumption. If you activate it, it is present.

PR #567, opened 2021-09-14
[12.0][FIX] website_sale_product_minimal_price: default variant

chienandalu:
When there's no minimal price (all the variant prices are the same) keep
the default first combination behavior, that is: take it counting on the
attribute values sequences.

cc @Tecnativa TT31880

please review @sergio-teruel @pedrobaeza @CarlosRoca13

### DIFF
--- a/website_sale_product_minimal_price/README.rst
+++ b/website_sale_product_minimal_price/README.rst
@@ -52,6 +52,20 @@ Usage
 #. Click on product, the price displayed is the minimal variant price. On the chart
    under the buy button you can see the different taxes applied to the product.
 
+Known issues / Roadmap
+======================
+
+* When there are some variants with the same minimal price, we won't get the expected
+  attribute value sequence order to get the first combination but the default
+  `product.product` order. This is like this because we stretch the possible variants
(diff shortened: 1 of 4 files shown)


pedrobaeza (commented):
This is a partial solution, as maybe there are 10 variants, 5 of them with one price, and the other 5 with another. We should try to get the order inside the same price level, but we can put that in the ROADMAP for now.

chienandalu:
Yes, that will be more complicated, as we stretch the cheapest product finding to a single result. We should store pairs of products and prices and compare them with final minimal price, getting the product ids and restricting the possible combinations to those variants and ideally somehow trying to rely on the super() to make the final pick. I'll put in the ROADMAP for now.

pedrobaeza (approved):

sergio-teruel (approved):

CarlosRoca13 (approved):

pedrobaeza:
Merging manually as ocabot doesn't work for now.

pedrobaeza:
Please replicate it in v13.

PR #571, opened 2021-09-21
[13.0][IMP]website_sale_hide_empty_category improvements

Rad0van:
Don't know if you'll like this but I needed such functionality and I am sure it can be improved further. Generally the changes provide:

- Change from bool to int in counting number of products in categories
- Added option to display the number of products in each category based on the above
- Only published products are being counted (doesn't make sense to include non-published products)

Any comments/improvements are welcome.

### DIFF
--- a/website_sale_hide_empty_category/models/product_public_category.py
+++ b/website_sale_hide_empty_category/models/product_public_category.py
@@ -12,10 +12,16 @@ class ProductPublicCategory(models.Model):
         compute="_compute_has_product_recursive",
     )
 
+    count_product_recursive = fields.Integer(
+        string="Number of products in this category or one of its children",
+        compute="_compute_has_product_recursive",
+    )
+
     @api.depends("product_tmpl_ids", "child_id.has_product_recursive")
     def _compute_has_product_recursive(self):
         for category in self:
-            category.has_product_recursive = bool(
-                category.product_tmpl_ids
-                or any(child.has_product_recursive for child in category.child_id)
+            product_count = len(category.product_tmpl_ids) + sum(
+                child.has_product_recursive for child in category.child_id
             )
+            category.count_product_recursive = product_count
+            category.has_product_recursive = bool(product_count)
--- a/website_sale_hide_empty_category/README.rst
+++ b/website_sale_hide_empty_category/README.rst
@@ -26,7 +26,7 @@ Website Sale - Hide Empty Categories
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
 Removes any category links from the Shop which have no products associated
-to them.
+to them. Also allows displaying number of products in individual categories.
 
 **Table of contents**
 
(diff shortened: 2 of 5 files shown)


website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -7,15 +7,14 @@
...
-    def _compute_has_product_recursive(self):
+    @api.depends("product_tmpl_ids", "child_id.count_product_recursive")
+    def _compute_count_product_recursive(self):
         for category in self:
-            category.has_product_recursive = bool(
-                category.product_tmpl_ids
-                or any(child.has_product_recursive for child in category.child_id)
-            )
+            category.count_product_recursive = len(
+                category.product_tmpl_ids.filtered("is_published")
```
chienandalu: Why do you need to change all this stuff when this is the only change needed to make this improvement work?

Rad0van: Not sure I understand what you mean. Which stuff you have in mind? Names of the property and compute function? Well to match what they do. Or you have something else in mind? That calculation of number of products in that category?

Rad0van: @chienandalu I think I understand now after playing with _website_sale_product_brand_ and trying to make it un/publishable. That `.filtered("is_published")` part is not really necessary as it is added through _ir.rule_ as need be (public user only seeing published products).

The question remains whether you object renaming that property and method. Should I leave the original names that suggest they are of boolean type? Please comment and I would change as necessary.

chienandalu: I think it would be less disruptive to just add the product counter as a complementary field. Keep in mind that custom extensions could be broken removing a field just in the middle of a version. The same goes with the views. Custom themes could rely on the current xml architecture, and no ecommerce holder will be happy when a 500 raises after updating the module.

You can add some migration notes anyway to remove the boolean in the next version.

Rad0van: But in such case wouldn't a separate module be more appropriate option? But then I am not sure about performance? Would separate (from boolean evaluation) counting affect performance or would ORM handle this without any penalty?

Or how about  just changing the return type and leave the function names intact so that it wouldn't break anything?

Or maybe (that looks the best for me) something like this - a mix of both:
- change the return type of the compute function to integer
- add complementary product counter that just takes the result of the compute function

What would you suggest? Thanx a lot.

chienandalu: I was going to propose that, but I think the two fields can coexist in this module. You can compute the two fields altogether and let the boolean be set by the counter (`bool(count_result)`). The compute doesn't return anything anyway, so you can add that behavior.

So the changes would be to keep everything as it is now adding the counter field to the compute.

Then you can just add the custom view extension, as it's not going to harm anything pre-existing

Maybe for v14 (you can put it on the roadmap) the module can be split on two:

- The one that renders the count
- The one that hides the category depending on that count.

Rad0van: Thanx a lot for input. I am not sure about v14 - I'd rather suggest renaming the module. Would seem to me that splitting it would be creating very small modules... But that can be decided later on I guess. Going to adjust the module now.

PR #572, opened 2021-09-24
[13.0][IMP] website_sale_product_attachment: Introduce website name

pedrobaeza:
On previous version, attachments had 2 fields for adding both attachment name and file name.

Now on v13, there's only one, that is filled with the file name. On initial migration, it was considered that this field is enough, but putting file names on the website product page can be ugly, limited and confusing, so we are adding here a new field to store the name we want to give it for the website e-commerce product page.

It also includes migration script for recovering the old information if coming from v12.

@Tecnativa TT32160

### DIFF
--- a/website_sale_product_attachment/migrations/13.0.2.0.0/post-migration.py
+++ b/website_sale_product_attachment/migrations/13.0.2.0.0/post-migration.py
@@ -1,0 +1,16 @@
+# Copyright 2021 Tecnativa - Pedro M. Baeza
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+from psycopg2 import sql
+
+from odoo import tools
+
+
+def migrate(cr, version):
+    """Recover name column from v12 if exists to be used as website name."""
+    column = "openupgrade_legacy_13_0_name"
+    if tools.column_exists(cr, "ir_attachment", column):
+        cr.execute(
+            sql.SQL("UPDATE ir_attachment SET website_name = {}").format(
+                sql.Identifier(column)
+            )
+        )
--- a/website_sale_product_attachment/migrations/13.0.2.0.0/pre-migration.py
+++ b/website_sale_product_attachment/migrations/13.0.2.0.0/pre-migration.py
@@ -1,0 +1,11 @@
+# Copyright 2021 Tecnativa - Pedro M. Baeza
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    """Remove offending view for allowing to update."""
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    view = env.ref("website_sale_product_attachment.download_icons", False)
+    if view:
+        view.unlink()
--- a/website_sale_product_attachment/models/ir_attachment.py
+++ b/website_sale_product_attachment/models/ir_attachment.py
@@ -1,4 +1,5 @@
 # Copyright 2020 Tecnativa - Jairo Llopis
+# Copyright 2021 Tecnativa - Pedro M. Baeza
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
 from odoo import fields, models
@@ -12,4 +13,9 @@ class IrAttachment(models.Model):
         comodel_name="product.template",
         help="Attachment publicly downladable from eCommerce pages "
         "in these product templates.",
+    )
+    website_name = fields.Char(
+        string="Name in e-commerce",
+        help="The name of the download that will be displayed on the e-commerce "
+        "product page. If not filled, the filename will be shown by default.",
     )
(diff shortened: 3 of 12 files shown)


chienandalu (approved):

victoralmau (approved):

website_sale_product_attachment/migrations/13.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+from odoo import tools
+
+
+def migrate(cr, version):
+    """Recover name column from v12 if exists to be used as website name."""
+    column = "openupgrade_legacy_13_0_name"
+    if tools.column_exists(cr, "ir_attachment", column):
+        cr.execute(
+            sql.SQL("UPDATE ir_attachment SET website_name = {}").format(
```
yajo, line 13: I think you're missing a `WHERE` here, to affect only website product attachments, isn't it?

website_sale_product_attachment/models/ir_attachment.py
```diff
@@ -13,3 +14,7 @@ class IrAttachment(models.Model):
         help="Attachment publicly downladable from eCommerce pages "
         "in these product templates.",
     )
+    website_name = fields.Char(
```
yajo, line 17: I think the field name is confusing, because it seems to indicate the name of the website, and not the name of the attachment when displayed in the website. Even more when the field is displayed just next to the website dropdown in the backend:

imagen


Also, the attachment name probably needs to be translated. If you're browsing the site in English you might want to download the "Instructions manual" while when browsing in Spanish, you want the "Manual de instrucciones".

So this would be better IMHO:

```suggestion
    website_name = fields.Char(
        string="Name in website",
        translate=True,
```

website_sale_product_attachment/views/product_template.xml
```diff
@@ -16,7 +17,8 @@
                         or unpublish it from your database.
                     </div>
                     <field name="website_attachment_ids" nolabel="1">
-                        <tree>
+                        <tree default_order="website_name,name,id">
+                            <field name="website_name" />
                             <field name="name" />
```
yajo: Also to avoid confusion in backend UI:

```suggestion
                            <field name="name" string="File name" />
```

website_sale_product_attachment/views/product_template.xml
```diff
@@ -16,7 +17,8 @@
                         or unpublish it from your database.
                     </div>
                     <field name="website_attachment_ids" nolabel="1">
-                        <tree>
+                        <tree default_order="website_name,name,id">
```
yajo, line 20: This will produce different orders in backend and frontend. Is that what you want?

imagen

To avoid that, `website_name` should be a computed writable field, and order on that only. Or a `sequence` field could be used instead.

yajo (changes requested):

website_sale_product_attachment/migrations/13.0.2.0.0/post-migration.py
pedrobaeza: Not really, as all of them can have a name on the previous version (or at least, it will be the same as the filename), so we don't want to restrict having a value for this under a condition (like `public=True`) that can vary across time.

website_sale_product_attachment/models/ir_attachment.py
pedrobaeza: String changed in the next push to `Name in e-commerce` for being more specific, but the translatable property is another thing, as this will mean that sorting will depend on this translation. As our use case doesn't involve translations for now, I have added it to the ROADMAP for being handled later if ever needed.

website_sale_product_attachment/views/product_template.xml
pedrobaeza: Although you are right, I prefer to not over-complicate the design for that only case where you don't fill one of the fields, as we can only sort by stored fields, and having 3 fields for such simple feature is too much (and think also if adding translations. This simply won't work).

pedrobaeza:
@Yajo changes pushed

website_sale_product_attachment/migrations/13.0.2.0.0/pre-migration.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2021 Tecnativa - Pedro M. Baeza
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    """Remove offending view for allowing to update."""
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    view = env.ref("website_sale_product_attachment.download_icons", False)
+    view.unlink()
```
yajo: This will fail if not found:

```suggestion
    if view:
        view.unlink()
```

website_sale_product_attachment/readme/ROADMAP.rst
```diff
@@ -0,0 +1,2 @@
+- Make translatable the name of the attachment in the e-commercer. This will
+  mean to deal with the sort of the translated terms.
```
yajo:
```suggestion
- Make translatable the name of the attachment in the e-commerce. This will
  mean to deal with the sorting of the translated terms.
```

yajo (changes requested):

pedrobaeza:
@Yajo changes done

yajo (approved):

PR #594, opened 2021-12-01
[FIX] website_sale_stock_provisioning_date: timezone

chienandalu:
We want to show the provisioning date in the context of the company
timezone. Otherwise we could get missleading dates.

cc @Tecnativa TT33244

ping @ernestotejeda @pedrobaeza

### DIFF
--- a/website_sale_stock_provisioning_date/hooks.py
+++ b/website_sale_stock_provisioning_date/hooks.py
@@ -1,0 +1,22 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+def pre_init_hook(cr):
+    """Precompute all the dates at once"""
+    cr.execute("ALTER TABLE stock_move ADD COLUMN IF NOT EXISTS date_provisioning DATE")
+    cr.execute(
+        """
+        UPDATE stock_move sm0
+        SET date_provisioning = (
+            sm.date_expected AT TIME ZONE COALESCE(rp.tz, 'UTC')
+        )::date
+        FROM stock_move sm
+        LEFT JOIN res_users ru ON sm.write_uid = ru.id
+        LEFT JOIN res_partner rp ON ru.partner_id = rp.id
+        WHERE
+            sm0.id = sm.id
+            AND sm.date_expected IS NOT NULL
+            AND sm.state NOT IN ('cancel', 'done')
+    """
+    )
--- a/website_sale_stock_provisioning_date/models/stock_move.py
+++ b/website_sale_stock_provisioning_date/models/stock_move.py
@@ -1,0 +1,23 @@
+# Copyright 2021 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from pytz import timezone
+
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    date_provisioning = fields.Date(
+        string="Date expected", compute="_compute_date_provisioning", store=True,
+    )
+
+    @api.depends("date_expected")
+    def _compute_date_provisioning(self):
+        """We want to store the date in the context of the user that sets it avoiding
+        offsets due to timezones. This way, when it's shown in the website, we can
+        ensure that it's our date and not the user's"""
+        self.date_provisioning = False
+        tz = timezone(self.env.user.tz or "UTC")
+        for move in self.filtered("date_expected"):
+            move.date_provisioning = move.date_expected.astimezone(tz).date()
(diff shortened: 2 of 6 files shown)


pedrobaeza:
Not sure why using such tz in a calendar that may not be assigned. I think you have to use the user timezone that is visiting your website.

chienandalu:
Well, the problem here is the following. We get a move that has a planned date of 2022-01-01 00:00:00 (UTC+1). So the date in which the company will receive it is 2022-01-01, but as it's stored in UTC, different users could get different dates. For example, a user from Portugal (UTC+0) would see that the planned date is 2021-12-31, and that's what we want to avoid.

pedrobaeza:
I think we are trying to solve a problem that comes from a bad user input. If a user introduces in the expected date `01/01/2022 00:00:00`, and their timezone is GMT+2, in fact the day that will come will be 31/12/2021, not the other one. It should introduce `01/01/2022 12:00:00` as the way to set exactly a realistic date (you are not going to pickup the stock on midnight). Anyways, I insist that if you want to "decipher" the date that the user is seeing when they introduce the expected date, then you must use user timezone, not a probably undefined company calendar timezone.

chienandalu:
It will be a problem with good user inputs and large timezone differences. We want to show the date the company receives it in their timezone. We can discuss whether or not the company resource calendar is reliable or not (there aren't many alternatives, or at least not that simple)

chienandalu:
@pedrobaeza Changes done according to what we commented internally

website_sale_stock_provisioning_date/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+    cr.execute(
+        """
+       UPDATE stock_move sm0
+       SET date_expected_date = (
+           sm.date_expected AT TIME ZONE COALESCE(rp.tz, 'UTC')
+       )::date
+       FROM stock_move sm
+       LEFT JOIN res_users ru ON sm.write_uid = ru.id
+       LEFT JOIN res_partner rp ON ru.partner_id = rp.id
+       WHERE sm.date_expected IS NOT NULL AND sm0.id = sm.id
```
pedrobaeza: Shouldn't you limit this to the non done moves?

website_sale_stock_provisioning_date/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from pytz import timezone
+
+from odoo import api, fields, models
+
+
+class StockMove(models.Model):
+    _inherit = "stock.move"
+
+    date_expected_date = fields.Date(
```
pedrobaeza: Isn't better to call it `date_provisioning`?

website_sale_stock_provisioning_date/models/stock_move.py
```diff
@@ -0,0 +1,20 @@
...
+        string="Date expected", compute="_compute_date_expected_date", store=True,
+    )
+
+    @api.depends("date_expected")
+    def _compute_date_expected_date(self):
+        self.date_expected_date = False
+        tz = timezone(self.env.user.tz or "UTC")
```
pedrobaeza, line 21: You can put a note saying that it's expected that when the `date_expected` is changed, the user that is changing it (thus `self.env.user`), expects the day that they are seeing

website_sale_stock_provisioning_date/models/stock_move.py
chienandalu: It was my initial call, but it doesn't fit when the move isn't an incoming one, I think...

pedrobaeza: Well, this field name is very weird right now. It's true the other is not the finest one, but is not bad at all I think.

website_sale_stock_provisioning_date/hooks.py
chienandalu: Sure. Done

website_sale_stock_provisioning_date/models/stock_move.py
chienandalu: Done

pedrobaeza (approved):

website_sale_stock_provisioning_date/models/stock_move.py
chienandalu: ok then

website_sale_stock_provisioning_date/hooks.py
```diff
@@ -0,0 +1,22 @@
...
+        SET date_provisioning = (
+            sm.date_expected AT TIME ZONE COALESCE(rp.tz, 'UTC')
+        )::date
+        FROM stock_move sm
+        LEFT JOIN res_users ru ON sm.write_uid = ru.id
+        LEFT JOIN res_partner rp ON ru.partner_id = rp.id
+        WHERE
+            sm0.id = sm.id
+            AND sm.date_expected IS NOT NULL
+            AND sm.state IN ('waiting', 'confirmed', 'partially_available', 'assigned')
```
pedrobaeza: For the possible draft moves?

```suggestion
            AND sm.state != 'done
```

chienandalu: Changed discarding done and cancelled moves

ernestotejeda (approved):

PR #595, opened 2021-12-01
[13.0][ADD] website_sale_product_assortment: New module to use product assortments on e-commerce

CarlosRoca13:
cc @Tecnativa TT32825

@sergio-teruel @chienandalu


CarlosRoca13:
@chienandalu I added the change that we commented

CarlosRoca13:
I add some test and demo data

CarlosRoca13:
I had to delete the demo data for avoid errors on other tests. The PR is ready for review

chienandalu (commented):
I don't know if it's an issue of the main module, but here's the problem:

- Everything works ok with regular user's partners.
- But when it comes to Public User, the domain isn't properly applied due to its conditions of archived record.
- So it seems the context active_test=False is needed somewhere 

It doesn't make sense that a logged in user can't interact with some products while the public user does, don't you think?

website_sale_product_assortment/models/product_template.py
```diff
@@ -0,0 +1,137 @@
...
+        partner = self.env.user.partner_id
+        assortments = (
+            self.env["ir.filters"].sudo().search([("is_assortment", "=", True)])
+        )
+        not_allowed_product_dict = {}
+        for assortment in assortments:
+            if (
+                partner & assortment.sudo().all_partner_ids
+            ) and assortment.website_availability in ["no_purchase"]:
+                not_allowed_products = self.env["product.product"].search(
```
sergio-teruel: This search can cause performance issues

sergio-teruel (changes requested):

CarlosRoca13:
I added the condition on _get_combination_info for avoid running this search for each product, this will just be launched if we access to the product sheet.

And I added active_test=False when we get the partners involved

@sergio-teruel @chienandalu

website_sale_product_assortment/README.rst
```diff
@@ -0,0 +1,101 @@
...
+
+Configuration
+=============
+
+To see this module working, you have to define a product assortment and select
+an option on website availability field.
+
+#. **Don't apply restriction**: This option will not set any kind of restriction on 
+product items.
+#. **Avoid show products not available**: This option will hide on the e-commerce, the
```
pedrobaeza:
```suggestion
#. **Avoid to show non available products**: This option will hide on the e-commerce, the 
```

website_sale_product_assortment/README.rst
```diff
@@ -0,0 +1,101 @@
...
+To see this module working, you have to define a product assortment and select
+an option on website availability field.
+
+#. **Don't apply restriction**: This option will not set any kind of restriction on 
+product items.
+#. **Avoid show products not available**: This option will hide on the e-commerce, the 
+products that are not added to the products domain. If a product template has at least
+one allowed variant to show, the product will appear on the product items view but only
+that variants will be able to be bought. 
+#. **Avoid purchase products not available**: This option will restrict to buy the
```
pedrobaeza:
```suggestion
#. **Avoid selling not available products**: This option will restrict to buy the 
```

website_sale_product_assortment/README.rst
```diff
@@ -0,0 +1,101 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To see this module working, you have to define a product assortment and select
+an option on website availability field.
```
pedrobaeza:
```suggestion
an option on the website availability field.
```

website_sale_product_assortment/README.rst
```diff
@@ -0,0 +1,101 @@
...
+product items.
+#. **Avoid show products not available**: This option will hide on the e-commerce, the 
+products that are not added to the products domain. If a product template has at least
+one allowed variant to show, the product will appear on the product items view but only
+that variants will be able to be bought. 
+#. **Avoid purchase products not available**: This option will restrict to buy the 
+products that are added to the assortment on the e-commerce. To inform the clients,
+two more fields where added, "Message when unavailable" and "Assortment information".
+The first one will add a short description to the product items and sheets and the other
+one will set a detailed description on the products sheet.
```
pedrobaeza:
```suggestion
two more fields were added: "Message when unavailable" and "Assortment information".
The first one will add a short description to the product item and the other one will set a 
detailed description on the product sheet. This second one is editable from the website editor.
```

website_sale_product_assortment/README.rst
```diff
@@ -0,0 +1,101 @@
...
+~~~~~~~
+
+* Tecnativa
+
+Contributors
+~~~~~~~~~~~~
+
+* `Tecnativa <https://www.tecnativa.com>`_:
+
+    * Carlos Roca
```
pedrobaeza, line 76: Please put me as contributor, being the initial designer.

website_sale_product_assortment/controllers/website_sale.py
```diff
@@ -0,0 +1,68 @@
...
+    def _get_search_domain(
+        self, search, category, attrib_values, search_in_description=True
+    ):
+        """Overriding _get_search_domain method to avoid show product templates that
+        has all their variants not allowed to be shown.
+        """
+        res = super()._get_search_domain(
+            search, category, attrib_values, search_in_description=search_in_description
+        )
+        allowed_product_variants, assortment_restriction = self._get_products_allowed()
```
pedrobaeza: Can you avoid calling 2 times to the same method?

website_sale_product_assortment/models/product_template.py
pedrobaeza: This seems to be still pending

website_sale_product_assortment/models/sale_order.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2021 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    warning_message = fields.Char()
```
pedrobaeza: Why do you need this?

pedrobaeza (changes requested):
Good work overall. We need to fine-tune the performance and to avoid to do several times the same thing.

website_sale_product_assortment/controllers/website_sale.py
CarlosRoca13: I don't know were it is called 2 times... There are 2 different methods that call `_get_products_allowed()`

pedrobaeza: Yes, in this file you can see several calls to the same method, and you should do it only once for performance reasons.

CarlosRoca13:
Please @sergio-teruel @pedrobaeza review it again :smile:

francesco-ooops:
@CarlosRoca13 @carlosdauden is this finished? can you rebase the PR for testing? thanks!

CarlosRoca13:
@francesco-ooops done ;)

francesco-ooops (changes requested):
tested with this setup, it does not seem to be working: 

image

francesco-ooops:
I see that selection for "Availability in website" has 4 options, three of those are described in readme:

> #. **Don't apply restriction**: This option will not set any kind of restriction on
> product items.
> #. **Avoid to show non available products**: This option will hide on the e-commerce, the
> products that are not added to the products domain. If a product template has at least
> one allowed variant to show, the product will appear on the product items view but only
> that variants will be able to be bought.
> #. **Avoid selling not available products**: This option will restrict to buy the
> products that are added to the assortment on the e-commerce. To inform the clients,
> two more fields were added: "Message when unavailable" and "Assortment information".
> The first one will add a short description to the product item and the other one will set a
> detailed description on the product sheet. This second one is editable from the website editor.

what should happen if the field is left blank?

CarlosRoca13:
@francesco-ooops sorry for the delay

> tested with this setup, it does not seem to be working:

I think is not a problem of the module, the problem comes from the configuration of the assortment. You are trying with the public user, and this is a special user, because it's not active, you have to add it to the domain explicitly `['id', '=', <id of public user>]`

> what should happen if the field is left blank?

Should happen the same as **Don't apply restriction**

francesco-ooops:
> > what should happen if the field is left blank?
> 
> Should happen the same as **Don't apply restriction**

what is the downside of using directly "don't apply restriction"? 
It's confusing to have two options for the same behavior (and one of those not described)

francesco-ooops:
> @francesco-ooops sorry for the delay
> 
> > tested with this setup, it does not seem to be working:
> 
> I think is not a problem of the module, the problem comes from the configuration of the assortment. You are trying with the public user, and this is a special user, because it's not active, you have to add it to the domain explicitly `['id', '=', <id of public user>]`

Clear, thanks. Since it's a specific case, I think the most clear implementation would be a flag "apply for public user" (maybe in website availabilty options?), or if that's not possible just add it to the readme, what do you think?

francesco-ooops:
Hi @CarlosRoca13, any feedback to my remarks?

CarlosRoca13:
@francesco-ooops changes done, I add the boolean to add the public user and make the selection required to just have the 3 options defined

francesco-ooops:
Hi @CarlosRoca13 , is this module multi-website ready? I think a m2m where to choose which website to apply this assortment to would be good

francesco-ooops:
> Hi @CarlosRoca13 , is this module multi-website ready? I think a m2m where to choose which website to apply this assortment to would be good

@CarlosRoca13 what do you think?

CarlosRoca13:
> > Hi @CarlosRoca13 , is this module multi-website ready? I think a m2m where to choose which website to apply this assortment to would be good
> 
> @CarlosRoca13 what do you think?

Yes it's a good idea, I will add this feature

CarlosRoca13:
Done @francesco-ooops

francesco-ooops:
Thanks a lot! So I tested this setup (only active product assortment):

image

Correct me if I'm wrong: this means that partner Mitchell Admin, on Website 2, will only see 4 chairs products, right?

Given that no other product assortment is active, on Website 1 all products should be available for sale, but they aren't: 

image

I also think that using m2m tags for websites could help avoid duplication of assortment records: 

image

CarlosRoca13:
@francesco-ooops Now it should be working well, and changed m2o by m2m

francesco-ooops (approved):
@CarlosRoca13 great! 

Functional review ok

pedrobaeza (approved):

PR #596, opened 2021-12-17
[13.0][IMP] website_snippet_product_category: Change website published field by published in snippet field

sergio-teruel:
Odoo core does not allow publish or no a public category.
This module is using the field webiste_published to display categories into snippet.
This field creates confusion, the user think that if do this the category does not appear in left tree category.

cc @Tecnativa TT33532

ping @Tardo @CarlosRoca13

### DIFF
--- a/website_snippet_product_category/README.rst
+++ b/website_snippet_product_category/README.rst
@@ -39,7 +39,7 @@ You can select what categories can be shown by the snippet and an image for the 
 
 #. Go to website (backend) > Configuration > Products > eCommerce Categories
 #. Create or Edit one
-#. You can see two new options "Website Published" and "Category Image"
+#. You can see two new options "Published in product category snippet" and "Category Image"
 
 Usage
 =====
@@ -105,6 +105,7 @@ Contributors
   * Alexandre D. Díaz
   * Pedro M. Baeza
   * Carlos Roca
+  * Sergio Teruel
 
 Other credits
 ~~~~~~~~~~~~~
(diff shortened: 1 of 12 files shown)


website_snippet_product_category/migrations/13.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Active public categories based on previous is_published field
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE product_public_category
+        SET is_published_in_snippet = is_published
```
pedrobaeza: You need to rename the column in pre instead, or when getting here, it won't exist (normal ORM drops it).

Tardo (approved):
It's ok! maybe can be more specific: "Published in product category snippet"... no?

website_snippet_product_category/migrations/13.0.2.0.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Active public categories based on previous is_published field
+    """
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE product_public_category
+        SET published_in_product_category_snippet = is_published
```
pedrobaeza: Still the same problem here

sergio-teruel:
Changes Done!!

sergio-teruel:
Use rename_fields instead rename_columns..

PR #597, opened 2021-12-21
12.0 [ADD] website_sale_cart_expire

tack-yabeQuartile:
This code is taken from 13.0 branch's code.
We need to add this module also in 12.0 branch.

PR #598, opened 2021-12-21
[MIG] website_sale_cart_expire: Migration to 12.0

tack-yabeQuartile:
This module has been back-ported from 13.0 branch.

This PR supersedes [the PR #597](https://github.com/OCA/e-commerce/pull/597).


pedrobaeza:
It seems your commit with the migration to 12 is missing.

tack-yabeQuartile:
I think no changes are needed except for version expressions as long as I have checked the behavior in own 12.0 based odoo environment.

tack-yabeQuartile:
Error below happens now on runboat/build. 
Please let me know if you know what to do next.

```
+ pip install -r test-requirements.txt
ERROR: ./setup/website_sale_cart_expire is not a valid editable requirement.
It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
```

pedrobaeza:
If you want runboat working, the `setup` folder must be correctly generated, but it's not needed at all for the PR. It's only for testing in live.

kakurai8 (approved):
I checked it with runbot, and it worked as expected for the function.

- Set the time in website_sale_cart_expire.
- Add to cart from website
- The quotation SO is generated when viewed in sale.
- If the customer is still logged in when the time expires, the quote will not be cancelled.
- If the customer leaves (closes the page) after a certain period of time, the status is canceled and Logs remain in the chatter.

tack-yabeQuartile:
Thanks advise and review!

takuyatokunaga (approved):
It seems working as expcted,

If the customer is not active for the time set in Setting> General Setting> Website> Orders Followup
- The cart becomes "Empty"
- The SO status becomes "Canselled"

tack-yabeQuartile:
@pedrobaeza
Thanks advise me, and this PR can be merged now. 
Please proceed to merge this PR, otherwise let me know what I have to do.

In addition to above, there is a consultation.
We find that some tools (such a google chrome addon of "Tab Session Manager" to keep browser session) prevent auto canceling function of this module.
Is another OCA module like "Inactive Sessions Timeout" effective to avoid failure of canceling action? 
If yes, we will try it. if no, I think this is worthy to submit as issue and I will do so.
Please give me your opinion.

pedrobaeza:
The commit message is not correct. It should be `[MIG] website_sale_cart_expire: Migration to 12.0`, and a qualified reviewer should approve this for being merge.

About the other question, I don't know.

tack-yabeQuartile:
> The commit message is not correct. It should be `[MIG] website_sale_cart_expire: Migration to 12.0`, and a qualified reviewer should approve this for being merge.

Thank you for answers. I have changed the title. and I wait that a qualified reviewer check this PR.

> About the other question, I don't know.

OK, I got it. I try the OCA module and then submit an issue if it is needed.

pedrobaeza:
You have changed the PR title, not the migration commit message.

tack-yabeQuartile:
I pushed empty commit with migration commit message.
Thanks to give me advice!

pedrobaeza:
But it shouldn't be an empty/extra commit, but the previous one [replace version info from 13.0 to 12.0](https://github.com/OCA/e-commerce/pull/598/commits/e449ce765487d7a20eab1d6c7785df0d2c63659e)

chienandalu (commented):
> We find that some tools (such a google chrome addon of "Tab Session Manager" to keep browser session) prevent auto canceling function of this module.

It could be. The the `sale_get_order()` method in the `website` model tries to get an id from the session. Does it load cancelled or in draft mode?

tack-yabeQuartile:
> Does it load cancelled or in draft mode?

Our customer uses "Tab Session Manager" when he uses Odoo. His quotation remains in draft mode all the time, he said and we also confirmed.
After execute scheduled action manually, the state is not changed. "draft" state is kept from created till just now.

chienandalu:
Look that this module looks at the write date of the order to consider it expired. If an extension is constantly reloading the page in the users side, the order will be updated when the method `sale_get_order()` is called. The issue is in the customer's side. Not too much to do about it IMO.

tack-yabeQuartile:
@chienandalu
Thank you for your comment! We tell it the customer. 
And then we investigate ways to avoid automatic updates the order.

chienandalu:
> And then we investigate ways to avoid automatic updates the order.

That I can tell you it's not going to be easy :)

Another approach would be able to choose the criteria to consider a sale.order expired. For example, a website setting with the Date field to consider.

chienandalu:
Os this missing something to be merged?

PR #599, opened 2021-12-24
[13.0][FIX] website_snippet_carousel_product: Use flexible height

Tardo:
This prevents flattened images from being displayed.

** There is no difference between the new and the old because there are no elements that make the card expand
**New:**
cp_new

**Old:**
cp_old_t

cc @tecnativa


carlosdauden (approved):

chienandalu (approved):

pedrobaeza:
And what limits the height? Any other CSS? Is it recommended to have variable heights?

Tardo:
The maximum height is limited by the field being used (image_256)

PR #611, opened 2022-01-30
12 add website sale affiliate product restriction

fcayre:


pedrobaeza:
Isn't this the same as website_product_assortment that is proposed as a PR?

PR #615, opened 2022-02-04
[FIX] website_sale_hide_price : bad inherit_id for the product view

legalsylvain:
Trivial patch.

Rational : 

- the current ``xpath expr="//div[hasclass('css_quantity')]"`` intherit from ``website_sale.product`` view but ``css_quantity`` is defined in ``website_sale.product_quantity``

ref : https://github.com/odoo/odoo/blob/12.0/addons/website_sale/views/templates.xml#L602

regards.

CC : @fcayre

### DIFF
--- a/website_sale_hide_price/views/website_sale_template.xml
+++ b/website_sale_hide_price/views/website_sale_template.xml
@@ -21,6 +21,8 @@
                     website.website_show_price
                 </attribute>
             </xpath>
+        </template>
+        <template id="product" inherit_id="website_sale.product_quantity">
             <xpath expr="//div[hasclass('css_quantity')]" position="attributes">
                 <attribute name="t-if">
                     website.website_show_price


legalsylvain:
Note : CI is red because of other module. (bad inheritance also) in module ``website_sale_vat_required``. See runboat logs.

Note 2 : main 12.0 seems red also.

pedrobaeza:
Can you maybe fix it as well?

legalsylvain:
> Can you maybe fix it as well?

well, after investigation there is incompatibility between module ``website_sale_checkout_country_vat`` and ``website_sale_vat_required`` : both replace ``vat`` field...

- https://github.com/OCA/e-commerce/blob/12.0/website_sale_checkout_country_vat/views/templates.xml#L9
- https://github.com/OCA/e-commerce/blob/12.0/website_sale_vat_required/views/templates.xml#L5

however, I don't know how to fix that. (my first contribution on ``OCA/e-commerce`` + I don't use that modules.

Maybe authors could take a look ? @ernestotejeda, @Tardo, @sergio-teruel, @Tardo

Please let me know when main branch is green, I'll rebase that PR.

regards.

pedrobaeza:
Isn't this the same as #603 ?

PR #616, opened 2022-02-22
[FIX] website_sale_product_minimal_price: Integration with website_sale_b2x_alt_price

CarlosRoca13:
As now the minimum price is set asynchronously, we need to set manually the alt_price value given by the website_sale_b2x_alt_price module and avoid the same value in both prices.

Before change:
imagen

After change:
imagen

cc @Tecnativa TT34707

ping @sergio-teruel @chienandalu

### DIFF
--- a/website_sale_product_minimal_price/controllers/main.py
+++ b/website_sale_product_minimal_price/controllers/main.py
@@ -29,17 +29,21 @@ class WebsiteSaleVariantController(VariantController):
             combination = template._get_combination_info(
                 product_id=product_id, add_qty=add_qty, pricelist=pricelist
             )
-            res.append(
-                {
-                    "id": template.id,
-                    "price": combination.get("price"),
-                    "distinct_prices": has_distinct_price,
-                    "currency": {
-                        "position": template.currency_id.position,
-                        "symbol": template.currency_id.symbol,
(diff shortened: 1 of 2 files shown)


website_sale_product_minimal_price/controllers/main.py
```diff
@@ -29,17 +29,21 @@ def get_combination_info_minimal_price(self, product_template_ids, **kw):
...
+                "distinct_prices": has_distinct_price,
+                "currency": {
+                    "position": template.currency_id.position,
+                    "symbol": template.currency_id.symbol,
+                },
+            }
+            # As now the minimum price is set asynchronously, we need to set the second
+            # price value given by the website_sale_b2x_alt_price module and avoid the
+            # same value in both prices.
+            if combination.get("alt_price", False):
```
chienandalu: No need to fallback to `False` as no key gets you `None` which is falsey for the `if` as well

```suggestion
            if combination.get("alt_price"):
```

chienandalu (approved):
Minor tweak. Tested ok in any case. Thanks! :+1:

CarlosRoca13:
Change done @chienandalu

victoralmau (approved):

PR #617, opened 2022-02-23
[13.0][FIX/FW] website_sale_product_brand: image proportions and website.published mixin

chienandalu:
This PR includes:

- [x] A fix to correctly display the brands
- [x] A FW of #426 

Before this patch, images were deformed if they didn't meet precise
proportions.

Before:

image

After:

image

cc @Tecnativa TT34753

please review @Tardo @CarlosRoca13

### DIFF
--- a/website_sale_product_brand/views/product_brand.xml
+++ b/website_sale_product_brand/views/product_brand.xml
@@ -38,25 +38,19 @@
                         <t t-if="brand_rec">
                             <div class="row">
                                 <t t-foreach="brand_rec" t-as="o">
-                                    <div class="col-lg-3 d-lg-inline-block">
+                                    <div
+                                        class="card border-0 col-lg-3 col-md-6 mb-4 text-center brand_item"
+                                    >
                                         <a
                                             t-att-href="keep('/shop/brands',brand = o.id)"
                                         >
-                                            <div class="row text-center">
-                                                <div class="col">
-                                                    <img
-                                                        itemprop="image"
-                                                        width="200px"
-                                                        height="200px"
-                                                        class="img"
-                                                        t-attf-src="/website/image/product.brand/#{o.id}/logo#{'?max_width=300&amp;max_height=300'}"
-                                                    />
-                                                </div>
-                                            </div>
-                                            <div class="row text-center">
-                                                <div class="col">
-                                                    <t t-esc="o.name" />
-                                                </div>
+                                            <img
+                                                class="card-img-top img-fluid brand_item_image"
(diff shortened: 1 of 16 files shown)


website_sale_product_brand/views/product_brand.xml
```diff
@@ -38,25 +38,19 @@
...
-                                                    />
-                                                </div>
-                                            </div>
-                                            <div class="row text-center">
-                                                <div class="col">
-                                                    <t t-esc="o.name" />
-                                                </div>
+                                            <img
+                                                class="card-img-top img-fluid brand_item_image"
+                                                t-attf-src="/website/image/product.brand/#{o.id}/logo?max_width=200&amp;max_height=200"
```
pedrobaeza: Should we hard-code these max sizes instead letting the class to say?

chienandalu: Yes, it's unnecessary. Removed.

CarlosRoca13 (approved):

website_sale_product_brand/views/product_brand.xml
```diff
@@ -38,25 +38,19 @@
...
-                                                    />
-                                                </div>
-                                            </div>
-                                            <div class="row text-center">
-                                                <div class="col">
-                                                    <t t-esc="o.name" />
-                                                </div>
+                                            <img
+                                                class="card-img-top img-fluid brand_item_image"
+                                                t-attf-src="/website/image/product.brand/#{o.id}/logo"
```
Tardo, line 49: Wouldn't it be enough to use img-fluid and tell the controller to give us a 200x200 image?
```suggestion
                                                t-attf-src="/website/image/product.brand/#{o.id}/logo/200x200"
```

chienandalu: It doesn't work as that doesn't return a 200x200 canvas if the image isn't 1:1

chienandalu:
@Tardo can you update your review?

Tardo (commented):
In runboat it seems that the domain is not applied when you select a brand.

chienandalu:
What do you mean. I just tested it and it's working?

chienandalu:
> In runboat it seems that the domain is not applied when you select a brand.

Fixed :) It was a matter of setting the right context to avoid misbehaviors along with other modules

Tardo (approved):
Great!

pedrobaeza:
/ocabot merge minor

Please fw-port it to v14

chienandalu:
The merge bot didn't succed

PR #618, opened 2022-02-25
[MIG] website_sale_require_login: Migration to 15.0

rquevedo:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

BT-apadron:
In order, there are no changes in this Pull request. I created another Pull request with the migration where preserve commits history. https://github.com/OCA/e-commerce/pull/670

PR #620, opened 2022-03-02
12.0 add website sale payment slimpay

legalsylvain:


pedrobaeza:
Can't this be used outside e-commerce?

legalsylvain:
@fcayre : any point of view, regarding @pedrobaeza's remark ?

fcayre:
@pedrobaeza the e-commerce independent part is requested for addition here: https://github.com/OCA/account-payment/pull/457; this is the e-commerce part. cc @legalsylvain

pedrobaeza:
Well, other payment acquirer don't need special part for e-commerce. It seems also very weird to require something of connector-telephony for a payment acquirer, but it was just an observation. I don't plan for now to use this payment method.

fcayre:
@pedrobaeza I was asked to split my initial payment module into a website independent part (usable by ShopInvader for instance) and this one. This is not so weird for such SEPA-related payment method which can be used without e-commerce (requesting a mandate signature using Slimpay's web site and using it in an is_auto_pay contract for instance, is a possible use case for this), so I did it.
Second point regarding connector-telephony: I can probably make it work without the odoo module dependency and only with the phonenumbers python library, thanks for the remark, I'll make a try.

PR #621, opened 2022-03-03
[14.0][FIX] website_sale_product_description: Move the product description text closer to the image carousel

carmenbianca:
Previously, if the product details panel was very long, the description
text would start at the very bottom of that panel. This created a big
empty space between the image carousel and the product description.

Before:

Capture d’écran de 2022-03-03 15-31-29

After:

Capture d’écran de 2022-03-03 15-35-22

### DIFF
--- a/website_sale_product_description/views/website_sale_template.xml
+++ b/website_sale_product_description/views/website_sale_template.xml
@@ -2,7 +2,7 @@
 <odoo>
     <template id="product" inherit_id="website_sale.product">
         <xpath
-            expr="//div[@id='product_details']/../div[hasclass('col-md-6', 'col-xl-4')]"
+            expr="//div[@id='product_details']/../div[hasclass('col-md-6', 'col-xl-8')]/t"
             position="after"
         >
             <div class="o_not_editable">


carmenbianca:
An observation was made in https://github.com/coopiteasy/addons/pull/209#discussion_r818705462, that perhaps the expression could be simplified from `expr="//div[@id='product_details']/../div[hasclass('col-md-6', 'col-xl-8')]/t" position="after"` to `expr="//div[@id='o-carousel-product']" position="inside"` or some such. It would change the template from `website_sale.product` to `website_sale.shop_product_carousel`.

pedrobaeza:
Yeah, a simpler path is advisable. Probably when this was done in previous version, no better hook was available.

flotho:
Hi @carmenbianca 

is it a change request ?
> An observation was made in [coopiteasy/addons#209 (comment)](https://github.com/coopiteasy/addons/pull/209#discussion_r818705462), that perhaps the expression could be simplified from `expr="//div[@id='product_details']/../div[hasclass('col-md-6', 'col-xl-8')]/t" position="after"` to `expr="//div[@id='o-carousel-product']" position="inside"` or some such. It would change the template from `website_sale.product` to `website_sale.shop_product_carousel`.

carmenbianca:
Pushed the proposed change to the branch. Completely forgot about this PR.

pedrobaeza:
Are you proposing then a new PR?

carmenbianca:
@pedrobaeza No I think this PR needs to be re-opened. The changes strangely aren't reflected when clicking on 'Files changed', but they are definitely on the source branch.

pedrobaeza:
I'm not able to reopen it, so you have to resubmit the branch to a new PR I'm afraid.

PR #622, opened 2022-03-08
[14.0][IMP] website_sale_stock_available: make the available qty dependent on a defined field

enriquemartin:
The objective is to be able to define through UI on which field to use for calculating the quantity available on the website, by `product.template`.
It is interesting for example for manufactured products or with kit type BoMs.
The original behavior of the module is maintained by default, if no field is defined in the product.

### DIFF
--- a/website_sale_stock_available/__manifest__.py
+++ b/website_sale_stock_available/__manifest__.py
@@ -13,6 +13,8 @@
         "stock_available",
         "website_sale_stock",
     ],
-    "data": [],
+    "data": [
+        "views/product_views.xml",
+    ],
     "installable": True,
 }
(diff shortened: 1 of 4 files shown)


pedrobaeza:
This shouldn't be put at this level. This is relying on how `stock_available` is computing such field.

enriquemartin:
Thanks @pedrobaeza, but I don't understand. 
How to make `immediately_usable_qty` equal to `potential_qty`, for example? 
In addition, you may want to calculate the available for the website based on another criteria

pedrobaeza:
That's not the idea. The idea is that extra modules `stock_available_*` modifies what is stored in the field that is used here.

enriquemartin:
But in that case, if there is a `stock_available_*` module that modifies the value of `immediately_usable_qty`, it modifies for the whole environment right?
One of the goals here is to be able to specify another field just for this use on website, on the module that already exists.

pedrobaeza:
Then you should do that change on the base module or do another module for that. This one is strictly for taking such field.

PR #629, opened 2022-03-28
[FIX] website_sale_product_sort: README

remytms:
Quick fix in the example python code of the readme.

Append function does not return the list that it modifies in place.

If I remember well, the OCABot will update the Readme file when merging the PR, isn't it ?

### DIFF
--- a/website_sale_product_sort/readme/CONFIGURE.rst
+++ b/website_sale_product_sort/readme/CONFIGURE.rst
@@ -12,5 +12,6 @@ this:
 
     @api.model
     def _get_product_sort_criterias(self):
-      res = super()._get_product_sort_criterias()
-      return res.append(("default_code asc", _("Reference - A to Z")))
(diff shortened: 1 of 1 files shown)


chienandalu (approved):
> Append function does not return the list that it modifies in place.

Sure, silly one :smile: 

> If I remember well, the OCABot will update the Readme file when merging the PR, isn't it ?

Yes, although it's a good practice to avoid those administrative commits it's optional.

Rebase and catch the last core code that hopefully won't fail on travis with python 3.5

remytms:
Thanks the reply and for the review. :)

My branch is already based on the latest version of the 12.0 branch in this OCA repo. So I don't get where to find the last core code. Can you clarify what you are meaning ? :)

chienandalu:
Yes, I simply relaunched the travis jobs to get the new code upstream. Now they're green :)

PR #631, opened 2022-04-06
[MIG] website_sale_require_login: Migration to 14.0

suniagajose:


matteoopenf:
have you plan to fix the problem?

Issue #534, opened 2021-04-29
website_sale_secondary_unit wrong quantity calculation?


LuqueDaniel:
When adding a product to the cart by selecting a secondary unit of measure whose factor contains decimal values, the quantity isn't calculated correctly.

**Tested versions**: 12.0 and 13.0 both in Runbot.

### Steps to reproduce it and example

1. Go to inventory > Configuration > Settings and set the option Units of Measure *(This is required to use website_sale_secondary_unit)*
2. Create a product with the unit of measurement meters (m). 
3. Add to that product a secondary unit of measure based on meters and set a decimal factor (1.4 for example).
![imagen](https://user-images.githubusercontent.com/1286535/116567353-d6e30700-a907-11eb-9ea2-8a1d7ae1da13.png)
4. Publish the product.
![imagen](https://user-images.githubusercontent.com/1286535/116567816-42c56f80-a908-11eb-86b3-3490de27e05f.png)
5. Select the unit of measure (in the example Box 1.4m) and add it to the cart.
6. The calculation seems to only take the integer value.
![imagen](https://user-images.githubusercontent.com/1286535/116568711-fdee0880-a908-11eb-8526-26dee4c76cbb.png)

### Other tests

<details>
<summary>In a new sale order is displayed correctly</summary>

![imagen](https://user-images.githubusercontent.com/1286535/116569347-98e6e280-a909-11eb-8833-8b1cad6940f1.png)
</details>
<details>
<summary>Works well with integer quantity values</summary>

![imagen](https://user-images.githubusercontent.com/1286535/116570740-c8e2b580-a90a-11eb-9658-00d777a02a4a.png)

![imagen](https://user-images.githubusercontent.com/1286535/116569956-26c2cd80-a90a-11eb-9939-ce70ac3f698d.png)
</details>

chienandalu:
ping @sergio-teruel

sergio-teruel:
@LuqueDaniel Thanks for your issue... I will take a look!!!

LuqueDaniel:
I finally had time to look into it ^^

According to commit https://github.com/odoo/odoo/commit/a05a34366 Odoo doesn't support rounding of UOM in eCommerce.

[...]

PR #641, opened 2022-05-16
[FIX] website_sale_secondary_unit: add_qty rounding when not integer secondary unit factor

CarlosRoca13:
cc @Tecnativa

https://github.com/OCA/e-commerce/issues/534

check it please @sergio-teruel @carlosdauden


carlosdauden (approved):
Perhaps it would be better not to overwrite the original variable, but I approve of this change because this is not modified in the PR

sergio-teruel (approved):

PR #651, opened 2022-05-27
[MIG] website_sale_charge_payment_fee: Migration to 14.0

miguel-S73:
Standard migration to 14.0


miguel-S73:
@pedrobaeza Do you see this migration OK?

pedrobaeza:
I have removed the stale label, but sorry, I don't use this module. Please look for previous contributors / reviewers of it.

miguel-S73:
Hi @yostashiro , Can you review this changes, please?

yostashiro:
@miguel-S73 I am afraid I will not be able to review this PR anytime soon.  We did the 12.0 migration for a client who ended up not continuing with Odoo implementation for their business reasons, and we have not used the module since.

At one time we did consider to migrate the module to 15.0 for a client, and then we decided to take a simpler approach which does not depend on `website_sale`.  The module is [here](https://github.com/qrtl/mi7-custom/tree/15.0/payment_cod) but it is specifically for cash-on-delivery scenario.  I think it can be made more generic with some tweaking.

The possible disadvantage of this approach is that selecting the payment method does not instantly update the order content in the shop payment page, which is not important for our client (just showing the extra fee tag and an explanation is enough), while the advantage is to get rid of all the JS and controller hassle, making the design much simpler and more robust.  Please feel free to use the code in case you find it useful.

miguel-S73:
@yostashiro thank you for your suggestion but we need this approach.

miguel-S73:
Contributor @eLBati and
@OCA/e-commerce-maintainers
Can someone review this changes, please?

miguel-S73:
@pedrobaeza Can you reopen this PR? Once it's reopened, I will add myself as a contributor and maintainer to make it easier to merge.

pedrobaeza:
I can't reopen it. Please make another pull request with the same source branch instead.

PR #653, opened 2022-05-30
[13.0][FIX] website_sale_secondary_unit: Avoid show non published secondary_units

CarlosRoca13:
cc @Tecnativa

Before this changes, the non published and archived secondary_units were visible at e-comerce. With this changes we avoid to see them.

ping @sergio-teruel @carlosdauden

### DIFF
--- a/website_sale_secondary_unit/controllers/main.py
+++ b/website_sale_secondary_unit/controllers/main.py
@@ -33,3 +33,10 @@ class WebsiteSaleSecondaryUnit(WebsiteSale):
             set_qty=set_qty,
             display=display,
         )
+
+    def _prepare_product_values(self, product, category, search, **kwargs):
+        res = super()._prepare_product_values(product, category, search, **kwargs)
+        res["secondary_uom_ids"] = product.secondary_uom_ids.filtered(
+            lambda su: su.active and su.is_published
+        )
+        return res
--- a/website_sale_secondary_unit/views/templates.xml
+++ b/website_sale_secondary_unit/views/templates.xml
@@ -54,7 +54,7 @@
             expr="//div[@id='product_details']//t[@t-call='website_sale.product_price']"
             position="after"
         >
-            <t t-if="product.secondary_uom_ids">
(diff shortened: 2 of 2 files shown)


carlosdauden (approved):

website_sale_secondary_unit/controllers/main.py
```diff
@@ -33,3 +33,10 @@ def cart_update_json(
             set_qty=set_qty,
             display=display,
         )
+
+    def _prepare_product_values(self, product, category, search, **kwargs):
+        res = super()._prepare_product_values(product, category, search, **kwargs)
+        res["secondary_uom_ids"] = product.secondary_uom_ids.filtered(
+            lambda su: su.active and su.is_published
```
cesar-tecnativa, line 40: su.active is not needed

cesar-tecnativa (changes requested):

website_sale_secondary_unit/controllers/main.py
CarlosRoca13: If we don't set it, as you can see in the description, non published and archived secondary_units were visible at e-comerce

cesar-tecnativa: I thought that ORM made it by default when you use filtered

CarlosRoca13: Yes but the field has this context https://github.com/OCA/product-attribute/blob/c570ac4984c83bcae877208a9ec1ebcbfd707f1a/product_secondary_unit/models/product_template.py#L14 to allow seeing the archived secondary units at product level.

cesar-tecnativa: :+1:

cesar-tecnativa (approved):

sergio-teruel (approved):

pedrobaeza:
/ocabot merge minor

Please fw-port it to upper versions

pedrobaeza:
Check if fw-port is needed

PR #656, opened 2022-06-10
[15.0] [MIG] `website_sale_cart_expire`

ivantodorovich:
Standard migration plus a few extra feature and fixes, each in a separate commit to ease review and/or backporting


petrus-v (approved):
Thanks for this migration !

Code review looks good to me.

website_sale_cart_expire/models/website.py
```diff
@@ -0,0 +1,61 @@
...
+        expire_date = fields.Datetime.now() - timedelta(hours=self.cart_expire_delay)
+        return [
+            ("website_id", "=", self.id),
+            ("state", "=", "draft"),
+            ("write_date", "<=", expire_date),
+            # We don't want to cancel carts that are already in payment.
+            "|",
+            ("transaction_ids", "=", False),
+            "!",
+            ("transaction_ids.state", "in", ["pending", "authorized", "done"]),
```
petrus-v, line 35: It's common to see the negative term `"!"` used directly as it's translated by odoo from operator, so I would suggest but won't lock merging process for that 

```suggestion
            ("transaction_ids.state", "not in", ["pending", "authorized", "done"]),
```

petrus-v (approved):

website_sale_cart_expire/models/website.py
ivantodorovich: In this case it's not the same, though.

Using the `!` operator it means: _Orders that don't have any transaction in these states_
Using the `not in` operator it means: _Orders having transactions in different states than these_

ioans73 (approved):
LGTM 👍🏻

PR #657, opened 2022-06-13
[15.0][ADD] website_sale_stock_provisioning_time.

nikunjantala:


pedrobaeza:
There's already https://github.com/OCA/e-commerce/tree/13.0/website_sale_stock_provisioning_date. Migrate it instead.

jans23:
Both modules work a bit different. With module website_sale_stock_provisioning_date we couldn't display the expected provisioning time (in work days) on a particular product and it requires the expected stock move to be planned in Odoo. In our workflow we don't have such expected stock move but for some products we require several days to prepare an individual order before it can be shipped. We didn't find website_sale_stock_provisioning_date suitable to reflect such.

pedrobaeza:
OK, please reflect such differences in the README.

website_sale_stock_provisioning_time/models/product_template.py
```diff
@@ -0,0 +1,9 @@
+from odoo import fields, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    provisioning_time_in_workdays = fields.Integer(
+        string="Provisioning time in workdays"
```
manengel: workdays --> work days I think.

jans23: It seems both to be ok:

* https://dict.leo.org/englisch-deutsch/arbeitstage
* https://www.dict.cc/?s=arbeitstage
* https://www.deepl.com/translator#de/en/arbeitstage

manengel: O.K. then the module does look good to me.

pedrobaeza:
This pull request is incorrect containing a lot of undesired changes. Closing, and if re-submitted, it should be with another name that don't collapse with the existing one.

PR #659, opened 2022-06-25
[13.0] [FIX] website_sale_cart_expire: possible cron crash

gurneyalex:
Some modules may raise exceptions when a sale.order is canceled, and
a single order in this case will make the entire cron run rollback
and fail, causing an accumulation of expired carts.

I recently faced this situation with sale events where a miconfiguration
of an event caused the cancelation of sale orders to fail because the
remaining number of seats was negative.

This patch fixes the situation by calling action_cancel() individually
on each cart and checking for an ORM exception, which is logged together
with the cart database id, but not re-raised which allow other carts to
be processed by the cron.

### DIFF
--- a/website_sale_cart_expire/models/website.py
+++ b/website_sale_cart_expire/models/website.py
@@ -2,10 +2,13 @@
 # @author Iván Todorovich <ivan.todorovich@gmail.com>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
 from datetime import timedelta
 
-from odoo import _, api, fields, models
+from odoo import _, api, exceptions, fields, models
 from odoo.osv import expression
+
+_logger = logging.getLogger(__name__)
 
 
 class Website(models.Model):
@@ -37,9 +40,12 @@ class Website(models.Model):
             website._get_cart_expire_delay_domain() for website in websites
         ]
         carts_to_expire = self.env["sale.order"].search(
-            expression.OR(carts_to_expire_domains)
+            expression.OR(carts_to_expire_domains), order="id ASC",
         )
         # Expire carts
         for cart in carts_to_expire:
-            cart.message_post(body=_("Cart expired"))
-        carts_to_expire.action_cancel()
+            try:
+                cart.action_cancel()
+                cart.message_post(body=_("Cart expired"))
+            except exceptions.except_orm:
+                _logger.exception("Unable to cancel expired cart id %s", cart.id)
--- a/website_sale_cart_expire/models/website.py
+++ b/website_sale_cart_expire/models/website.py
@@ -45,7 +45,8 @@ class Website(models.Model):
         # Expire carts
         for cart in carts_to_expire:
             try:
-                cart.action_cancel()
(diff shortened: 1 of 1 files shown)


chienandalu (approved):

website_sale_cart_expire/models/website.py
```diff
@@ -37,9 +40,12 @@ def _scheduler_website_expire_cart(self):
...
         )
         # Expire carts
         for cart in carts_to_expire:
-            cart.message_post(body=_("Cart expired"))
-        carts_to_expire.action_cancel()
+            try:
+                cart.action_cancel()
+                cart.message_post(body=_("Cart expired"))
+            except exceptions.except_orm:
+                _logger.exception("Unable to cancel expired cart id %s", cart.id)
```
ivantodorovich: Just ignoring orm exceptions is not good. In case there's one, and in case `action_cancel` performed any write before raising the exception, we'd be commiting that too.

It's solved with a savepoint though:

```suggestion
            try:
                with self.env.cr.savepoint():
                    cart.action_cancel()
                    cart.message_post(body=_("Cart expired"))
            except Exception:
                _logger.exception("Unable to cancel expired cart id %s", cart.id)
```

ivantodorovich: It might be a good idea to autocommit after each cart process. This would help in case there are lots of carts to expire, and a limited cron cpu time to do it. Typically in this situation we'd be stuck in an endless cron loop that starts over and over trying to process always the same records.

```py
import threading

# ...

    @api.model
    def _scheduler_website_expire_cart(self):
        autocommit = not getattr(threading.currentThread(), "testing", False)

# ...
        for cart in carts_to_expire:
            # ...
            if autocommit:
                self.env.cr.commit()

**Review by ivantodorovich — CHANGES_REQUESTED** [review]

File: website_sale_cart_expire/models/website.py [hunk 2]
```diff
@@ -37,9 +42,25 @@ def _scheduler_website_expire_cart(self):
...
         )
         # Expire carts
-        for cart in carts_to_expire:
-            cart.message_post(body=_("Cart expired"))
-        carts_to_expire.action_cancel()
+        with api.Environment.manage():
+            for cart in carts_to_expire:
+                if autocommit:
+                    with registry(self.env.cr.dbname).cursor() as new_cr:
+                        new_env = api.Environment(
```
ivantodorovich, line 52: Hm, new cursors + new environments seems a bit excessive now, and redundant with the savepoint

Here's an example of what I meant about autocommit: https://github.com/odoo/odoo/blob/2387366379029d7bf9188e5465cfb4896f11e3b1/addons/event/models/event_mail.py#L169-L184

Naturallty, the cron needs to be changed as well, to pass the autocommit argument, example:
https://github.com/odoo/odoo/blob/2387366379029d7bf9188e5465cfb4896f11e3b1/addons/event/data/event_data.xml#L9

Basically we want this to be the default for cron execution, but not when called programatically and **definitely not** when running tests

Sorry in my previous comment I wasn't clear, I had to recheck this pattern

PR #666, opened 2022-07-20
[15.0][MIG] website_sale_comparison_hide_price: Migration to version 15.0

willytal:
cc @Tecnativa TT37331


website_sale_comparison_hide_price/views/website_sale_template.xml
```diff
@@ -25,11 +25,6 @@
                 website.website_show_price and not product.website_hide_price
             </attribute>
         </xpath>
-        <xpath expr="//form[@action='/shop/cart/update']" position="attributes">
```
pedrobaeza: You need to look for other appearances of something similar to this. In this version there are several mechanisms, and we can't allow to show the price in one place.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #672, opened 2022-08-13
[13.0][FIX] website_sale_stock_provisioning_date: reception in two steps + website warehouse

pedrobaeza:
2 fixes:

-----------------

[FIX] Cover reception in 2 steps

Steps to reproduce the problem:

- Configure your website warehouse to receipt in 2 steps.
- Do a purchase of an out of stock product and confirm it.
- Set the purchased product to show the provisioning date on your e-commerce.
- Navigate to the product e-commerce page.

Current behavior:

No provisioning date is shown.

Expected behavior:

The provisioning date of your reception is shown.

The problem comes from searching only moves with direct supplier location as source and internal location as destination. We search instead for chained moves that have original location in supplier ones.

-----------------

[FIX] Restrict website warehouse

If any warehouse is selected in your website, the condition to show or not the provisioning date should be restricted to such warehouse.

We use the provided method `_get_warehouse_available` to be consistent with what `website_sale_stock` performs to get the available stock.

@Tecnativa TT38566

### DIFF
--- a/website_sale_stock_provisioning_date/models/product_product.py
+++ b/website_sale_stock_provisioning_date/models/product_product.py
@@ -1,4 +1,5 @@
 # Copyright 2019 Tecnativa - Ernesto Tejeda
+# Copyright 2022 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import fields, models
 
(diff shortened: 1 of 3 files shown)


sergio-teruel (approved):

pedrobaeza:
Added commit for formatting date according user lang.

PR #676, opened 2022-08-30
[13.0][IMP] website_sale_stock_list_preview: Set context to compute as website_sale_stock_available

CarlosRoca13:
When website_sale_stock_available is installed, the grid view is showing
different allowed quantities than when we access to the products sheet.

With this changes we add compatibility with this module and if
stock_available_immediately is installed we will get same value.

cc @Tecnativa TT38689

please @victoralmau @pedrobaeza review this


victoralmau (approved):
Seems to be a change that makes sense.

pedrobaeza (approved):

PR #677, opened 2022-09-01
[13.0][IMP] website_sale_stock_list_preview: Make tests more resilient

CarlosRoca13:
cc @Tecnativa TT38689

ping @pedrobaeza @chienandalu


pedrobaeza:
Why they are more resilient now?

CarlosRoca13:
As the value of `available_threshold` was set by default if some module decrease this default, the test will fail, as the text rendered will be "In stock" instead of "5.000 Units available".

pedrobaeza:
Please put that on the commit message then.

CarlosRoca13:
Done @pedrobaeza

pedrobaeza (approved):

chienandalu (approved):

PR #678, opened 2022-09-03
[14.0][IMP] Update copier template + switch to GH actions

pedrobaeza:
@Tecnativa


pedrobaeza:
Gitlab is right now under maintenance...

PR #682, opened 2022-09-06
[ADD] website_sale_custom_filter

ajaniszewska-dev:


website_sale_custom_filter/readme/USAGE.rst
```diff
@@ -0,0 +1,31 @@
+Module allows to define product filters which can be used on website.
```
i-vyshnevska: to move to DESCRIPTION.rst as that is a primary section, at least one sentence there

TDu:
To fix the precommit error, you can use this command
`copier -f update`
Do that in another PR:

website_sale_custom_filter/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+{
+    "name": "eCommerce Product Filters",
+    "category": "Website/Website",
+    "sequence": 50,
+    "summary": "Advanced website product filters",
+    "website": "https://github.com/OCA/e-commerce",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "version": "15.0.0.1.0",
+    "development_status": "Alpha",
+    "depends": ["website", "sale", "mail", "website_sale", "base", "product"],
```
TDu:
```suggestion
    "depends": ["website", "sale", "mail", "website_sale", "product"],
```
We don't usually mention the `base` module if there is other dependencies

website_sale_custom_filter/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    ],
+    "demo": [],
+    "assets": {
+        "web.assets_frontend": [
+            "website_sale_custom_filter/static/src/scss/website_sale.scss",
+            "website_sale_custom_filter/static/src/js/website_sale_filters.js",
+        ],
+    },
+    "installable": True,
+    "application": True,
```
TDu:
```suggestion
    "application": False,
```

website_sale_custom_filter/security/ir.model.access.csv
```diff
@@ -0,0 +1,5 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_website_filter_public,website.filter.public,model_website_sale_custom_filter,,1,0,0,0
+access_website_filter_values_public,website.filter.values.public,model_website_sale_custom_filter_value,,1,0,0,0
```
i-vyshnevska, line 3: do we need full access for group_website_designer?

website_sale_custom_filter/models/website_sale_custom_filter_value.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+from odoo import api, fields, models
+
+
+class WebsiteSaleCustomFilterValue(models.Model):
+    _name = "website.sale.custom.filter.value"
+    _description = "website.sale.custom.filter.value"
+
+    @api.depends("value_filter_id", "value_filter_id.domain")
```
i-vyshnevska:
```suggestion
    @api.depends("value_filter_id.domain")
```
this will cover both

website_sale_custom_filter/security/ir.model.access.csv
ajaniszewska-dev: yes we need for that - as per module specification:
```
Only users from group website.group_website_designer should have CRUD rights on these models.
Public users must have READ access.
```

website_sale_custom_filter/models/website_sale_custom_filter.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2022 Camptocamp
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class WebsiteSaleCustomFilter(models.Model):
+    _name = "website.sale.custom.filter"
+    _description = "website.sale.custom.filter"
+    _inherit = "mail.thread"
```
TDu, line 10: Does it need to have the mail thread functionality ?

website_sale_custom_filter/models/website.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+
+from odoo import models
+
+
+class Website(models.Model):
+    _inherit = "website"
+
+    def get_filters(self, category=False):
```
TDu:
```suggestion
    def get_website_sale_custom_filters(self, category=False):
```

website_sale_custom_filter/models/website_sale_custom_filter_value.py
```diff
@@ -0,0 +1,45 @@
...
+    value_filter_id = fields.Many2one("ir.filters", string="Value filter ID")
+    selected_product_tmpl_ids = fields.Many2many(
+        "product.template",
+        string="Selected product template",
+        readonly=True,
+        compute="_compute_selected_product_templates",
+    )
+
+    html_color = fields.Char(
+        string="Color", help="Here you can set a specific HTML color.", default=""
```
TDu:
```suggestion
        string="Color", help="Here you can set a specific HTML color."
```

TDu (commented):
Some unit tests would be welcome

website_sale_custom_filter/models/website_sale_custom_filter.py
ajaniszewska-dev: according to module specification:
`Create new model website.sale.custom.filter (inheriting mail.thread) with following fields: `

website_sale_custom_filter/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "eCommerce Product Filters",
+    "category": "Website/Website",
+    "sequence": 50,
+    "summary": "Advanced website product filters",
+    "website": "https://github.com/OCA/e-commerce",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "version": "15.0.0.1.0",
```
simahawk:
```suggestion
    "version": "15.0.1.0.0",
```

website_sale_custom_filter/__manifest__.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2022 Camptocamp
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "eCommerce Product Filters",
+    "category": "Website/Website",
+    "sequence": 50,
```
simahawk: is sequence relevant?

website_sale_custom_filter/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+    "summary": "Advanced website product filters",
+    "website": "https://github.com/OCA/e-commerce",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "version": "15.0.0.1.0",
+    "development_status": "Alpha",
+    "depends": ["website", "sale", "mail", "website_sale", "product"],
```
simahawk:
```suggestion
    "depends": ["website_sale"],
```

i-vyshnevska:
good for testing

website_sale_custom_filter/__manifest__.py
grindtildeath:
```suggestion
    "depends": ["website_sale"],
```
Other modules are implicit dependencies of website_sale, so no need to list them.

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,133 @@
...
+            values["custom_checkbox_filters"] = self.get_custom_checkbox_filter_values()
+            # check if any custom filter changed so we use
+            # it as condition to show "clean filter" btn
+            values["custom_val_filter_changed"] = any(
+                [
+                    i in request.httprequest.args
+                    for i in ["cust_filter", "min_cust_filter", "max_cust_filter"]
+                ]
+            )
+        return super()._get_additional_shop_values(values)
```
grindtildeath: AFAICS on https://github.com/odoo/odoo/pull/102531/files#diff-83686977d1607bdb0a3e755e9d22ec4496c1dc4caebfff4bee3902ddd1b3e866R236-R238 `super` is returning an empty dict, so it looks to me like you should first call super and then update its values, or I'm missing something :thinking:

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,133 @@
...
+        # how to handle if field from filter is related to another model?
+        for num_filter in numerical_filters_available_on_website:
+            col_name = num_filter.numerical_filter_field_id.name
+            model_name = num_filter.numerical_filter_field_id.model
+            if (
+                model_name == "product.product"
+            ):  # there are fields related to product.template sometimes
+                model_name = "product.template"
+            query = f"""
+            SELECT MIN({col_name}), MAX({col_name}) FROM {model_name.replace('.','_')}"""
```
grindtildeath: :warning: the table can be different than replacing `.` with `_` as it's defined by the `_table` property. (I know it's not the case for the possible values, but better avoid using `replace` for this)
```suggestion
            SELECT MIN({col_name}), MAX({col_name}) FROM {request.env[model_name]._table}"""
```

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,133 @@
...
+                    value_filter_data[f_id] = [int(val)]
+        return value_filter_data
+
+    def get_custom_range_filter_values(self):
+        cust_filter_min_max_vals = self.get_min_max_custom_filters()
+        curr_website = request.env["website"].get_current_website()
+        numerical_filters_available_on_website = request.env[
+            "website.sale.custom.filter"
+        ].search(
+            [("website_ids", "=", curr_website.id), ("filter_type", "=", "numerical")]
```
grindtildeath:
```suggestion
            [("website_ids", "in", curr_website.id), ("filter_type", "=", "numerical")]
```
I guess you should use `in` instead of `=` if the filter is defined for multiple websites.

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,133 @@
...
+                "available_min_value"
+            ] = available_min_value
+            range_filter_data[curr_filter_id][
+                "available_max_value"
+            ] = available_max_value
+            range_filter_data[curr_filter_id]["min_value"] = available_min_value
+            range_filter_data[curr_filter_id]["max_value"] = available_max_value
+            if cust_filter_min_max_vals and cust_filter_min_max_vals.get(
+                curr_filter_id
+            ):
```
grindtildeath:
```suggestion
            if cust_filter_min_max_vals.get(curr_filter_id):
```
Nitpicking, but AFAICS `cust_filter_min_max_vals` will always be a dict, so no need to check if any key is defined in it.

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,133 @@
...
+                if cust_filter_min_max_vals[curr_filter_id].get("min"):
+                    range_filter_data[curr_filter_id][
+                        "min_value"
+                    ] = cust_filter_min_max_vals[curr_filter_id]["min"]
+                if cust_filter_min_max_vals[curr_filter_id].get("max"):
+                    range_filter_data[curr_filter_id][
+                        "max_value"
+                    ] = cust_filter_min_max_vals[curr_filter_id]["max"]
```
grindtildeath: Couldn't we split all of this into subfunctions? Something like:
```
def get_custom_range_filter_values(self):
    (...)
    for num_filter in numerical_filters_available_on_website:
        curr_filter_id = str(num_filter.id)
        range_filter_data[curr_filter_id] = {}
        range_filter_data[curr_filter_id].update(num_filter._prepare_filter_default_website_values())
        range_filter_data[curr_filter_id].update(self._prepare_filter_actual_website_values(cust_filter_min_max_vals.get(curr_filter_id)))
    return range_filter_data

def _prepare_filter_actual_website_values(self, min_max_values)
    values = {}
    if min_max_values.get("min"):
        values["min_value"] = min_max_values["min"]
    if min_max_values.get("max"):
        values["max_value"] = min_max_values["max"]
    return values
```
And then on `website.sale.custom.filter` you could have:
```
def _prepare_filter_default_website_values(self):
    self.ensure_one()
    values = {}
    col_name = self.numerical_filter_field_id.name
    model_name = num_filter.numerical_filter_field_id.model
    if (
        model_name == "product.product"
    ):  # there are fields related to product.template sometimes
        model_name = "product.template"
    query = f"""
    SELECT MIN({col_name}), MAX({col_name}) FROM {self.env[model_name]._table}"""
    self.env.cr.execute(query)
    available_min_value, available_max_value = self.env.cr.fetchone()

    values["available_min_value"] = available_min_value
    values["available_max_value"] = available_max_value
    values["min_value"] = available_min_value
    values["max_value"] = available_max_value
    return values
```
I feel it makes the code much more readable with shorter variable names and dedicated functions.

website_sale_custom_filter/models/website_sale_custom_filter_value.py
```diff
@@ -0,0 +1,45 @@
...
+        "website.sale.custom.filter",
+        ondelete="cascade",
+        required=True,
+        string="Filter ID",
+    )
+    # display = fields.Selection(
+    #     [("checkbox", "Checkbox")],
+    #     default="checkbox",
+    #     string="Display type",
+    # )
```
ajaniszewska-dev: note: this is because we decided we need to separate color values from checkbox values and treat them as separate filter type, not filter values. So we don't need to distinguish between display type anymore as it can be only one type and there is no choice.

website_sale_custom_filter/models/product.py
```diff
@@ -0,0 +1,37 @@
...
+from odoo import api, models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    @api.model
+    def _search_get_detail(self, website, order, options):
+        res = super()._search_get_detail(website, order, options)
+        # modify base_domain to include selected filters
```
grindtildeath: Would be good to explain a bit why you select the first element in `base_domain`.

website_sale_custom_filter/models/product.py
```diff
@@ -0,0 +1,37 @@
...
+                filter_id = CustomFilterObj.browse(int(f_id))
+                filtering_field = filter_id.numerical_filter_field_id.name
+                min_val = val_id.get("min_value", False)
+                max_val = val_id.get("max_value", False)
+                # build domain
+                if min_val and min_val != val_id["available_min_value"]:
+                    domain.append((filtering_field, ">=", min_val))
+                if max_val and max_val != val_id["available_max_value"]:
+                    domain.append((filtering_field, "<=", max_val))
+        res["base_domain"][0] = domain
```
grindtildeath: IMO appending elements into the domain is a dangerous way to extend an existing domain.
I would instead use `osv.expression.AND` and `osv.expression.OR` to do it safely.
(I again feel that splitting the function would make it more readable)
```
@api.model
def _search_get_detail(self, website, order, options):
    res = super()._search_get_detail(website, order, options)
    original_domain = res["base_domain"][0]
    new_domain_elements = []

    if options.get("custom_checkbox_filters", False):
        checkbox_filter_domain = self._get_checkbox_filter_domain(options.get("custom_checkbox_filters"))
        if checkbox_filter_domain:
            new_domain_elements.append(checkbox_filter_domain)

    if options.get("custom_value_filters", False):
        value_filter_domain = self._get_value_filter_domain(options.get("custom_value_filters"))
        if value_filter_domain:
            new_domain_elements = .append(value_filter_domain)
    if new_domain_elements
    res["base_domain"][0] = AND(original_domain + new_domain_elements)
    return res

def _get_checkbox_filter_domain(self, option_values):
    subdomains = []
    for _, val_id in options["custom_checkbox_filters"].items():
        filter_value = CustomFilterValueObj.browse(val_id)
        selected_products = filter_value.selected_product_tmpl_ids.ids
        if selected_products:
            subdomains.append([("id", "in", selected_products)])
    return OR(subdomains)

def _get_value_filter_domain(self, option_values):
    subdomains = []
    for f_id, val_id in options["custom_value_filters"].items():
        filter_id = CustomFilterObj.browse(int(f_id))
        filtering_field = filter_id.numerical_filter_field_id.name
        min_val = val_id.get("min_value", False)
        max_val = val_id.get("max_value", False)
        # build domain
        if min_val and min_val != val_id["available_min_value"]:
            subdomains.append([(filtering_field, ">=", min_val)])
        if max_val and max_val != val_id["available_max_value"]:
            subdomains.append([(filtering_field, "<=", max_val)])
    return AND(subdomains)
```

grindtildeath (commented):
I didn't had time yet to check JS, tests, or view, but at least here are some improvement suggestions for controllers and models

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,119 @@
...
+        req = dict(request.httprequest.args)
+        min_max_filter_vals = {}
+        if req.get("min_cust_filter", False):
+            for vals in req["min_cust_filter"].split(","):
+                f_id, f_min_val = vals.split("_")
+                f_min_val = int(f_min_val)
+                if not min_max_filter_vals.get(f_id, False):
+                    min_max_filter_vals[f_id] = {"min": f_min_val}
+                else:
+                    min_max_filter_vals[f_id]["min"] = f_min_val
```
simahawk:
```suggestion
            for vals in req.get("min_cust_filter", "").split(","):
                f_id, f_min_val = vals.split("_")
                f_min_val = int(f_min_val)
                if not min_max_filter_vals.get(f_id, False):
                    min_max_filter_vals[f_id] = {"min": f_min_val}
                else:
                    min_max_filter_vals[f_id]["min"] = f_min_val
```
and you can indent back everything by one tab, same for the block below

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,119 @@
...
+                if not min_max_filter_vals.get(f_id, False):
+                    min_max_filter_vals[f_id] = {"max": f_max_val}
+                else:
+                    min_max_filter_vals[f_id]["max"] = f_max_val
+        return min_max_filter_vals
+
+    def get_custom_checkbox_filter_values(self):
+        req = dict(request.httprequest.args)
+        value_filter_data = {}
+        if req.get("cust_filter"):
```
simahawk: same as above

website_sale_custom_filter/controllers/main.py
```diff
@@ -0,0 +1,119 @@
...
+        cust_filter_min_max_vals = self.get_min_max_custom_filters()
+        curr_website = request.env["website"].get_current_website()
+        numerical_filters_available_on_website = request.env[
+            "website.sale.custom.filter"
+        ].search(
+            [("website_ids", "in", curr_website.id), ("filter_type", "=", "numerical")]
+        )
+
+        for num_filter in numerical_filters_available_on_website:
+            curr_filter_id = str(num_filter.id)
```
simahawk: do you really need this string conversion?

website_sale_custom_filter/models/website_sale_custom_filter_value.py
simahawk: to trash then?

website_sale_custom_filter/static/src/js/website_sale_filters.js
```diff
@@ -0,0 +1,118 @@
...
+        events: {
+            'change input[type="checkbox"]': "_onChangeCustomAttribute",
+        },
+
+        /**
+         * @private
+         * @param {Event} ev
+         */
+        _onChangeCustomAttribute(ev) {
+            const search = $.deparam(window.location.search.substring(1));
```
simahawk, line 48: Not a blocker but you can use URLSearchParam both to parse and to build the url later

website_sale_custom_filter/controllers/main.py
simahawk: Agree

simahawk (changes requested):

website_sale_custom_filter/controllers/main.py
ajaniszewska-dev: we get filter id as part of  string from request so either i cast here as string or after retrieving data from request to int - changed to 2nd option.

website_sale_custom_filter/controllers/main.py
ajaniszewska-dev: afaics in worst case this will return list with empty string which gets evaluated as a True so here i also need to add if-cond. to be sure vals is True (non-empty string), otherwise we end up with code trying to process this string.
```
(Pdb) req.get("min_cust_filter","").split(",")
['']
(Pdb) bool(req.get("min_cust_filter","").split(","))
True

ajaniszewska-dev:
note: applied filters are only adding new products to result, not narrowing down the subset of products selected by previously applied filters -> to adjust implementation.

PR #692, opened 2022-11-14
Sync repo template

ajaniszewska-dev:


TDu (approved):

.t2d.yml
```diff
@@ -0,0 +1,42 @@
+language: python
```
simahawk: I'm not sure this file is still needed

PR #712, opened 2022-12-20
[13.0][IMP] website_sale_secondary_unit: mobile UX

chienandalu:
For mobile users adding an extra column in the cart summary makes that the others content won't fit very well, specially for products with long names. As we can infer the sale line information without the secondary units computation computation column we choose to hide it if the screen is too small as Odoo does with the product image column:

Peek 20-12-2022 16-55

cc @Tecnativa TT37137

please review @sergio-teruel @pedrobaeza

### DIFF
--- a/website_sale_secondary_unit/views/templates.xml
+++ b/website_sale_secondary_unit/views/templates.xml
@@ -166,10 +166,10 @@
             </t>
         </xpath>
         <xpath expr="//th[hasclass('td-qty')]" position="after">
-            <th class="text-center td-qty">Quantity</th>
+            <th class="text-center td-qty d-none d-md-table-cell">Quantity</th>
         </xpath>
         <xpath expr="//td[hasclass('td-qty')]" position="after">
-            <td class="text-center td-qty">
+            <td class="text-center td-qty d-none d-md-table-cell">
                 <span
                     t-esc="int(line.product_uom_qty) == line.product_uom_qty and int(line.product_uom_qty) or line.product_uom_qty"
                 />


website_sale_secondary_unit/static/src/scss/website_sale_secondary_unit.scss
```diff
@@ -1,7 +1,14 @@
...
-.oe_website_sale input.js_secondary_quantity {
-    min-width: 48px;
-    text-align: center;
+.oe_website_sale {
+    input.js_secondary_quantity {
+        min-width: 48px;
+        text-align: center;
+    }
+    @include media-breakpoint-down(sm) {
+        .td-psu {
```
pedrobaeza: Do you need this extra class? Isn't there a standard one for this?

pedrobaeza (approved):

website_sale_secondary_unit/static/src/scss/website_sale_secondary_unit.scss
chienandalu: There's td-img which isn't very generic, I'm afraid.

pedrobaeza: I was thinking on Bootstrap `sm-`, `xs-` classes that are intended for that.

chienandalu: Ok, yes. Re-review the new simpler approach.

PR #716, opened 2022-12-27
[16.0][MIG] website_sale_hide_price: Migration to 16.0

JrAdhoc:


chienandalu (approved):
Tested. Looks good :+1: Thanks!

pedrobaeza (commented):
Something that I have discovered that is not probably of the migration itself is that if you have attributes with extra price, they are still shown:

Selección_002

so please cover also this option, or at least put it on the roadmap section.

JrAdhoc:
@pedrobaeza thanks for your feedback. I added a t-if on that element with the same condition as the other elements present in that view. Alredy ammended the PR aswell

pedrobaeza (approved):
/ocabot merge nobump

tricorebv:
Hey, I noticed the price is still showing up to Public User when searching for a product in the search bar.

Screenshot 2023-02-03 at 10 39 07

Am I missing something to hide it?

chienandalu:
@tricorebv sure. It looks like an issue with the search box. Can you open the issue or propose a patch?

PR #721, opened 2023-01-04
[ADD] website_sale_attribute_filter_range

miguel-S73:
reopen  https://github.com/OCA/e-commerce/pull/486

Reopened PR and I applied first change suggestion from @pedrobaeza  in this comment https://github.com/OCA/e-commerce/pull/486#pullrequestreview-980936335


chienandalu (commented):
The behavior doesn't seem to be solid. Only a product should be filtered in this example:

Peek 04-01-2023 09-30

miguel-S73:
@chienandalu 
Whether you apply the filter or remove it, the filter form is submitted. When the filter is submitted, all the filter values are passed and if there are products without an attribute, as the case is  in your example, these products don't appear.

zeroheure:
Why not reusing the price filter JS from Odoo ? (just curious)

PR #725, opened 2023-01-12
[REF] website_sale_secondary_unit: Reuse existing JS variable

SirTakobi:
Get quantity from already declared variable instead of searching it again


SirTakobi:
CI should be fixed after https://github.com/OCA/e-commerce/pull/754

PR #733, opened 2023-01-23
[15.0][MIG] website_sale_require_legal: Migration to version 15.0

pilarvargas-tecnativa:
cc @Tecnativa TT38355


pilarvargas-tecnativa:
I think the tests will continue to fail because it does not store the terms and conditions acceptance message in the partner or in the sales order. I think it is not a local problem.

website_sale_require_legal/static/tests/tours/tour.js
```diff
@@ -0,0 +1,75 @@
...
+        },
+        {
+            trigger: "a[href='/shop/cart']",
+            extra_trigger: "sup.my_cart_quantity:contains('1')",
+        },
+        {
+            trigger: ".btn:contains('Process Checkout')",
+        },
+        {
+            trigger: "a[href='/shop/checkout']",
```
CarlosRoca13:
```suggestion
            trigger: ".progress-wizard-steplabel:contains('Address')",
```

Try to correct the test by changing this step to this other

CarlosRoca13 (changes requested):

pedrobaeza:
Is this module still needed? Core has something for this if I don't remember bad.

CarlosRoca13:
I'm not sure at all... 

The module `website_legal_page` is the one that could be merged with `account`, but I don't like this solution, because `account` in a website without sales is not required, but we would like to add a legal page for possible logins

pedrobaeza:
I'm not talking about `account`, but `website_sale` includes a custom view to put legal conditions.

pilarvargas-tecnativa:
This module adds:
That it is necessary to accept the terms and conditions when defining an address.
Log a note in the partner and in the sale order when such terms are accepted.

website_sale_require_legal/static/tests/tours/tour.js
pilarvargas-tecnativa: The error occurs again at the same step

CarlosRoca13: So I'm not seeing the same error... Last error was related to this step https://github.com/OCA/e-commerce/pull/733/files#diff-23db7de503e2d4676b7fe9f495e9cdc5f4ed0e0304b51b68f6407766dc4a2062R27 now it is related to this other https://github.com/OCA/e-commerce/pull/733/files#diff-23db7de503e2d4676b7fe9f495e9cdc5f4ed0e0304b51b68f6407766dc4a2062R56

pilarvargas-tecnativa: You are right, the previous error was repeated because I had changed the code in the wrong step, when I have done it in the corresponding step the error is different, however it is a different error than the one I have in the local installation.

website_sale_product_attribute_filter_category/readme/DESCRIPTION.rst
```diff
@@ -1,4 +1,4 @@
 This module extends the functionality of website sale module to allow to group
 product attributes filter by their categories.
-Also you can select if you want display folded or un folded category in website
+Also you can select if you want display folded or unfolded category in website
```
chienandalu: This slipped into your branch from other migration

website_sale_product_attribute_filter_category/views/templates.xml
```diff
@@ -44,12 +44,12 @@
                                     <i class="fa fas" />
                                     <span t-out="attr_category[1]" />
                                 </a>
-                                <ul
+                                <div
```
chienandalu: Also from other  PR

chienandalu (changes requested):
Careful there:

chienandalu (approved):
:smile: :+1:

CarlosRoca13 (approved):

PR #736, opened 2023-01-25
[ADD] website_sale_hide_empty_category: backport from 15.0

ilyasProgrammer:


pedrobaeza:
It must include the commit history, and this module is very old:

https://odoo-community.org/shop/website-sale-hide-empty-categories-4469#attr=4426

It's not from v15.

PR #738, opened 2023-01-26
[15.0][FIX] website_sale_product_attribute_filter_category: Removed products_attributes_collapsible template + migration script

pilarvargas-tecnativa:
The products_attributes_collapsible template has been removed because the functionality of the module is to add the attribute categories in the filter section. This template added the feature of collapsible attributes that should be independent.

A new module has been created to fold attributes independently: https://github.com/OCA/e-commerce/pull/740

cc @Tecnativa TT41345

@chienandalu @pedrobaeza could you check that everything is correct?

ping @stefan-tecnativa


website_sale_product_attribute_filter_category/migrations/15.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,17 @@
...
+    env["ir.ui.view"].search(
+        [
+            (
+                "key",
+                "=",
+                "website_sale_product_attribute_filter_category"
+                ".products_attributes_collapsible",
+            )
+        ]
+    ).unlink()
```
chienandalu: During upgrade we try to minimize ORM operations:
```suggestion
    openupgrade.logged_query(
        env.cr,
        """
        DELETE FROM ir_ui_view
        WHERE key = 'website_sale_product_attribute_filter_category.products_attributes_collapsible'
        AND website_id IS NOT NULL
        """,
    )
```

chienandalu (approved):
:+1:

website_sale_product_attribute_filter_category/migrations/15.0.1.0.0/pre-migration.py
pilarvargas-tecnativa: changes made! thanks for the suggestion :)

pedrobaeza (approved):
/ocabot merge major

chienandalu:
Oh it isn't fine! The migration script won't match the version!

pedrobaeza:
There's no problem. The module was just merged, and this is for the migration path.

I launch again the command:

/ocabot merge major

PR #744, opened 2023-02-02
[15.0][FIX] website_sale_attribute_filter_category: Added configuration option in menu "/shop/Customize".

pilarvargas-tecnativa:
Add a configuration option to customise the filtering area by attributes. When the module is installed, the only option to deactivate the categories in the filtering area is to uninstall it, so adding a configuration option will allow you to activate or deactivate the way in which the filtering area is displayed without having to uninstall it. 
By default the option will be active to keep the same behaviour when installing this module.
This arises because of the conflict with the tests of another module and not being able to deactivate this functionality from them.

cc @Tecnativa TT41498

@chienandalu @CarlosRoca13 please review :


chienandalu (approved):

pedrobaeza:
Shouldn't it be active=False by default for avoiding problems in the integration tests?

pilarvargas-tecnativa:
> Shouldn't it be active=False by default for avoiding problems in the integration tests?

It shouldn't be a problem since having the configuration option allows you to disable it from other modules, something you couldn't do before.

pedrobaeza:
But tests are not aware of other installed modules, so I see a problem here. It should be disabled IMO, and add in the CONFIGURE.rst file about the new option.

pilarvargas-tecnativa:
> But tests are not aware of other installed modules, so I see a problem here. It should be disabled IMO, and add in the CONFIGURE.rst file about the new option.

You are right, I check it and make the change.

website_sale_product_attribute_filter_category/readme/USAGE.rst
```diff
@@ -1,3 +1,4 @@
 #. Go to Website Shop.
 #. Active filter by attributes in "Customize" option in top menu.
+#. Active attributes categories filters in "Customize" option in top menu.
```
pedrobaeza:
```suggestion
#. Activate filter by attributes in "Customize" option in top menu.
#. Activate attributes categories filters in "Customize" option in top menu.
```

pilarvargas-tecnativa: Corrected, thanks

CarlosRoca13 (approved):

Issue #739, opened 2023-01-26
[website_sale_invoice_address] - display billing address as "billing" in checkout


francesco-ooops:
**Is your feature request related to a problem?**

This module changes invoice address in SO for ecommerce orders, as per [description](https://github.com/OCA/e-commerce/tree/15.0/website_sale_invoice_address#usage); however, in checkout user still sees partner's address as billing address. 

This can be confusing as it could lead the customer to believe that their order is being invoiced at the wrong address.

![image](https://user-images.githubusercontent.com/55489873/214904562-a402e800-8f03-4fbb-b409-8cb5792fcd16.png)

**Describe the solution you'd like**
Display the billing address that will be applied in SO to user during checkout.

chienandalu:
Thanks  @francesco-ooops I haven't researched it but my guess is that it comes from a qweb template context. A patch proposal will be welcome as always :smile: :+1:

PR #748, opened 2023-02-08
[IMP] display partner invoice  address as billing in checkout

Aldeigja:
Changed website_sale_invoice_address.confirmation template, to display the billing address that will be applied in SO to user during checkout. As described in Issue [#739](https://github.com/OCA/e-commerce/issues/739)

### DIFF
--- a/website_sale_invoice_address/views/templates.xml
+++ b/website_sale_invoice_address/views/templates.xml
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="website_sale_invoice_address.confirmation"
+        inherit_id="website_sale.confirmation"
+        priority="100"
+    >
+
+
+        <xpath expr="//span[hasclass('address-inline')]" position="attributes">
+            <attribute name="t-esc">
+                order.partner_invoice_id
+            </attribute>
+        </xpath>
+
+        <xpath expr="//t[@t-set='same_shipping']" position="attributes">
+            <attribute name="t-value">
+                bool(order.partner_shipping_id==order.partner_invoice_id or only_services)
+            </attribute>
+        </xpath>
+
+
+    </template>
+
+    <template
+        id="website_sale_invoice_address.payment"
+        inherit_id="website_sale.payment"
+        priority="100"
+    >
+
+
+        <xpath expr="//span[hasclass('address-inline')]" position="attributes">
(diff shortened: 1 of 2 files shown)


francesco-ooops (changes requested):
@Aldeigja functional not ok, billing address in checkout is not the one set in user's parent company

image

chienandalu (commented):
I think you have to modify the `website_sale.checkout` template as well

website_sale_invoice_address/views/templates.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="website_sale_invoice_address.confirmation"
+        inherit_id="website_sale.confirmation"
+        priority="100"
+    >
+
+        <xpath expr="//span[hasclass('address-inline')]" position="replace">
```
chienandalu: Why not simply change the attribute itself?
```suggestion
        <xpath expr="//span[hasclass('address-inline')]" position="attributes">
```

website_sale_invoice_address/views/templates.xml
```diff
@@ -0,0 +1,39 @@
...
+
+    </template>
+
+    <template
+        id="website_sale_invoice_address.payment"
+        inherit_id="website_sale.payment"
+        priority="100"
+    >
+
+        <xpath expr="//span[hasclass('address-inline')]" position="replace">
```
chienandalu:
```suggestion
        <xpath expr="//span[hasclass('address-inline')]" position="attributes">
```

chienandalu (changes requested):

website_sale_invoice_address/views/templates.xml
Aldeigja: I've made the corrections, thanks for your review!

francesco-ooops:
@Aldeigja tested, on address selection field it's displayed correctly, but if user tries to edit billing address, the address being edited is the one of the child contact itself 

[VIDEO](https://recordit.co/EQrqBrFGIu)

@chienandalu any suggestions on how to handle this?

chienandalu:
@francesco-ooops I think @Aldeigja pushed more changes. Did you check it after that?

francesco-ooops:
Hi @chienandalu , our customer dropped this module so in the end this fix is not complete

PR #754, opened 2023-02-15
12.0 update copier

victor-champonnois:
@pedrobaeza


victor-champonnois:
How am I supposed to fix the following issue ?
```
Run manifestoo -d . check-dev-status --default-dev-status=Beta
website_sale_resource_booking (Production/Stable) depends on sale_resource_booking (Beta)
```
Should I put website_sale_resource_booking as beta ?

pedrobaeza:
Better to change `website_sale_resource_booking` to `Beta`.

victor-champonnois:
@pedrobaeza arghh, two modules are incompatible (they both replace the "vat" field in the views), so I excluded them in the manifests. Now the tests don't pass.

pedrobaeza:
Put them on the `rebel_module_groups` question for forcing them to be installed and tested separately. I think putting one of them is enough.

carmenbianca (approved):
LGTM

victor-champonnois:
@gurneyalex @gfcapalbo @simahawk this is ready to be merged

.copier-answers.yml
```diff
@@ -0,0 +1,27 @@
+# Do NOT update manually; changes here will be overwritten by Copier
+_commit: v1.14.1
+_src_path: https://github.com/OCA/oca-addons-repo-template.git
+ci: GitHub
+dependency_installation_mode: OCA
```
SirTakobi: The default is PIP (https://github.com/OCA/oca-addons-repo-template/blob/fe5b6f5c35c3d13822d080a9c9e20df209f3a6c5/copier.yml#L114), is there any specific reason why OCA is selected instead?

.copier-answers.yml
```diff
@@ -0,0 +1,27 @@
...
+github_enable_makepot: true
+github_enable_stale_action: true
+github_enforce_dev_status_compatibility: true
+include_wkhtmltopdf: false
+odoo_version: 12.0
+org_name: Odoo Community Association (OCA)
+org_slug: OCA
+rebel_module_groups:
+- website_sale_checkout_country_vat
+- website_sale_vat_required
```
SirTakobi, line 20: In Travis configuration only **website_sale_vat_required** had a dedicated test environment, why with Copier is it needed for **website_sale_checkout_country_vat** too?

SirTakobi (changes requested):
Thanks for the PR!
Just asking for clarification for a few obscure changes

.copier-answers.yml
victor-champonnois: no reason. It must have been a mistake. It's fixed

.copier-answers.yml
victor-champonnois: I don't really how github CI works, but having website_sale_checkout_country_vat in the common test environment didn't work. `website_sale_vat_required` and `website_sale_checkout_country_vat` are not compatible (due to a "replace" position in the views)

SirTakobi: If the modules are not compatible, that should be stated in the modules (with `excludes` keyword in the manifest), not just the CI configuration.

Still seems strange that they are not compatible just now, but Travis didn't use to complain.
What do you mean with _didn't work_? Is there an error shown in some log?

carmenbianca:
bump to un-stale

pedrobaeza:
There's still some red condition.

carmenbianca:
I'll have a look at removing the red.

carmenbianca:
Hopefully fixed, copy-pasting commit message below.

Should I mark these two modules incompatible somewhere in the manifest? Not sure how to do that. Five minutes of duckduckgo'ing haven't resulted in me finding an 'incompatible' field.

---

[FIX] website_sale_checkout_country_vat: Make simultaneously installable with website_sale_vat_required

These need to be simultaneously installable for runboat to function.

The problem: They both edit/replace the 'vat' element in the address
form in the e-commerce. website_sale_checkout_country_vat replaces it
with a country dropdown thing, and website_sale_vat_required replaces it
with a near-identical element that is mandatory to fill in instead of
optional.

_Technically_ both modules eventually end up with a <input ...
name="vat"/> element which they are both replacing, which means that the
modules would be replacing the same element, theoretically making them
both simultaneously installable. However,
website_sale_checkout_country_vat calls a template to create that
element, and Odoo isn't quite smart enough to do a replace on an element
created by a template call.

The solution to the problem is to make the existing vat element
invisible in website_sale_checkout_country_vat. This means that
website_sale_vat_required can easily find an element to replace.

In any case, these modules are _not_ compatible with one another without
spending some effort.

One downside of this fix is that there are now two 'vat' elements on the
page: one visible created by the template call and one marked invisible.
The former one is listed first, and therefore used when getting the
data. However, there may be a scenario where this could break stuff,
maybe. I just haven't found it.

carmenbianca:
OK I found the 'excludes' keyword for the manifest, but it breaks runboat, because it tries to install both.

PR #764, opened 2023-03-01
[13.0][FIX] website_sale_hide_price: hide the price in the add_to_cart option

pilarvargas-tecnativa:
When the "add_to_cart" option was activated, the add to cart button was not hidden and the product could be added and this functionality has been corrected.

I have also added a fix for the "Buy Now Button" option which displayed a "Buy now" button in the product view which allowed the product to be added.

cc @Tecnativa TT41911


website_sale_hide_price/views/website_sale_template.xml
```diff
@@ -29,22 +29,29 @@
         </xpath>
         <xpath expr="//a[@id='add_to_cart']" position="attributes">
             <attribute name="t-if">
-                    website.website_show_price and not product.website_hide_price
-                </attribute>
+                website.website_show_price and not product.website_hide_price
+            </attribute>
```
chienandalu: Why this tab changes? :thinking:

website_sale_hide_price/views/website_sale_template.xml
```diff
@@ -29,22 +29,29 @@
...
         </xpath>
     </template>
     <template id="website_search_box" inherit_id="website.website_search_box">
         <xpath expr="//input[@name='search']" position="attributes">
-            <attribute
-                name="t-att-data-display-price"
-            >website and website.website_show_price</attribute>
+            <attribute name="t-att-data-display-price">
+                website and website.website_show_price
+            </attribute>
```
chienandalu: Here only the format is changed as well. Is it something prettier is doing?

website_sale_hide_price/views/website_sale_template.xml
```diff
@@ -29,22 +29,29 @@
...
-                    website.website_show_price and not product.website_hide_price
-                </attribute>
+                website.website_show_price and not product.website_hide_price
+            </attribute>
+        </xpath>
+    </template>
+    <template id="products_add_to_cart" inherit_id="website_sale.products_add_to_cart">
+        <xpath expr="//a[hasclass('a-submit')]" position="attributes">
+            <attribute name="t-if">
+                website.website_show_price and not product.website_hide_price
```
chienandalu, line 55: This isn't working

chienandalu (changes requested):

website_sale_hide_price/views/website_sale_template.xml
pilarvargas-tecnativa: I use a shortcut on my code to tabulate whole blocks (ctrl+shift+i), normally it only tabulates the block that is selected, maybe it was my mistake. Should I change it back to the way it was?

chienandalu: Yes, as these are useless changes that hide the proper fix

website_sale_hide_price/views/website_sale_template.xml
pilarvargas-tecnativa: Check again with the new change to another template to see if it works because in my local it works fine, if not we will have to check it in another way.

chienandalu (changes requested):
Ok, what I'm seeing is that this change isn't compatible with the module `website_sale_stock_force_block` which changes the same attributes in the `website_sale.products_add_to_cart` template...

As we're not currently mantaining that module, put a Known issue in the readme remarking such incompatibility

chienandalu (approved):
Squash your commits anyway :smile: :+1: 

please review @sergio-teruel

CarlosRoca13 (approved):
Code review :)

sergio-teruel (approved):

PR #772, opened 2023-03-14
[15.0][FIX] website_sale_hide_price, hide price on results of search bar

ChrisOForgeFlow:
When user search a product on website search bar, results render prices of products with website_hide_price field active

### DIFF
--- a/website_sale_hide_price/models/product_template.py
+++ b/website_sale_hide_price/models/product_template.py
@@ -38,3 +38,19 @@ class ProductTemplate(models.Model):
             }
         )
         return combination_info
+
+    def _search_render_results(self, fetch_fields, mapping, icon, limit):
+        """Hide price on the search bar results"""
+        results_data = super()._search_render_results(
+            fetch_fields, mapping, icon, limit
+        )
+        for product, data in zip(self, results_data):
+            if product.website_hide_price:
+                data.update(
+                    {
+                        "price": "<span>%s</span>"
+                        % (product.website_hide_price_message or ""),
+                        "list_price": "",
+                    }
+                )
+        return results_data


website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,18 @@ def _get_combination_info(
             }
         )
         return combination_info
+
+    def _search_render_results(self, fetch_fields, mapping, icon, limit):
+        res = super()._search_render_results(fetch_fields, mapping, icon, limit)
+        new_result = []
+        for data in res:
+            product = self.env["product.template"].browse(data.get("id"))
```
chienandalu: It looks like in the original method you can rely on zipping de results along with the recordset so you don't have to browse it again (this will underperform over large amounts of records):

https://github.com/odoo/odoo/blob/fa525cbcf63c4e35f224f6df41e9bc9fe356c7b7/addons/website_sale/models/product_template.py#L391

website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,18 @@ def _get_combination_info(
...
+            if product.website_hide_price:
+                data.update(
+                    {
+                        "price": "<span>%s</span>"
+                        % (product.website_hide_price_message or "")
```
chienandalu: Should we be aware of list_price as well?

website_sale_hide_price/models/product_template.py
ChrisOForgeFlow: Done!, thank you

website_sale_hide_price/models/product_template.py
ChrisOForgeFlow: It seems to me that it could cause problems if that value is used for some later computation, it is better to just hide the legend for the end user

ChrisOForgeFlow: Done! You are right should be erased in same way

website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,21 @@ def _get_combination_info(
             }
         )
         return combination_info
+
+    def _search_render_results(self, fetch_fields, mapping, icon, limit):
+        results_data = super()._search_render_results(
+            fetch_fields, mapping, icon, limit
+        )
+        new_result = []
```
chienandalu: You don't need to create a new list as you're update the original references so you can safely return `results_data`
```suggestion
```

website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,21 @@ def _get_combination_info(
...
+        for product, data in zip(self, results_data):
+            if product.website_hide_price:
+                data.update(
+                    {
+                        "price": "<span>%s</span>"
+                        % (product.website_hide_price_message or ""),
+                        "list_price": "<span>%s</span>"
+                        % (product.website_hide_price_message or ""),
```
chienandalu: This is only used when the discount is shown (like in before/after price), so we don't need to repeat the message, just leaving the key empty.
```suggestion
                        "list_price": ""),
```

website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,21 @@ def _get_combination_info(
             }
         )
         return combination_info
+
+    def _search_render_results(self, fetch_fields, mapping, icon, limit):
```
chienandalu, line 42:
```suggestion
    def _search_render_results(self, fetch_fields, mapping, icon, limit):
        """Hide price on the search bar results""""
```

chienandalu (changes requested):

website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,21 @@ def _get_combination_info(
...
+        """Hide price on the search bar results""""
```
chienandalu: Too many quote marks :sweat_smile: 
```suggestion
        """Hide price on the search bar results"""
```

website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,21 @@ def _get_combination_info(
...
+        for product, data in zip(self, results_data):
+            if product.website_hide_price:
+                data.update(
+                    {
+                        "price": "<span>%s</span>"
+                        % (product.website_hide_price_message or ""),
+                        "list_price": ""),
+                    }
+                )
+            new_result.append(data)
```
chienandalu:
```suggestion
```

chienandalu (changes requested):

website_sale_hide_price/models/product_template.py
```diff
@@ -38,3 +38,19 @@ def _get_combination_info(
...
+        results_data = super()._search_render_results(
+            fetch_fields, mapping, icon, limit
+        )
+        for product, data in zip(self, results_data):
+            if product.website_hide_price:
+                data.update(
+                    {
+                        "price": "<span>%s</span>"
+                        % (product.website_hide_price_message or ""),
+                        "list_price": ""),
```
chienandalu:
```suggestion
                        "list_price": "",
```

chienandalu (approved):
Thanks!

chienandalu:
@pilarvargas-tecnativa can you review?

pilarvargas-tecnativa (approved):
Reviewed, thanks!

chienandalu:
@ChrisOForgeFlow Will you forward port the patch to 16.0?

PR #781, opened 2023-04-05
[15.0][FIX] website_sale_product_attribute_filter_collapse: fix test error

pilarvargas-tecnativa:
website_folded is a field of the product.attribute.category model that is added in the website_sale_product_attribute_filter_category module.
website_sale_product_attribute_filter_collapse is an independent module and gives an error in this field, so the solution is to remove it when creating an attribute category in the tests.

cc @Tecnativa TT41385

@pedrobaeza @chienandalu please review this :)


chienandalu (approved):
:+1:

pedrobaeza (approved):
/ocabot merge patch

PR #783, opened 2023-04-19
[16.0][MIG] website_sale_vat_required

Abranes:
Migration to 16.0


chienandalu (changes requested):
Check tours

Abranes:
> Check tours

@chienandalu I modified the test according to Odoo 16.0 `shop_buy_product` tour test:
https://github.com/odoo/odoo/blob/83613b56cadf1a12331db9d4d41ab688d68768e9/addons/website_sale/tests/test_sale_process.py#L69-L79

As you can see, Odoo now requires the `payment_custom` module to execute the `shop_buy_product` tour, which means that in our case, this test would not be executed since we do not depend on that module.

chienandalu:
> As you can see, Odoo now requires the `payment_custom` module to execute the `shop_buy_product` tour, which means that in our case, this test would not be executed since we do not depend on that module.

But then we're not performing the tests at all and we won't detect regressions automatically. If the Odoo core tour is useless now, we should discard it and do our own

website_sale_vat_required/tests/test_website_sale_vat_required.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests.common import HttpCase
+
+
+class TestWebsiteSaleVatRequired(HttpCase):
+    def test_website_sale_vat_required(self):
+        if self.env["ir.module.module"]._get("payment_custom").state != "installed":
```
pilarvargas-tecnativa, line 8: Functionality tested locally, thank you!
I agree with David, the tests should be changed as this step can be avoided to test this module without relying on the payment methods module. @Abranes what do you think about making the change?

pilarvargas-tecnativa:
Superseed by: https://github.com/OCA/e-commerce/pull/841

PR #790, opened 2023-05-02
[I18N] website_sale_b2x_alt_price: update translation (Spanish)

pilarvargas-tecnativa:
cc @Tecnativa TT42540

@CarlosRoca13 @pedrobaeza please review


pedrobaeza:
Talked directly. It should be done on Weblate.

PR #787, opened 2023-04-25
[16.0][MIG] website_sale_checkout_skip_payment

rajpatel2240-odoo:

PR #796, opened 2023-05-10
[16.0][mig] website_sale_checkout_skip_payment: Migration to 16.0

dsolanki-initos:


pedrobaeza:
This is red. #802 is green, but later than this one. There's another previous PR in #787

PR #797, opened 2023-05-10
[16.0][FIX] website_sale_product_attribute_value_filter_existing: match signature

HekkiMelody:
match the function signature of shop with the original one from website_sale, to avoid potential issues with other modules overwriting the function and passing unexpected parameters


pedrobaeza (approved):
Checked in https://github.com/odoo/odoo/blob/a7be194f731370db0d8b0a2075e67676708fbd94/addons/website_sale/controllers/main.py#L318, so fast-tracking for fixing the potential side effect:

/ocabot merge patch

PR #800, opened 2023-05-17
[14.0] product_template_multi_link: improve UX by searching templates on default_code

sebalix:


product_template_multi_link/models/product_template.py
```diff
@@ -103,3 +104,37 @@ def get_links_by_code(self, code):
...
+                            [("default_code", operator, name)],
+                            [(self._rec_name, operator, name)],
+                        ]
+                    ),
+                ]
+            )
+            # Reset 'name' so base '_name_search' won't add '_rec_name'
+            # to 'args' (already added above).
+            # See 'odoo.models.BaseModel._name_search'.
+            name = ""
```
sebalix, line 134: See https://github.com/odoo/odoo/blob/14.0/odoo/models.py#L1814-L1815

product_template_multi_link/models/product_template.py
```diff
@@ -103,3 +104,37 @@ def get_links_by_code(self, code):
...
+    def _name_search(
+        self, name, args=None, operator="ilike", limit=100, name_get_uid=None
+    ):
+        # NOTE: Odoo limits the search on the name of templates as soon as
+        # the 'id' field is present in 'args' domain, and this 'id' criteria is
+        # set by the view on purpose to avoid a search on variants.
+        # Improve this search by looking also on template's default_code
+        # if there is only a domain on 'id'.
+        search_default_code = self.env.context.get("name_search_default_code")
+        if name and len(args or []) == 1 and args[0][0] == "id" and search_default_code:
```
sebalix, line 118: See https://github.com/odoo/odoo/blob/14.0/addons/product/models/product_template.py#L475-L478

TDu (approved):

product_template_multi_link/models/product_template.py
grindtildeath, line 134: Actually you could simplify your code by returning the result of `_search` in your function as it's what Odoo does in lots of places when a `name_search` has to consider multiple fields.

Something along these lines could work just as well:
```
    @api.model
    def _name_search(self, name, args=None, operator='ilike', limit=100, name_get_uid=None):
        search_default_code = self.env.context.get("name_search_default_code")
        if search_default_code and not (operator == 'ilike' and not (name or '').strip()):
            args = args or []
            domain = ['|', ('name', operator, name), (search_default_code, operator, name)]
            return self._search(expression.OR([domain, args]), limit=limit, access_rights_uid=name_get_uid)
        return super()._name_search(name, args=args, operator=operator, limit=limit, name_get_uid=name_get_uid)
```
at least it's easier from v16.0, cf: https://github.com/odoo/odoo/pull/86588

EDIT: I ought to check your link first, as with the heavy override in the standard, might just be safer to call `super` as you did.

grindtildeath (approved):
Not blocking

sebalix:
Added another use case: search on templates with default codes of variants (like Odoo std behavior) as the default code of a template is empty as soon as it gets some variants.

product_template_multi_link/models/product_template.py
```diff
@@ -103,3 +104,38 @@ def get_links_by_code(self, code):
...
+                    expression.OR(
+                        [
+                            [("default_code", operator, name)],
+                            [("product_variant_ids.default_code", operator, name)],
+                            [(self._rec_name, operator, name)],
+                        ]
+                    ),
+                ]
+            )
+            # Reset 'name' so base '_name_search' won't add '_rec_name'
```
simahawk, line 131: :1st_place_medal:

simahawk (approved):

simahawk:
@sebalix pre-commit is not happy. And don't forget to squash on the next push :pray:

sebalix:
@simahawk done

PR #804, opened 2023-05-25
[16.0][MIG] website_sale_firstname

HekkiMelody:
Migration from v11 to v16.

In the process of migrating, I rewrote almost the entire controller part to better fit odoo 16, and I replaced the fields in the form as well.

Based on https://github.com/OCA/partner-contact/tree/16.0/partner_firstname


HekkiMelody:
I can't figure out how to test the `address()` function itself.

Every time I try to do something like:
```
self.WebsiteSaleController = WebsiteSaleFirstname() # The controller in my module
with MockRequest(self.env, website=self.website) as req:
            req.httprequest.method = "GET"
            self.WebsiteSaleController.address()
```

I get an error when calling `address` that ends with:

```sh
  File "/opt/odoo/custom/src/odoo/odoo/http.py", line 697, in route_wrapper
    if routing['type'] == 'http':  # _generate_routing_rules() ensures type is set
KeyError: 'type'
```

It seems somehow related to the empty `@http.route` decorator...

chienandalu:
> I can't figure out how to test the `address()` function itself.
> 
> Every time I try to do something like:
> 
> ```
> self.WebsiteSaleController = WebsiteSaleFirstname() # The controller in my module
> with MockRequest(self.env, website=self.website) as req:
>             req.httprequest.method = "GET"
>             self.WebsiteSaleController.address()
> ```
> 
> I get an error when calling `address` that ends with:
> 
> ```shell
>   File "/opt/odoo/custom/src/odoo/odoo/http.py", line 697, in route_wrapper
>     if routing['type'] == 'http':  # _generate_routing_rules() ensures type is set
> KeyError: 'type'
> ```
> 
> It seems somehow related to the empty `@http.route` decorator...

I'd try playing with the new `path` and `routing` keywords for `MockRequest` (https://github.com/odoo/odoo/blame/1e96c21eed9556f87bb8dae2384580dd784f12c0/addons/website/tools.py#L18)

Honeyxilia:
> I can't figure out how to test the `address()` function itself.
> 
> Every time I try to do something like:
> 
> ```
> self.WebsiteSaleController = WebsiteSaleFirstname() # The controller in my module
> with MockRequest(self.env, website=self.website) as req:
>             req.httprequest.method = "GET"
>             self.WebsiteSaleController.address()
> ```
> 
> I get an error when calling `address` that ends with:
> 
> ```shell
>   File "/opt/odoo/custom/src/odoo/odoo/http.py", line 697, in route_wrapper
>     if routing['type'] == 'http':  # _generate_routing_rules() ensures type is set
> KeyError: 'type'
> ```
> 
> It seems somehow related to the empty `@http.route` decorator...

If you wish, you could also set up a Werkzeug Client, then post directly on the `/shop/address` endpoint with the `post` Werkzeug client, and be able to retrieve a HTTP response from which you could verify whether the form validation passed or not c:

website_sale_firstname/data/formbuilder_whitelist.xml
Honeyxilia, line 1: To better maintain history, it would have been nice to do `git mv website_sale_firstname/data/data.xml website_sale_firstname/data/formbuilder_whitelist.xml` c:

website_sale_firstname/views/templates.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 Alessandro Uffreduzzi - PyTech SRL -->
+<odoo>
+    <template id="website_sale_address_firstname" inherit_id="website_sale.address">
```
Honeyxilia, line 4: In this template overload, I think it would be good to also override the `field_required`, in order to replace `phone,name` by `firstname,lastname,phone`

Honeyxilia (commented):
Other than the 2 remarks I posed, it's LGTM (code review + local testing)

website_sale_firstname/data/formbuilder_whitelist.xml
HekkiMelody: Hello! Thanks for the review

I double-checked because I remembered doing `git mv` at the time. I did the process again.

The moment I also fix the formatting (from 8 spaces to 4 spaces) and `git add` again, git loses track of the fact that it was moved, probably because too high a % of the file has changed (even though it's just some spaces that were removed...)

I ended up splitting it into two commits (one for the move, one for the formatting) so history is preserved!

website_sale_firstname/views/templates.xml
HekkiMelody: Thank you for pointing it out. Added in a different commit (since it's a slight change to the logic compared to the previous module)

HekkiMelody:
@OCA/e-commerce-maintainers can this be reopened please? thank you!

PR #809, opened 2023-06-01
[16.0][MIG] website_sale_require_legal

Highcooley:


Highcooley:
This module is obsolete in V16.0 since it can be activated by default now.

chienandalu:
Well, it can still have it's public as this module targets the address form step and records the acceptance metadata. In Odoo core, the acceptance is made in the payment step and no metadata of the acceptance is recorded (as far as I recall)

PR #814, opened 2023-06-15
[ADD][16.0] new module website_sale_search_fuzzy_multiword

PicchiSeba:
I am submitting a module that allows to use multi-word keywords in the website search.

As of now we are successfully using this module in some applications where the website default language is not English.
But I am encountering problems when the website language is set to en_US.

Nevertheless I am submitting this module anyway because I think it would benefit the community once we figure out where the problem lies.


elvise:
@PicchiSeba can you rebase ?

zeroheure:
This is interesting. Why is it blocked ?

PicchiSeba:
@zeroheure 
Thank you for the interest!

We require more feedback in order to proceed. Would you like to try the module and tell us if you find any problem?

FrancescoBallerini:
> @zeroheure Thank you for the interest!
> 
> We require more feedback in order to proceed. Would you like to try the module and tell us if you find any problem?

Hello, we'd like to try the module, could you rebase for us? Thank you

PicchiSeba:
Hi @FrancescoBallerini, I rebased as requested.

Please let me know if this could interest you, or if not, what problems you incurred into. Thanks in advance for the feedback

FrancescoBallerini (commented):
Thank you for rabasing @PicchiSeba , I can confirm that your module is improving the research, I have tests it with en_UK because as you already pointed out it raises an error on en_US language. 

Unfortunately I cannot give any consistent hint on the error as I don't have a running odoo16 instance to debug locally. It seems to be related to translation and/or possibly, missing record names or translation for that record which causes an improper comparation between a falsy value and the jsonb object which contains translated terms. Well this is already specified by the error log but I cannot tell more than this as I generally work on version 14 and translation system changed a lot between versions.

That being said, I can confirm that your module is improving the search ! A couple of screenshot from my tests:

WIthout fuzzy multiword installed:

Screenshot 2024-04-13 151128

With fuzzy multiword installed:

Screenshot 2024-04-13 150930

I will leave a pending review for now, will approve this for sure if you manage to solve the issue with the default installed language! Thank you

PR #819, opened 2023-07-11
[16.0][FIX] website_sale_hide_price display: price in search bar and update cache

### LINKED ISSUE #818: [16.0] website_sale_hide_price
I have an unexplained behavior when using website_sale_hide_price on v16.0 (that I reproduce on Runboat).

It seems that Odoo keeps a cache on server side of requested pages (list of products or specific product) and does not update depending on whether I have the right to see the prices or not...
Restarting Odoo resets the behavior

## Module
website_sale_hide_price

## Describe the bug
The show / hide price are not properly applied

## To Reproduce
**Affected versions**: 16.0

1. Install website_sale_hide_price (+ configure a few products to be displayed on shop)
2. Configuration archived public user partner to not diplay prices
3. Disconnect
4. Go to shop page
5. As this is the first time you load this page, the behavior is as expected : no price on product list, no price when clicking on a product
6. Connect with portal user (or internal user)
7. Go to shop page
8. The prices are not displayed on pages that were visited before with public user, but are shown on pages which were not visited before

The same behaviour is observed when doing the opposite : displaying first pages with logged in user then disconnect : the prices are still there.
You can empty cache, reload page, use private navigation, even change navigator, you get the same behavior until you restart Odoo server...

**Expected behavior**
Pages should be updated depending on connected user

remi-filament:
Before this commit, the price was still displayed in search bar for users that are not allowed to see price (the check was only done on boolean from product_template)

Also, there was a problem with cache causing the list of products and individual products to not being updated (fixes #818)

### DIFF
--- a/website_sale_hide_price/models/product_template.py
+++ b/website_sale_hide_price/models/product_template.py
@@ -44,8 +44,11 @@ class ProductTemplate(models.Model):
         results_data = super()._search_render_results(
             fetch_fields, mapping, icon, limit
         )
+        website_show_price = (
+            self.env["website"].get_current_website().website_show_price
+        )
         for product, data in zip(self, results_data):
-            if product.website_hide_price:
+            if product.website_hide_price or not website_show_price:
                 data.update(
                     {
                         "price": "<span>%s</span>"
--- a/website_sale_hide_price/views/website_sale_template.xml
+++ b/website_sale_hide_price/views/website_sale_template.xml
@@ -1,5 +1,18 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
+    <template id="products" inherit_id="website_sale.products">
+        <!-- adding website and website.website_show_price to t-cache in order to update cache if parameter modified
+             this is needed if you connect with a different user that should not see the prices
+             see https://github.com/OCA/e-commerce/issues/818 -->
+        <xpath
+            expr="//div[hasclass('o_wsale_products_grid_table_wrapper')]"
+            position="attributes"
+        >
+            <attribute
+                name="t-cache"
+            >pricelist,products,website,website.website_show_price</attribute>
+        </xpath>
+    </template>
     <template id="product_price" inherit_id="website_sale.product_price">
         <xpath expr="//div[@itemprop='offers']" position="attributes">
             <attribute name="t-if">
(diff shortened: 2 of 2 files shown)


website_sale_hide_price/models/product_template.py
```diff
@@ -44,8 +44,13 @@ def _search_render_results(self, fetch_fields, mapping, icon, limit):
         results_data = super()._search_render_results(
             fetch_fields, mapping, icon, limit
         )
+        website_show_price = False
+        if self.env.context.get("website_id"):
+            website_show_price = (
+                self.env["website"].get_current_website().website_show_price
+            )
```
chienandalu: Can be simplified
```suggestion
        website_show_price = self.env["website"].get_current_website().website_show_price
```

chienandalu (approved):
Thanks @remi-filament Tiny nitpicking:

website_sale_hide_price/models/product_template.py
remi-filament: Done, thanks for the tip !

website_sale_hide_price/views/website_sale_template.xml
```diff
@@ -1,5 +1,15 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
+    <template id="products" inherit_id="website_sale.products">
+        <xpath
+            expr="//div[hasclass('o_wsale_products_grid_table_wrapper')]"
+            position="attributes"
+        >
+            <attribute
+                name="t-cache"
+            >pricelist,products,website,website.website_show_price</attribute>
```
pedrobaeza, line 13: Why is this needed? A comment should help to clarify for avoiding to remove it in future revisions.

pedrobaeza (approved):
/ocabot merge minor

website_sale_hide_price/views/website_sale_template.xml
remi-filament: This is needed to reload cache in case website_show_price changes (this is a compute which depends on the user being connected whether he is allowed to see prices are not)
Without this, the product list is kept in cache (with or without prices), so you get the exact same list when connecting with different users which do not have the same rights to see price (you get the prices (until next Odoo restart) if the first user to request page after restart of Odoo is allowed to see prices, no price if the first user to request page is not allowed)
This fixes the issue #818 
I added short explanation and link to issue in XML file as suggested

PR #822, opened 2023-08-01
Update website_sale_template.xml

a-khalil-ecoservice:
fix the following error message:

Element '<xpath expr="//a[hasclass(&#39;o_we_buy_now&#39;)]">' cannot be located in parent view

### DIFF
--- a/website_sale_hide_price/views/website_sale_template.xml
+++ b/website_sale_hide_price/views/website_sale_template.xml
@@ -45,10 +45,10 @@
     </template>
     <!-- Hide buy now button if "Buy Now Button" option is enabled and not website_show_price -->
     <template id="product_buy_now" inherit_id="website_sale.product_buy_now">
-        <a 
-            role="button" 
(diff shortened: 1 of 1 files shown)


chienandalu (commented):
Thanks @a-khalil-ecoservice can you give steps to reproduce the issue in a runboat?

a-khalil-ecoservice:
@chienandalu 
when I will to update the modul "website_sale_hide_price" I get the error.

chienandalu:
> @chienandalu when I will to update the modul "website_sale_hide_price" I get the error.

It looks to me that an old website_view architecture is the one bugging your upgrade. The current code is ok, what you need to do is either:

- update your website copies of "website_sale.o_we_buy_now" to the main template code
- make the proper adaptations in your theme/custom

IMO this PR is not valid

pleirb:
Hello, how are you all?
Something similar is happening to me, could it have anything to do with it?

imagen

PR #823, opened 2023-08-02
[13.0][FIX] website_sale_stock_available_display: don't show negatives

chienandalu:
Previous implementation intended this but the value was never shown as we give the previously formatted one.

cc @Tecnativa TT44655

@CarlosRoca13 @sergio-teruel plesae review


sergio-teruel (approved):

sergio-teruel:
port to v15?? I not sure.. @CarlosRoca13

CarlosRoca13 (approved):
@sergio-teruel In version 15.0, this module is merged with website_sale_stock, and when the `free_qty` is less or equal than 0 the message shown is defined on `out_of_stock_message`, so it would not show negatives.

LGTM :smile: Thanks

PR #825, opened 2023-08-09
[16.0][MIG] website_sale_product_attribute_filter_collapse: Migration to version 16.0

pilarvargas-tecnativa:
cc @Tecnativa TT44382

@chienandalu @CarlosRoca13 please review :)


pedrobaeza:
/ocabot migration website_sale_product_attribute_filter_collapse

website_sale_product_attribute_filter_collapse/views/templates.xml
```diff
@@ -5,112 +5,60 @@
     <template
         id="products_attributes_collapsible"
         inherit_id="website_sale.products_attributes"
-        name="Collapsible attribute filters"
```
chienandalu: Why removing this?

website_sale_product_attribute_filter_collapse/views/templates.xml
```diff
@@ -5,112 +5,60 @@
...
-        active="False"
```
chienandalu: Well, this changes the default behavior of the module. You should keep it IMO

chienandalu (changes requested):

website_sale_product_attribute_filter_collapse/views/templates.xml
pilarvargas-tecnativa: I was adapting the configuration to the web editor but I deleted more than I should have. It is now restored as it was and the behaviour has been corrected to be the same. thanks

chienandalu (approved):
Now is great :slightly_smiling_face: :+1:

website_sale_product_attribute_filter_collapse/__manifest__.py
```diff
@@ -3,14 +3,15 @@
 {
     "name": "Website Sale Attribute Filter Collapse",
     "summary": "Allows the attributes of the categories to be folded",
-    "version": "15.0.1.1.1",
+    "version": "16.0.1.1.1",
```
CarlosRoca13:
```suggestion
    "version": "16.0.1.0.0",
```

CarlosRoca13 (commented):
Please set correct version

website_sale_product_attribute_filter_collapse/__manifest__.py
pilarvargas-tecnativa: thanks :)

CarlosRoca13 (approved):

PR #826, opened 2023-08-10
[16.0][MIG] website_sale_order_type: Migration to version 16.0

pilarvargas-tecnativa:
cc @Tecnativa TT44380

@stefan-tecnativa @chienandalu please review :)


pedrobaeza:
This one shouldn't be needed if the field is computed writable (which I think it is), but if not, the patch should be to make it on the base module, like we have done with this one: https://github.com/OCA/sale-workflow/commit/788e2bc1d19261c0b7a2d15f10846007230afba2

chienandalu:
> This one shouldn't be needed if the field is computed writable (which I think it is), but if not, the patch should be to make it on the base module

Yes, this can be a good chance to fix that

pilarvargas-tecnativa:
Fix made in https://github.com/OCA/sale-workflow/pull/2639

pilarvargas-tecnativa:
This migration PR has been reopened as the fix made in sale_order_type did not cover the e-commerce flow. Therefore this module is still needed.

@pedrobaeza can you check it too?

I have decided to extend the methods that set the values that are overwritten in the e-commerce flow instead of the method that prepares the dictionary and calls these methods so that in case they are called from anywhere else the intended effect of this module is still in effect.

website_sale_order_type/models/website.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Website(models.Model):
+    _inherit = "website"
+    _description = "Allowing sale_order_type to work with website_sale."
+
+    def sale_get_payment_term(self, partner):
+        payment_term_id = super().sale_get_payment_term(partner)
+        if partner.sale_type.payment_term_id:
+            payment_term_id = partner.sale_type.payment_term_id.id
+        return payment_term_id
```
pedrobaeza: And the same for the other:

```suggestion
        if partner.sale_type.payment_term_id:
            return partner.sale_type.payment_term_id.id
        return super().sale_get_payment_term(partner)
```

pedrobaeza (commented):
OK for overriding these methods instead.

website_sale_order_type/models/website.py
pilarvargas-tecnativa: thanks!

pedrobaeza (approved):

website_sale_order_type/models/website.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models
+
+
+class Website(models.Model):
+    _inherit = "website"
+    _description = "Allowing sale_order_type to work with website_sale."
```
chienandalu: You shouldn't change the model `_description`

chienandalu (approved):
:+1: 

Just a minor comment:

website_sale_order_type/models/website.py
pedrobaeza: Good catch! This is inherited by the migration on v15, and provokes a bad problem, renaming the whole model, and sending the button in mails with such text.

stefan-tecnativa (approved):

pedrobaeza (changes requested):
`_description` should be removed.

website_sale_order_type/models/website.py
pilarvargas-tecnativa: Done, thanks!

chienandalu (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #827, opened 2023-08-17
[16.0][ADD] New module website sale product image sample

manuelcalerosolis:
This module allows you to select the product variants by displaying a thumbnail image.

Quick sample

image


website_sale_product_image_sample/readme/ROADMAP.rst
```diff
@@ -0,0 +1,5 @@
+* For the correct functioning of the module I had to rewrite .css_attribute_color, impacting the way of representing the selection by colors.
```
chienandalu: What about adding an extra class that overrides those styles only for the image thumbnails?

chienandalu (commented):
Nice addon @manuelcalerosolis Some comments

- I couldn't set the attribute image, is the view extension missing?

manuelcalerosolis:
The image is obtained from the variant (product.product), it is so by customer requirements.

￼

Buscar esto en Google

chienandalu:
> The image is obtained from the variant (product.product), it is so by customer requirements.

Ah, that makes sense :)

About the styles, did you explore adding an extra class so you don't need to change the base one?

manuelcalerosolis:
@chienandalu

I tried creating a new css .css_attribute_product_image but the thumbnail image doesn't stay selected.

The problem is this line 

`t-attf-class="css_attribute_product_image #{'active' if ptav in combination else ''}"
`

I think the "combination" variable doesn't work with another css selector

peek_no_select

chienandalu:
> @chienandalu
> 
> I tried creating a new css .css_attribute_product_image but the thumbnail image doesn't stay selected.
> 

Yes, that's because the js logic uses that class as a selector. I meant an additional class that would add up/override the other styles. Something like this:

```xml
t-attf-class="css_attribute_color css_attribute_product_image #{'active' if ptav in combination else ''} #{'custom_value' if ptav.is_custom else ''}"
```

manuelcalerosolis:
@chienandalu 

Thanks for your help, I have created a new .css and added it to the old one as you indicated, now it works correctly.

chienandalu (approved):
Great! Thanks :)

ioans73 (approved):
Functional review 👍🏻

HaraldPanten:
@pedrobaeza Can this one be merged? THX

HaraldPanten (approved):

PR #833, opened 2023-08-28
[16.0][FIX] website_sale_hide_price: show alert when price is hidden

yibudak:
When the price is hidden on the website, we have a field in both the `website` and `product.template` model about why it is hidden. But this alert is not visible in the frontend even if the price is hidden because it has the wrong condition.

To reproduce the error:
1- Hide prices to Public User.
2- Define a price hidden alert on the website in general settings.
3- Check the product price and the alert.

In this PR I believe I have solved the error, I am checking if the alert is defined and if the price is currently hidden.

### DIFF
--- a/website_sale_hide_price/views/website_sale_template.xml
+++ b/website_sale_hide_price/views/website_sale_template.xml
@@ -38,10 +38,10 @@
             >pricelist,product,website,website.website_show_price,product.website_hide_price</attribute>
         </xpath>
         <xpath expr="//a[@id='add_to_cart']" position="before">
-            <t t-set="user_authenticated" t-value="user_id != website.user_id" />
             <div
                 class="alert alert-info"
-                t-if="product.website_hide_price and (product.website_hide_price_message or website.website_hide_price_default_message) and user_authenticated"
+                t-if="(product.website_hide_price_message or website.website_hide_price_default_message)
+                 and not (website and website.website_show_price and not product.website_hide_price)"
             >
                 <i class="fa fa-info-circle" /> <span
                     t-esc="product.website_hide_price_message or website.website_hide_price_default_message"


website_sale_hide_price/views/website_sale_template.xml
```diff
@@ -38,10 +38,10 @@
             >pricelist,product,website,website.website_show_price,product.website_hide_price</attribute>
         </xpath>
         <xpath expr="//a[@id='add_to_cart']" position="before">
-            <t t-set="user_authenticated" t-value="user_id != website.user_id" />
```
yibudak, line 41: We don't need this variable because we check otherwise if prices are visible in other views. It should be the same here as it is in the other views.

yibudak:
@pedrobaeza @chienandalu @pilarvargas-tecnativa @manuelcalerosolis @ChrisOForgeFlow

pedrobaeza:
Squash commits into one and use a proper commit message:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

yibudak:
@pedrobaeza done, thanks.

chienandalu (commented):
Hi, @yibudak Thanks for your contribution :)

Some customers don't want that the public user can see the message. Anyway, you could add a website configuration to allow both behaviors.

yibudak:
> Hi, @yibudak Thanks for your contribution :)
> 
> Some customers don't want that the public user can see the message. Anyway, you could add a website configuration to allow both behaviors.

Thanks for your kind reply @chienandalu. If a client does not want to show a price hidden message to the public user, they can simply leave this field blank in the general settings. Also, this special case is not mentioned in the module description, in such cases, isn't it better to fork the module and customize it for your own use? Thanks again.

pedrobaeza:
If you leave it blank, it doesn't appear either to logged users (B2B use). What you propose is not a fix, but a change of behavior, so or you do such customization module with your new behavior, or the fix is to change the README, stating current behavior, or you can add the configuration option to have all the features in the same one.

yibudak:
Imagine a "normal" user who installs this module and doesn't know much about coding. Do you think this person would install the module based on how the code works or based on what is written in the README? If you believe that he will install the module by reading the code, then he can customize it as he wants for the special cases you mentioned. But since this would not be possible for a normal user, this is a fix, even if you claim otherwise.

chienandalu:
@yibudak We're not saying you're  proposal isn't reasonable. Just set it behind a website setting (two lines of code) and put it on the readme and we'll happily accept it.

yibudak:
@chienandalu The current situation meets my and my client's needs because we chose to use this module exactly according to what is written in the README and we fixed where it was wrong. If you think that the 2 lines of extra code you mentioned will answer the needs of the community, you can do feat with this PR. Thank you.

pedrobaeza:
Sorry, but this is not how things go in community. If you are not going to do the change, this will be hardly accepted to be merged, as it's modifying previous behavior that was intended by the original contributors (us). We don't have the need you and your customers have, so we are not going to spend time doing such change now that you refuse to do for getting this PR mergeable.

yibudak:
This is exactly how the community works. I fixed the mistake made by the almighty "original contributors" and made it work as described in the README. And let me tell you, we don't need your monarchy either. This attitude will only cost you.

pedrobaeza:
Ok, as you prefer.

chienandalu:
sorry @yibudak but you're not being reasonable. This isn't a fix as the behavior is intended. I took the time to check on the original customer requisites that our customers paid for and came back with the suggestion to make it optional. We could even discuss the defaults, but not it such rude terms of course.

PR #840, opened 2023-09-07
[16.0][FIX] website_sale_cart_expire: Protect _setExpirationDate

papulo79:
Protect _setExpirationDate assignment error,  added format check before setting the datetime in _str_to_datetime_ in order to avoid error message in website.

It checks "/shop/cart/get_expire_date" url to retrieve the information, but when it's installed some module that restricts the access to some url (like website_require_login) it retrieves HTML from the redirection.

Function evaluates "expireDate" as string and tries to convert it to datetime, but it's not a date so raises throws error.

Now checks if the string has a valid format date before try to convert it otherwise sets "expireDate" to undefined.

### DIFF
--- a/website_sale_cart_expire/readme/CONTRIBUTORS.rst
+++ b/website_sale_cart_expire/readme/CONTRIBUTORS.rst
@@ -1,3 +1,4 @@
 * `Camptocamp <https://www.camptocamp.com>`_
 
      * Iván Todorovich <ivan.todorovich@gmail.com>
+     * Pablo De Andrés <pablo.deandres@factorlibre.com>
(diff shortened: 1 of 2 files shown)


ivantodorovich:
IMO this is a misconfiguration / limitation of `website_require_login`. I'm actually not that familiar with that module to know.

The error you're experiencing is (sort of) expected. This module is expecting that endpoint to be accessible in order to work and to refresh the timer. We can't just silently ignore this IMO.

papulo79:
@ivantodorovich Yep, I got your point, thanks for sharing your idea. I won't continue for this path.

PR #844, opened 2023-09-13
[15.0][FIX]website_sale_invoice_address: One of the elements in the tour test cannot be found using the current repository

manuelregidor:
With the current modules in this repository, the <a> element containing "Large Meeting Table" cannot be found, as it is placed in page /shop/2 (and not in page /shop).

With this fix, a different product placed in page /shop is used in the test tour.

### DIFF
--- a/website_sale_invoice_address/static/src/js/website_sale_invoice_address_tour.js
+++ b/website_sale_invoice_address/static/src/js/website_sale_invoice_address_tour.js
@@ -24,7 +24,7 @@ odoo.define("website_sale_invoice_address.tour", function (require) {
     tour.register(
         "website_sale_invoice_address_tour",
         {
-            url: "/shop",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Isn't better to start with a search of such element (`/shop?search...`) for always being able to locate it? If not, and more demo data are added, we will be in the same problem later.

manuelregidor:
@HaraldPanten

manuelregidor:
> Isn't better to start with a search of such element (`/shop?search...`) for always being able to locate it? If not, and more demo data are added, we will be in the same problem later.

Done

HaraldPanten (approved):
LGTM. THX Manuel and Pedro for your effort.

chienandalu (approved):
Thanks :)

PR #845, opened 2023-09-13
[15.0][IMP] website_sale_secondary_unit: Hide/show uom name in secondary quantity description and delete column in cart summary

pilarvargas-tecnativa:
There were two quantity columns in the purchase summary, one with the secondary unit quantity and one with the product quantity. A quantity column in the purchase summary has been removed to give more clarity to the view. When product names or units of measure have a considerable number of characters, having two columns causes a crowding of the view and therefore a feeling of content clutter. Instead, their content has been moved so that both the secondary unit quantity and the quantity of product purchased appear in the same column.

Before:
image

After:
image

We have also added the option to hide the name of the secondary unit of measure in the secondary unit description template in case it matches the unit of measure of the product as depending on where it is being called from it can be redundant.

Before:
image

After:
image

cc @Tecnativa TT45032

@chienandalu @pedrobaeza please review


pedrobaeza (approved):

chienandalu (approved):
Cool :sunglasses: :+1:

chienandalu:
@pilarvargas-tecnativa rebase to get the fixed tests

pedrobaeza:
CI fixed in other PR:

/ocabot merge minor

pedrobaeza:
Please rebase to see the problem.

pilarvargas-tecnativa:
> Please rebase to see the problem.

I've pushed again because I had removed a "/" separator. It was an error in the tests, when changing the html they failed but my database was a bit dirty and I hadn't detected it locally either, now it's fixed :) Thanks!

PR #849, opened 2023-10-02
[15.0][FIX] website_sale_checkout_skip_payment: order confirmation

chienandalu:
The module wasn't working as expected and the orders weren't confirmed and the emails weren't sent notifying the salesmen.

Also we provide some compatibility with module like sale_financial_risk which will avoid confirming order when the partner risk is exceeded so we can just let them as sent.

cc @Tecnativa TT45113

please review @CarlosRoca13 @pilarvargas-tecnativa

### DIFF
--- a/website_sale_checkout_skip_payment/__manifest__.py
+++ b/website_sale_checkout_skip_payment/__manifest__.py
@@ -20,8 +20,8 @@
         "views/res_config_settings_views.xml",
     ],
     "assets": {
-        "web.assets_frontend": [
-            "website_sale_checkout_skip_payment/static/src/js/*.js",
+        "web.assets_tests": [
+            "website_sale_checkout_skip_payment/static/src/tests/tours/**/*.js",
         ],
     },
 }
--- a/website_sale_checkout_skip_payment/controllers/main.py
+++ b/website_sale_checkout_skip_payment/controllers/main.py
@@ -1,19 +1,16 @@
 # Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
 # Copyright 2017 David Vidal <david.vidal@tecnativa.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo.http import request, route
 
-from odoo import http
-from odoo.http import request
-
-from odoo.addons.payment.controllers.portal import PaymentPortal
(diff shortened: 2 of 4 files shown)


pilarvargas-tecnativa:
https://github.com/OCA/e-commerce/blob/94e03b0dc542c7f28bfe87c23f3a6fecaaf637b6/website_sale_checkout_skip_payment/static/src/js/website_sale_checkout_skip_payment_tour.js#L21

The tests in the test fail because the portal user goes to the address after this step, you would have to add the step:
```
{
      trigger: ".btn-primary:contains('Next')",
},
```

chienandalu:
> The tests in the test fail because the portal user goes to the address after this step, you would have to add the step:
> 
> ```
> {
>       trigger: ".btn-primary:contains('Next')",
> },
> ```

That was a symptom not the cause :wink: The issue here was the compatibility with website_sale_vat_required

chienandalu:
It's green now

CarlosRoca13 (approved):
Thanks :)

pilarvargas-tecnativa (approved):
thanks! :)

website_sale_checkout_skip_payment/__manifest__.py
```diff
@@ -20,7 +20,7 @@
         "views/res_config_settings_views.xml",
     ],
     "assets": {
-        "web.assets_frontend": [
+        "web.assets_tests": [
             "website_sale_checkout_skip_payment/static/src/js/*.js",
```
pedrobaeza: Shouldn't this be reorganized for not importing as glob the general JS folder (and avoiding side effects if something more is added there)? Maybe `website_sale_checkout_skip_payment/static/src/js/tours/*.js`

chienandalu: Done in src/tests/tours in Odoo's style

chienandalu:
> @pedrobaeza your merge command was aborted due to failed check(s), which you can inspect on [this commit of 15.0-ocabot-merge-pr-849-by-pedrobaeza-bump-patch](https://github.com/OCA/e-commerce/commits/f82507c23265836f89e55f07144e30bc5135e6cf).
> 
> After fixing the problem, you can re-issue a merge command. Please refrain from merging manually as it will most probably make the target branch red.

tour glitch

PR #850, opened 2023-10-03
[16.0][MIG] website_sale_require_legal: Migration to version 16.0

pilarvargas-tecnativa:
supersedes https://github.com/OCA/e-commerce/pull/809

First of all, this module is necessary as the acceptance of legal terms is done at the address level with a metadata record of the acceptance. Currently odoo only has the acceptance of terms and conditions at checkout level.

ping @chienandalu @CarlosRoca13 

cc @Tecnativa TT45248


pedrobaeza:
You can clarify the "disclaimer" in the README, using the new fragment `CONTEXT.rst`.

pilarvargas-tecnativa:
> You can clarify the "disclaimer" in the README, using the new fragment `CONTEXT.rst`.

done! thanks!

website_sale_require_legal/controllers/main.py
```diff
@@ -27,16 +27,15 @@ def _mandatory_legal_terms(self):
...
+        request.update_context(
             needs_legal=request.website.viewref(
                 "website_sale_require_legal.address_require_legal"
-            ).active,
+            ).active
         )
         result = super().checkout_form_validate(mode, all_form_values, data)
         # Unpatch context
-        request.context = old_context
+        request.update_context = old_context
```
chienandalu: This is wrong. Update context is a method not a dict.

```suggestion
        request.update_env(context=old_context)
```

pilarvargas-tecnativa: Thanks, I was confused when I looked at another code, the thing is that while I was debugging it did update the context with the old one in that way. Anyway this would be the correct way.

website_sale_require_legal/tests/test_ui.py
```diff
@@ -28,24 +34,18 @@ def setUp(self):
...
                 ("website_id", "!=", "False"),
             ]
         )
         partner = order.partner_id
-        # Assert that the sale order and partner have metadata logs
-        self.assertTrue(
-            order.message_ids.filtered(
-                lambda one: "Website legal terms acceptance metadata" in one.body
-            )
-        )
```
chienandalu: Why? One of the features is to attach metadata to the accepted terms of the order. And it works (I tested it setting the feature on). So you're missing something in your test

website_sale_require_legal/readme/CONTEXT.rst
```diff
@@ -0,0 +1,6 @@
+This module adds the functionality to require you to tick the checkbox to accept the
+legal terms in the address form. In addition, it provides a metadata record when the
+legal terms are accepted via the address form.
```
chienandalu: And in the sale.order when the legal terms are accepted there.

Indeed this module can be useful without the address acceptance active

website_sale_require_legal/tests/test_ui.py
pilarvargas-tecnativa: The test is complete again, the problem is that if the payment_custom module is not installed the test will run but the module behaviour means that as long as the sales order is not confirmed there will be no metadata in the order.

chienandalu (approved):
Thanks :+1:

website_sale_require_legal/readme/CONTEXT.rst
pilarvargas-tecnativa: I have completed the description of the context, thanks!

pedrobaeza:
Tests are not being executed as `payment_custom` is not installed. Is there any way to force this? Why do you need such module? Maybe you can create on the fly a payment provider.

pilarvargas-tecnativa:
> Tests are not being executed as `payment_custom` is not installed. Is there any way to force this? Why do you need such module? Maybe you can create on the fly a payment provider.

This is the way the tests are run in website_sale. You need to do the step to confirm the order to get the metadata in the sale order which is a part of the functionality of this module.  On the other hand I think if I define a new payment provider I think it is also necessary the payment_custom module to manage it from the e-commerce.
Another option would be to split the tests and perform the part where the payment is not necessary on one side and the metadata check when confirming the order on the other side and that way most of the test will be executed.

pedrobaeza:
I don't think you need to complete the checkout to test the features of this module, and actually, the last step of the tour is not checking anything, so I think you should remove the requirement of `payment_custom` module for the tests, and just remove the last step.

chienandalu:
> I don't think you need to complete the checkout to test the features of this module, and actually, the last step of the tour is not checking anything, so I think you should remove the requirement of `payment_custom` module for the tests, and just remove the last step.

We should be able to check if the metadata is saved with the core legal checkbox

pedrobaeza:
But they are saved on the partner, isn't it? And anyway, is the test really checking that?

chienandalu:
> But they are saved on the partner, isn't it? And anyway, is the test really checking that?

It's posted in the partner when the terms acceptance is checked in the address form and it's posted in the sale.order when the acceptance is made in the checkout page.

Both conditions are checked in the tests

pedrobaeza:
Isn't there any way to complete that without the module? If not, we are not going to test it never here in OCA.

pilarvargas-tecnativa:
> Isn't there any way to complete that without the module? If not, we are not going to test it never here in OCA.

If there is no payment, there is no acceptance of terms for the sales order and that part could not be tested, another thing is to separate that part as I said before, make a test for the part that does not need the payment step and another test with the part that does need it and so in OCA most of the functionality would be tested.

chienandalu:
I've been doing some research. The real problem is that although we could create a dummy payment method (example here: https://github.com/OCA/OCB/blob/16.0/addons/website_sale/tests/test_website_sale_add_to_cart_snippet.py#L14-L31) it doesn't seem trivial to finish the cart without a full fledged payment method.

Another option could be to have `payment_demo` as a base module for our CIs and rely our test on that one

pilarvargas-tecnativa:
> I've been doing some research. The real problem is that although we could create a dummy payment method (example here: https://github.com/OCA/OCB/blob/16.0/addons/website_sale/tests/test_website_sale_add_to_cart_snippet.py#L14-L31) it doesn't seem trivial to finish the cart without a full fledged payment method.
> 
> Another option could be to have `payment_demo` as a base module for our CIs and rely our test on that one

By testing locally the test provider makes the test cycle complete in its entirety, let's see what happens now once the checks are done in the PR.

pedrobaeza:
Can you please rebase after https://github.com/OCA/e-commerce/pull/861 is merged

pilarvargas-tecnativa:
@pedrobaeza is green! :)

pedrobaeza (approved):
/ocabot merge nobump

PR #854, opened 2023-10-10
[16.0][MIG] website sale product assortment

suker:
Migration to 16.0


suker:
Hi! can you take a look @CarlosRoca13?

Thanks in advance.

suker:
@OCA/website-maintainers anyone can take a look?

Thanks in advance.

danielduqma:
Hi @suker,

Even though you mentioned module's mantainer and OCA team related to this repo, this PR has been ignored twice and superseeded by https://github.com/OCA/e-commerce/pull/873. Sadly, your work has been in vain without even looked by maintainers. Please, close it as it no longer applies.

Thanks and regards.

pedrobaeza:
@xAdrianCif @suker you should put correctly the module name in the PR title for being searchable. There's no bad intentions here, just misunderstanding. The pings are lost into the hundreds of notifications from GitHub. We always look for existing PRs searching the module name. Our team is aware of this, and now they will search by individual words instead the whole module name, but you should also improve the PR title according guidelines.

We don't look on the migration issue, as this has proven other times to be links to non valid pull requests that were closed, and thus, it's very difficult to track the correct one, and that's why we use this technique.

danielduqma:
Thanks for the reply @pedrobaeza, reviewing this problem we noticed that PR title was wrong by our mistake, but we found strange that none of 8 mantainers mentioned here adviced this and warn it to be corrected. Anyway, you are totally right about the title because it is clearly mentioned in [OCA guidelines about migration](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0#technical-method-to-migrate-a-module-from-150-to-160-branch).

We take note of our problems to avoid this kind of problems in the future.

Regards.

pedrobaeza:
Take into account that there's a massification of notifications. For example, I receive about 3500 mails each day, so it's very difficult to link one thing to another, because if I receive a ping saying that this module migration is in PR, but I don't need it just right now, I'm not going to review it until I needed it, and when it's needed, then remembering that there was an opened PR for this is difficult. Even those, we already know a lot of done PRs that ping us that are incorrect (tests, commit history, etc), and they end up being closed by other maintainer or the contributor itself. The solution we looked was to search for opened PRs. After this, we include in our process to look for individual words also for avoiding this problem in the future.

PR #859, opened 2023-10-18
[15.0][IMP] e-commerce  : test performance improvement

josep-tecnativa:
CC @Tecnativa TT45458

### DIFF
--- a/website_sale_b2x_alt_price/tests/test_ui.py
+++ b/website_sale_b2x_alt_price/tests/test_ui.py
@@ -5,12 +5,23 @@ from odoo.tests.common import Form, HttpCase
 
 
 class UICase(HttpCase):
-    def setUp(self):
-        super().setUp()
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # Remove this variable in v16 and put instead:
+        # from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
+        DISABLED_MAIL_CONTEXT = {
+            "tracking_disable": True,
+            "mail_create_nolog": True,
+            "mail_create_nosubscribe": True,
+            "mail_notrack": True,
+            "no_reset_password": True,
+        }
+        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
         # Create and select a pricelist
         # to make tests pass no matter what l10n package is enabled
-        website = self.env["website"].get_current_website()
-        pricelist = self.env["product.pricelist"].create(
+        website = cls.env["website"].get_current_website()
+        pricelist = cls.env["product.pricelist"].create(
             {
                 "name": "website_sale_b2x_alt_price public",
                 "currency_id": website.user_id.company_id.currency_id.id,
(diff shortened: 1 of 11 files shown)


website_sale_b2x_alt_price/tests/test_ui.py
```diff
@@ -5,120 +5,131 @@
...
+        cls.training_accounting = cls.env["product.template"].create(
             {
                 "name": "Training on accounting - website_sale_b2x_alt_price",
                 "list_price": 100,
                 "type": "service",
                 "website_published": True,
-                "uom_id": self.ref("uom.product_uom_unit"),
-                "uom_po_id": self.ref("uom.product_uom_unit"),
+                "uom_id": cls.ref("uom.product_uom_unit"),
+                "uom_po_id": cls.ref("uom.product_uom_unit"),
```
chienandalu: (some others to check)
```suggestion
                "uom_id": cls.env.ref("uom.product_uom_unit"),
                "uom_po_id": cls.env.ref("uom.product_uom_unit"),
```

website_sale_b2x_alt_price/tests/test_ui.py
```diff
@@ -5,120 +5,132 @@
...
+        # redoo CI
+        cls.training_accounting = cls.env["product.template"].create(
             {
                 "name": "Training on accounting - website_sale_b2x_alt_price",
                 "list_price": 100,
                 "type": "service",
                 "website_published": True,
-                "uom_id": self.ref("uom.product_uom_unit"),
-                "uom_po_id": self.ref("uom.product_uom_unit"),
+                "uom_id": cls.env.ref("uom.product_uom_unit"),
```
pedrobaeza: And the same for other dictionary values for record creation (in other places it may be OK to put the recordset, not the ID):

```suggestion
                "uom_id": cls.env.ref("uom.product_uom_unit").id,
```

chienandalu (approved):
I'd say the error is a tour glitch

pedrobaeza:
Please solve conflicts.

josep-tecnativa:
> Please solve conflicts.

Done

website_sale_product_detail_attribute_image/tests/test_website_sale_product_detail_attribute_image.py
```diff
@@ -4,40 +4,49 @@
...
+        super().setUpClass()
+        cls.env = cls.env(
+            context=dict(
+                cls.env.context,
+                mail_create_nolog=True,
+                mail_create_nosubscribe=True,
+                mail_notrack=True,
+                no_reset_password=True,
+                tracking_disable=True,
+            )
```
yajo: suggestion: create a `DISABLED_MAIL_CONTEXT` variable and use that here:

```suggestion
            context=dict(
                cls.env.context,
                **DISABLED_MAIL_CONTEXT,
            )
```

This way, when fwporting to v16, you can just [import that variable](https://github.com/odoo/odoo/blob/d5c3d47f80d88301e2311c9c1d6eabedca8a562d/odoo/addons/base/tests/common.py#L10-L16) and the diff will be smaller.

The same applies elsewhere.

yajo (approved):

website_sale_product_detail_attribute_image/tests/test_website_sale_product_detail_attribute_image.py
pedrobaeza: Oh, TIL. @josep-tecnativa please change from now on to do such change. More compact and elegant, and also Odoo can expand this context if needed and we won't need any change in our part.

pedrobaeza:
@josep-tecnativa pending refactoring for being nice with the fw-port.

PR #871, opened 2023-11-08
[16.0][FIX] website_sale_hide_price: Error loading view after update

papulo79:
…configuration update.

Steps to reproduce:

- Load a runboat instance from the module.
- Go to "Settings > Website > Shop - Checkout process"
- Enable "Buy now" option
- Save settings.
- Error message appears:
imagen

Error explanation:

The xpath points to a node that has "t-attf-class" attribute instead "class"


papulo79:
@pedrobaeza do I need to fix anything otherwise the xpath? I don't know why are failing the tests.

Regards.

pedrobaeza:
Your change is not compatible with current existing modules

papulo79:
How can I check how my change is failing in order to fix the error? now is not possible to enable "Buy now" option. 

Right? Or Am I missing something?

pedrobaeza:
Check the logs on the CI and dig to see which is the problem.

papulo79:
I've found the error and fixed, I copied the correction line in two places by mistake.

What about the codecov/project? It's not related with my fix, right?

Regards.

chienandalu (approved):
LGTM Thanks! :+1:

chienandalu:
> What about the codecov/project? It's not related with my fix, right?

Don't worry about it nobody understands it :sweat_smile:

papulo79:
Thank you!! Let's give the fix some love.

danielduqma (approved):

pedrobaeza:
OK, let's merge it:

/ocabot merge minor

Now, for thanking the fast reviewing here, you can review other PRs in this repository for getting them merged as well.

pleirb:
Hello, how are you all?
Something similar is happening to me, could it have anything to do with it?

imagen

chienandalu:
Maybe something changed upstream. Research where that selector should be and if it's there. Odoo tries to not change xml views, but sometimes they just do and that breaks things...

In the other hand, normally this kind of issues have the origin in tweaking with the custom views inheritance or upgrading from older versions without dumping the old view structures...

pleirb:
> Tal vez algo cambió río arriba. Investiga dónde debe estar ese seleccionador y si está allí. Odoo trata de no cambiar las vistas de xml, pero a veces sólo lo hacen y eso rompe las cosas...
> 
> Por otro lado, normalmente este tipo de problemas tienen el origen en retocar con las vistas personalizadas herencia o actualización de versiones antiguas sin ver las estructuras de vista antiguas...



Thank you very much @chienandalu what you say is very likely, I come from a migration. I'll try to identify the problem. Do you know if there is any way to list the views that have a change from the default one in the module.

PR #885, opened 2024-01-15
[IMP] website_account_fiscal_position_partner_type: Show B2B/B2C information in sign up page

manuelregidor:
This improvement applies 2 changes:

1. It changes the location of the Partner Type field in the signup view. It is now underneath the email field, so it becomes the second field in the form.
website_account_fiscal_position_partner_type_1_

2. It adds the possibility of showing a message underneath the Partner Type field in the signup view. This message can be customised from the website configuration view and it can be different depenending on the chosen option. The message has to be set in HTML.

website_account_fiscal_position_partner_type_2
website_account_fiscal_position_partner_type_3
website_account_fiscal_position_partner_type_4

The reason why we propose this improvement is because some of our customers have noticed that sometimes, website users do not fully know the difference between being a B2B or a B2C customer, so adding information to make it clear could be helpful for them. Also, as the name they introduce when registering might depend on the B2B/B2C choice, placing this before the name field could also prevent them from making mistakes.


manuelregidor:
@HaraldPanten @ValentinVinagre

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):
LGTM

HaraldPanten:
@pedrobaeza @etobella do you mind merging these changes?

THX!

PR #896, opened 2024-02-08
[ADD] website_sale_product_name_fixed_height: add module

dessanhemrayev:
This module lets the user set for each website how many lines of description to display in /shop page, and if the name of the product does not fit the lines set, then the name is shortened with suspension dots


dessanhemrayev:
@francesco-ooops please check)

website_sale_fixed_name_height/models/product_template.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (C) 2024 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class ProductTemplate(models.Model):
+    _inherit = "product.template"
+
+    def get_row_count(self):
```
geomer198: Could you please add doc string for this method?

geomer198 (approved):
LGTM

website_sale_fixed_name_height/models/product_template.py
dessanhemrayev: Thanks for review, I'll add it)

website_sale_fixed_name_height/views/website_sale_template.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="products_item" inherit_id="website_sale.products_item" priority="99">
+        <xpath expr="//*[hasclass('o_wsale_products_item_title')]" position="replace">
```
HekkiMelody: **issue**: it's better to avoid using `replace` whenever possible. If I'm not mistaken, this can be done with a simple `position="attributes".

website_sale_fixed_name_height/views/website_sale_template.xml
```diff
@@ -0,0 +1,25 @@
...
+            <h6
+                class="o_wsale_products_item_title"
+                t-attf-style="display: -webkit-box;-webkit-box-orient: vertical;-webkit-line-clamp: #{product.get_row_count()};overflow: hidden;"
```
HekkiMelody: **praise**: Nice solution, I didn't know about this one.

**thought** According to MDN, compatibility is limited to only Firefox.

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

However, I tested it in Chromium (to be specific, I tested the one in this article https://css-tricks.com/line-clampin/ ) and it seems to work, and the article itself also claims it works on chromium-based browsers.

HekkiMelody (changes requested):
There is a discrepancy between documentation and code

Documentation claims you can set a different number of lines per website, but it really is per company and not per website. Maybe the setting could be indeed moved to the website model?

website_sale_fixed_name_height/views/website_sale_template.xml
dessanhemrayev: Thanks, I'll use it

francesco-ooops (approved):
Functional ok!

website_sale_fixed_name_height/views/website_sale_template.xml
```diff
@@ -0,0 +1,15 @@
...
+<odoo>
+    <template id="products_item" inherit_id="website_sale.products_item" priority="99">
+        <xpath
+            expr="//*[hasclass('o_wsale_products_item_title')]"
+            position="attributes"
+        >
+            <attribute
+                name="t-attf-style"
+                separator=" "
+                add="display: -webkit-box;-webkit-box-orient: vertical;-webkit-line-clamp: #{website.row_count};overflow: hidden; height: #{19*website.row_count}px;"
```
HekkiMelody: **thought**: specifying a height in pixels could create issues for accessibility and may not be compatible with user choices when it comes to font size.

IMO it'd be better to use a relative unit, like `em` or even better here: `lh` https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#line_height_units . 

```
height: #{website.row_count}lh;
```

It probably  won't be perfect in all situations but should still work better most of the times. What do you think?

dessanhemrayev: Thanks, I'll fix it)

HekkiMelody (approved):
Code review

francesco-ooops:
@pedrobaeza can this be merged? thank you!

pedrobaeza:
The module name should be `website_sale_product_name_fixed_height`

PR #899, opened 2024-02-20
[FIX] website_sale_suggest_create_account: changing the priority of a view

lef-adhoc:
I change the priority of the cart and short_cart_summary views because the website_sale_checkout_skip_payment module uses the same template and this produces an error.
[Template](https://github.com/OCA/e-commerce/blob/b9e397a6c79a4bdef07d1141ae549a36e37ff7fd/website_sale_checkout_skip_payment/views/website_sale_template.xml#L85C1-L103C16)

This video reproduces the error
[Video](https://drive.google.com/file/d/1k7MdI1FYSzFcBD7oF7MBzX-v2ZOK-kY7/view)

image

### DIFF
--- a/website_sale_suggest_create_account/views/website_sale.xml
+++ b/website_sale_suggest_create_account/views/website_sale.xml
@@ -1,6 +1,11 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <template id="cart" inherit_id="website_sale.cart" customize_show="True">
+    <template
+        id="cart"
+        inherit_id="website_sale.cart"
+        customize_show="True"
+        priority="18"
+    >
         <xpath expr="//div[@id='wrap']" position="before">
             <t t-set="user_authenticated" t-value="user_id != website.user_id" />
             <t
@@ -61,6 +66,7 @@
         id="short_cart_summary"
         inherit_id="website_sale.short_cart_summary"
         customize_show="True"
+        priority="18"
     >
         <xpath expr="//a[@t-attf-href='{{redirect_url}}']" position="attributes">
             <attribute name="class" />


matiasperalta1 (approved):

matiasperalta1:
@pedrobaeza Hi! everything seems good. Please, can you check this PR and merge if it's ok?

pedrobaeza:
But the problem comes when there are `replace`. Is that the case with any of the modules?

lef-adhoc:
@pedrobaeza gives error because when you have installed the module "website_sale_suggest_create_account" you are positioning yourself in this [part](https://github.com/odoo/odoo/blob/16.0/addons/website_sale/views/templates.xml#L1543) of the view with this [line](https://github.com/OCA/e-commerce/blob/b9e397a6c79a4bdef07d1141ae549a36e37ff7fd/website_sale_suggest_create_account/views/website_sale.xml#L65) (which eliminates what the class has), then when you install the "website_sale_checkout_skip_payment" module it gives error because with the xpath it looks for the tag a with the class that previously deleted the "website_sale_suggest_create_account" module [here](https://github.com/OCA/e-commerce/blob/16.0/website_sale_checkout_skip_payment/views/website_sale_template.xml#L86), that's why one solution is to give more priority to the "website_sale_checkout_skip_payment" view so that the class has not been deleted at the time of positioning in the same.

pedrobaeza:
OK, then it' some kind of "replace", but just for the attributes. As there's no easy way with this kind of changes, let's continue this way

Thanks for exposing me the problems.

Just one change to merge this: please put that explanation inside the commit message, so that we see it next time we wonder why the change.

lef-adhoc:
@pedrobaeza I wanted to inform you that I have added the message within the commit as you suggested. Could you please review it and confirm if it looks good? Thank you.

pedrobaeza (approved):
/ocabot merge minor

PR #901, opened 2024-02-27
[17.0] [MIG] website_sale_product_description

mmrondon:


Abranes:
Hi @mmrondon!

In Odoo 17.0, there is already a [field for eCommerce description](https://github.com/odoo/odoo/blob/a4e0e1bc7f53bc508570fb4b9d7957749f0ef55c/addons/website_sale/models/product_template.py#L51-L57), but this field is not displayed in any view.

In my opinion, the best approach is to use this field, create a view in products, and implement a migration script to migrate the values from the field in 16.0 to the new field in 17.0.

If you want, I can offer to do it myself.

Thanks!

chienandalu (commented):
I agree with @Abranes . Take in consideration that the backend view of `
description_ecommerce` was removed due to some kind of js/css incompatibilty as it's meant to be edited in the frontend: https://github.com/odoo/odoo/pull/138216

PR #902, opened 2024-02-29
[16.0][FIX] website_sale_stock_available: depends_context

danielduqma:
It's necessary to add key `website_sale_stock_available` to `depends_context` of compute method of `free_qty`. This could lead to get invalid cached content instead of computing with updated key.

FL-556-2511


pedrobaeza (approved):
Let's avoid this to be closed:

/ocabot merge patch

PR #903, opened 2024-02-29
[14.0][IMP] ajaxify cart _onClickAdd return resolved promise

FrancescoBallerini:
I wasn't sure about proposing this PR because this is not reproducible in runbot but it should be an improvement and in my case a valuable bug fixing, but I couldn't figure out the whole process behind this so I'd really like to see what you think.

Basically when I install [website_sale_ajaxify_cart](https://github.com/OCA/e-commerce/tree/14.0/website_sale_ajaxify_cart) and activate the feature, I click on "Add to cart" and the `_onClickAdd` event is correctly executed once, but after the first execution two classes will be added to the button:
- `o_website_btn_loading` 
-  `disabled` 

causing the button to stay forever in a disabled state like in screenshot (until I refresh the page )
image

These classes are really only added/remove by the execution of [makeButtonHandler method](https://github.com/odoo/odoo/blob/6c7f4559386776aaa2cfa8f3df5cd6d2b6c5ed95/addons/web/static/src/js/core/dom.js#L341) which is [used for async event handlers](https://github.com/odoo/odoo/blob/6c7f4559386776aaa2cfa8f3df5cd6d2b6c5ed95/addons/web/static/src/js/public/public_widget.js#L199-L201)  (`_onClickAdd` is async)

The issue here might be that if you don't resolve the pending promise you don't notify that the handler has done it's job, [as reported here](https://github.com/odoo/odoo/blob/6c7f4559386776aaa2cfa8f3df5cd6d2b6c5ed95/addons/web/static/src/js/core/dom.js#L290-L292) (makeAsyncHandler is the function calling makeButtonHandler).
This is not an issue if you switch to cart page like in the original Odoo workflow, but it could cause issues with "ajaxify workflow" I guess.

My main doubt about this analysis is why the non-resolved Promise is not causing the same issue on runbot.

The refactoring I have introduced in this commit has solved the issue for me, I don't think that resolving Promise has any drawback in this situation, what do you think?

Does anyone ever had similar issues ?

### DIFF
--- a/website_sale_ajaxify_cart/static/src/js/website_sale.js
+++ b/website_sale_ajaxify_cart/static/src/js/website_sale.js
@@ -10,12 +10,18 @@ odoo.define("website_sale_ajaxify_cart.website_sale", function (require) {
         /**
          * @private
          * @param {MouseEvent} ev
+         * @returns {Promise} resolved
          */
         _onClickAdd: function (ev) {
             this.isDynamic = Boolean($(ev.currentTarget).data("is-dynamic"));
             this.pageType = $(ev.currentTarget).data("page-type");
             this.targetEl = $(ev.currentTarget);
-            return this._super.apply(this, arguments);
+            var _super = this._super;
+            var args = arguments;
+            var self = this;
+            return Promise.resolve().then(function () {
+                _super.apply(self, args);
+            });
         },
         /**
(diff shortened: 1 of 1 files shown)


FrancescoBallerini:
Made some mess with local branches, sorry. Opened a new one

Issue #906, opened 2024-03-06
[12.0] website_sale_secondary_unit: product quantity isn't calculated with float factor


odooNextev:
When adding to the cart a product on an ecommerce using a secondary unit, it doesn't works properly with float factor
It always returns an incorrect quantity, which is different from the one selected, and it doesn't apply the conversion with the secondary unit
I'll show it on this short video: https://recordit.co/9ipAloRGGs

Affected versions:

- [ ] 12.0: https://github.com/OCA/e-commerce/issues/906

It works properly on version 13, try to backporting it still doesn't work as expected

PR #908, opened 2024-03-07
[12.0] [FIX] website_secondary_unit: working with float factor

odooNextev:
#906


odooNextev:
@OCA/e-commerce-maintainers can you review it?
Thanks in advance

PR #911, opened 2024-03-25
Improve '/page/product_brands' search domain

FrancescoBallerini:
Before commit:

All published brands are loaded in '/page/product_brands' route

After commit:

Publish brands are loaded in '/page/product_brands' only if they have at least one published product (we apply the same behaviour applied in '/shop')

Performance time increase a bit, search time goes up from 0,001s to 0,01s. It shouldn't be an issue since this search() is only used on '/page/product_brands' route (doesn't affect shop or other routes)

Perf sample (5 exec before/after) - tested on a DB with 16'000 products and 370 brands 

```
before:

Time taken: 0.00055665199943177867680788040161132812500000000000 seconds
Time taken: 0.00081052499990619253367185592651367187500000000000 seconds
Time taken: 0.00106807800057140411809086799621582031250000000000 seconds
Time taken: 0.00066884600073535693809390068054199218750000000000 seconds
Time taken: 0.00151757000003271969035267829895019531250000000000 seconds

after adding ("product_ids.is_published", "=", True) to domain

Time taken: 0.01136246300029597477987408638000488281250000000000 seconds
Time taken: 0.00896999799988407175987958908081054687500000000000 seconds
Time taken: 0.01567319999958272092044353485107421875000000000000 seconds
Time taken: 0.01054079599998658522963523864746093750000000000000 seconds
Time taken: 0.01098039499993319623172283172607421875000000000000 seconds
```

### DIFF
--- a/website_sale_product_brand/controllers/main.py
+++ b/website_sale_product_brand/controllers/main.py
@@ -47,7 +47,10 @@ class WebsiteSale(WebsiteSale):
     @http.route(["/page/product_brands"], type="http", auth="public", website=True)
     def product_brands(self, **post):
         b_obj = request.env["product.brand"]
-        domain = [("website_published", "=", True)]
+        domain = [
+            ("website_published", "=", True),
+            ("product_ids.is_published", "=", True),
+        ]
         if post.get("search"):
             domain += [("name", "ilike", post.get("search"))]
         brand_rec = b_obj.sudo().search(domain)


chienandalu (commented):
What about making the brand `website_published` computed stored depending on the products `is_pusblished` state?

FrancescoBallerini:
> What about making the brand `website_published` computed stored depending on the products `is_pusblished` state?

Thanks for the review! I would keep the brand `website_published` independent from the products `is_published` state: a user might need to temporarily unpublish a brand which has published products. Do you agree?

Thanks

chienandalu:
This would not be a problem. A computed stored field behaves very much like an onchange. However, it's true that you should decide whether or not to publish your brand depending on the state of the product...

I also wonder if with two separate queries we could gain some performance:

- First one: a read_group for published products grouped by brand.
- Second one: brands published and filtered by those ids gathered in the first query

FrancescoBallerini:
> I also wonder if with two separate queries we could gain some performance:
> 
> * First one: a read_group for published products grouped by brand.
> * Second one: brands published and filtered by those ids gathered in the first query

Hi @chienandalu I have made some tests, I had never used `read_group` method before so I am not sure if I executed it correctly, however from my tests it seems that the ORM mapping is winning over dobule query with groubpy.

I have prepared a patched version of the module where the controller main.py file is patched so it execute both queries and log execution time, so anyone could make additional tests if interested

[website_sale_product_brand_test_performance.zip](https://github.com/OCA/e-commerce/files/14781069/website_sale_product_brand_test_performance.zip)

From my tests I got similar pattern for every execution:

```
ORM search domain
Time taken: 0.00825832999998965533450245857238769531250000000000 seconds

ORM Groupby
Time taken: 0.01319241300006979145109653472900390625000000000000 seconds
```

FrancescoBallerini:
@chienandalu I realized that the "time taken" I reported in the last post for the single query (after adding `("is_published", "=", True)` in the domain) was a bit lower than the average I reported in the first post.. I hope that it wasn't misleading, I tested from the same machine but it was probably doing less work in background while testing :-) 

Basically the "time diff" average was about 0,005 seconds between two queries, which is not much but I would keep the search() at this point as it probably improves a bit the code readability as well. 

But let me know if you agree, also if you prefer I could repeat the test and post more lines

By the way, this is the code for double query:
```
brands_with_at_least_one_published_product = [
      item["product_brand_id"][0] for item in ProductTemplate.read_group(
      domain=[("is_published", "=", True), ("product_brand_id", "!=", False)],
      fields=["product_brand_id"],
      groupby=["product_brand_id"],
  )]

brand_domain = [("is_published", "=", True), ("id", "in", brands_with_at_least_one_published_product)]
brand_rec = b_obj.sudo().search(brand_domain)
```

it should be correct but let me know if I can improve it, thanks

PR #919, opened 2024-05-02
[16.0][IMP]add sequence on multi link

syera94:
@bguillot @legalsylvain 
Same change V14 to V16 to add a sequence in product template link


bguillot (approved):
Looks good, thanks !

PR #921, opened 2024-05-05
[17.0][FIX] website_sale_hide_price: qweb cache

stanislaw-ladorucki:
Fix caching of the website_sale.products view override.

Before this commit, this module added a view that added two price website related cache keys to the products grid wrapper. In Odoo 17.0, caching was removed for this element. In combination with this module, this resulted in re-rendering this portion of the view, only depending on the cache keys added by this module. Because of it, searched products didn't update when changing any of the search parameters.

This commit fixes it, by removing that override, and leaving the elements cache disabled.

### DIFF
--- a/website_sale_hide_price/views/website_sale_template.xml
+++ b/website_sale_hide_price/views/website_sale_template.xml
@@ -1,20 +1,5 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <template id="products" inherit_id="website_sale.products">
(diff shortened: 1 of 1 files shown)


theangryangel (approved):
Functionally seems correct, required since https://github.com/odoo/odoo/commit/540872d86b6f5e6d9ea935ddf66691d50b7e9778

pedrobaeza (approved):
/ocabot merge minor

chienandalu (approved):
LGTM

PR #927, opened 2024-05-24
[16.0][FIX] website_sale_stock_list_preview: Error trying to formatQuantity

CarlosRoca13:
Since the changes introduced in https://github.com/odoo/odoo/commit/a8c73ebc3deb5b74789461c05e98a2daba7f2526 we need to pass to the template of the remaining units to the function formatQuantity. So we need to pass the function.

The steps to reproduce the error:
1. Edit some product, example Customizable Desk
2. Select the option Show Available Qty and set it to 999999
3. Go to /shop

You will see the error

cc @Tecnativa TT49272

ping @pedrobaeza @victoralmau


victoralmau (changes requested):
Functional review OK, thanks!

pedrobaeza (approved):
/ocabot merge patch

PR #932, opened 2024-06-06
[16.0][IMP] website_sale_product_attribute_value_filter_existing: adds filter extraction in mobile view

lef-adhoc:
The mobile view had not been taken into account and the filters appeared even though no product had this filter.

### DIFF
--- a/website_sale_product_attribute_value_filter_existing/views/templates.xml
+++ b/website_sale_product_attribute_value_filter_existing/views/templates.xml
@@ -22,4 +22,15 @@
             <attribute name="t-if">attr_values_used &amp; v</attribute>
         </xpath>
     </template>
+    <template id="o_wsale_offcanvas" inherit_id="website_sale.o_wsale_offcanvas">
+        <xpath expr="//div[@t-foreach='a.value_ids']/div" position="attributes">
+            <attribute name="t-if">attr_values_used &amp; v</attribute>
+        </xpath>
+        <xpath
+            expr="//div[@t-if=&quot;a.display_type == 'color'&quot;]//t[@t-foreach='a.value_ids']/label"
+            position="attributes"
+        >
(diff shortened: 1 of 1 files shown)


pilarvargas-tecnativa:
Please put odoo version 16.0 in the title, thanks!

pilarvargas-tecnativa:
Can you explain about your test to get to the mistake? I can't seem to replicate it in mobile view. Have you changed the template with the editor? Thanks

lef-adhoc:
@pilarvargas-tecnativa Hello, I detail the steps below
Here create an attribute with two values:
image
Here I assign one of the attribute values to a product:
image
Then in the store in desktop view it works fine, it does not have the Test 2 attribute:
image
But in the mobile view it does have the Test 2 attribute:
image
Here is how it looks like from the cell phone:
image

pilarvargas-tecnativa:
Thank you very much!

On further review, this should also be added for ‘Select’ attributes

image

lef-adhoc:
@pilarvargas-tecnativa This is working for select attributes
image
image
image

pilarvargas-tecnativa:
Ok, I have repeated the test locally without the rest of the modules and that way it is correct. The error in my screenshot is due to website
_sale_attribute_filter_form_submit. This will have to be taken into account if both modules are used to make a bridge module to make them compatible. Thanks

@pedrobaeza Would it be appropriate to include the bridge module suggestion in the roadmap?

pilarvargas-tecnativa (approved):

pedrobaeza:
@miguel-S73 maybe it's something to fix in the module `website_sale_attribute_filter_form_submit`?

lef-adhoc:
@pedrobaeza Hello, will there be any news about this?

pedrobaeza (approved):
/ocabot merge minor

PR #940, opened 2024-06-17
[16.0][FIX] website_sale_vat_required: fix test

pilarvargas-tecnativa:
The tests were using a demo data product so if at any time this product does not exist the tests start to fail. To avoid this, a new product is created and used for testing. Also to locate the checkout button it is better to locate it by its container element as the classes can change depending on the configuration and modules used in the instance and it can be a problem to locate it.

Example screenshot where the checkout button has changed its classes and is not found:

image

cc @Tecnativa

@CarlosRoca13 @chienandalu please review


chienandalu (approved):
:+1:

pedrobaeza (approved):
/ocabot merge patch

PR #941, opened 2024-06-17
[16.0][FIX] website_sale_product_attribute_filter_collapse: fix test_tour

pilarvargas-tecnativa:
When using the demo product "Customizable Desk" in the standard behaviour there is an intermediate step in which the variant of the product to be added to the cart is chosen. This step may not be there if there are no variants really because some module has modified this and to avoid errors in the tests it is better to create a product in which we are sure that there are no variants and the step of selecting the variant before adding to the cart does not exist.

cc @Tecnativa


pedrobaeza (approved):

CarlosRoca13 (approved):
:+1:

PR #943, opened 2024-06-19
[16.0][DOC] website_sale_hide_price: message

danielduqma:
Helps and readme for message fields are clarified, as it can seem that message is shown on every case of hiding prices.

Anyway, this message is used in two places:
- In [product page](https://github.com/OCA/e-commerce/blob/b0ac4991e6809f2504f684da8763fb131edffd63/website_sale_hide_price/views/website_sale_template.xml#L23), where it is shown only when prices are hidden because of product configuration, not customer o website checks.
- In [search results](https://github.com/OCA/e-commerce/blob/828ce541fe6e5ea2c1b6ad08436e46d74e5a0529/website_sale_hide_price/models/product_template.py#L54-L55), on every case

What would be the right thing to do? This PR is focused on the first case.

FL-556-3533

### DIFF
--- a/website_sale_hide_price/README.rst
+++ b/website_sale_hide_price/README.rst
@@ -65,6 +65,9 @@ If you want to hide prices at website level
 #. Go to *Website > Configuration > Settings* and choose the website to hide the prices.
 #. In the *Shop - Products* section you have the option *Hide prices on website*
 
+Both *Default Hidden price message* field at website level and *Hidden price message*
+at product level are only shown on product page when price is hidden at product level.
+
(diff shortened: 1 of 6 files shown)


chienandalu (approved):
Thanks :)

oihane (approved):
LGTM

pedrobaeza (approved):
`[DOC]` as tag doesn't exist. It should be `[IMP]`, but let's not block for this time.

Please do the translations on Weblate when this gets merged.

/ocabot merge minor

PR #944, opened 2024-06-19
[16.0][FIX] website_sale_product_attachment: product template view ad…

miguelcb2003:
This solution fixes a bug  incorpored at https://github.com/OCA/e-commerce/commit/7f6d3e53bd0b33c98bff8fd239ed3a7f491851c0 which produces that the alert and table introduced for product templete view didn't show properly.
image

With this adjustment, both of them are restricted using a colspan, making them fill the correct grid boxes.

FL-556-3020

PR #945, opened 2024-06-27
[17.0][MIG] website_sale_product_attachment

adriresu:
Standar migration.


pedrobaeza:
Please include #944 and check CI.

dalonsod:
> Please include #944 and check CI.

#944 cherry-picked :+1: 

Regarding to test issue, I'm newbie in web tour tests, I might need some help or links with an example. I've inspected code anyway and haven't found which the problem could be. @yajo @victoralmau could you help me? Thanks!

yajo:
According to the error you find in logs:

    The ready "odoo.isTourReady('website_sale_product_attachment_tour')" code was always falsy

It looks like the wanted tour isn't even loaded. Maybe it's because its js definition is lost, or maybe because its trigger condition is never met.

If you run odoo with `--screenshots ./here --screencasts ../here` and have a valid local Chromium and ffmpeg installation, Odoo will save artifacts there that will help you to diagnose the problem.

PR #954, opened 2024-07-18
[17.0][MIG] website_sale_suggest_create_account

cem-adhoc:


chienandalu (approved):
Code and functional review :+1: 


please review @pilarvargas-tecnativa

website_sale_suggest_create_account/views/website_sale.xml
```diff
@@ -1,105 +1,63 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
     <template
-        id="cart"
-        inherit_id="website_sale.cart"
+        id="navigation_buttons"
```
pilarvargas-tecnativa, line 4: You should make a migration script for this change in the template id. It is not advisable to leave duplicate old views.

cem-adhoc: @pilarvargas-tecnativa Do you have any example of a migration script for this? It would be helpful to see how these cases are handled specifically. Thank you.

pilarvargas-tecnativa: For example:

https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/loyalty/16.0.1.0/pre-migration.py#L548-L564

But you must create the directory (migrations) and the migration files (pre or post) in the module itself.

cem-adhoc: @pilarvargas-tecnativa done. can you review it?

pilarvargas-tecnativa (approved):
Good! thanks! :D

PR #963, opened 2024-08-19
[IMP] website_sale_product_assortment: bullet list

SirAionTech:
Better bullet list rendering

### DIFF
--- a/website_sale_product_assortment/readme/CONFIGURE.rst
+++ b/website_sale_product_assortment/readme/CONFIGURE.rst
@@ -1,14 +1,14 @@
 To see this module working, you have to define a product assortment and select
 an option on the website availability field.
 
-#. **Don't apply restriction**: This option will not set any kind of restriction on
-product items.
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This is not `[REF]`, not being a code refactor, but `[IMP]` (or even `[FIX]`, as indeed this shouldn't be a numbered list).

SirAionTech:
> This is not `[REF]`, not being a code refactor, but `[IMP]` (or even `[FIX]`, as indeed this shouldn't be a numbered list).

Thanks for having a look!
It was a `REF` in my opinion because it does not `FIX` or `IMP`rove code, but `IMP` is ok too since it `IMP`roves the README, or I can use `FIX`, I can't see any clearly right/wrong tag.
Maybe a `DOC` would be more appropriate, but it does not exist :man_shrugging:

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #964, opened 2024-08-20
[16.0][FIX] website_sale_checkout_skip_payment: translate payment message

chienandalu:
- The default option in the field isn't detected as term to translate.
- Also, in a multilingual site we should provide that message in each language.

cc @Tecnativa TT50516

please review @pedrobaeza @victoralmau

### DIFF
--- a/website_sale_checkout_skip_payment/models/website.py
+++ b/website_sale_checkout_skip_payment/models/website.py
@@ -1,7 +1,7 @@
 # Copyright 2017 Sergio Teruel <sergio.teruel@tecnativa.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo import fields, models
+from odoo import _, fields, models
 from odoo.http import request
 
 
@@ -11,7 +11,10 @@ class Website(models.Model):
     website_sale_checkout_skip_message = fields.Text(
         string="Website Sale SKip Message",
         required=True,
-        default="Our team will check your order and send you payment information soon.",
+        translate=True,
+        default=lambda s: _(
+            "Our team will check your order and send you payment information soon."
+        ),
     )
     checkout_skip_payment = fields.Boolean(compute="_compute_checkout_skip_payment")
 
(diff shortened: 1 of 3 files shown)


website_sale_checkout_skip_payment/models/website.py
```diff
@@ -11,7 +11,10 @@ class Website(models.Model):
     website_sale_checkout_skip_message = fields.Text(
         string="Website Sale SKip Message",
         required=True,
-        default="Our team will check your order and send you payment information soon.",
+        translate=True,
+        default=lambda s: _(
```
pedrobaeza, line 15: I think this is not correct, as this should have all the languages in the default, not just one...

chienandalu: I don't quite get you... this is so we can trigger the translation term generation by Odoo. See here for example: https://github.com/OCA/OCB/blob/60933dcefcfba93a9a52439aa227923bde3afeb6/addons/website_event_track/models/event_track_stage.py#L21

pedrobaeza: I mean that we want to have from the beginning all the translations loaded into this field, not only the user's language one.

chienandalu: Ok, I understand but I can't figure out how to do that from the default value. The `default_get` method returns a simple key value for direct write...

pedrobaeza: Odoo is not doing this in any of their modules? Maybe if you return a JSON with languages as keys?

chienandalu: I can't find the way. A simple write will have this effect:

```python
>>> website.write({"website_sale_checkout_skip_message": {"en_US": "This is english", "es_ES": "Esto es español"}})
>>> website.website_sale_checkout_skip_message
"{'en_US': 'This is english', 'es_ES': 'Esto es español'}"
```

And with good reason, as it goes through the `convert_to_cache` method for that field class. The same goes for the `default_get` method https://github.com/OCA/OCB/blob/60933dcefcfba93a9a52439aa227923bde3afeb6/odoo/fields.py#L1245

So no black magic here, I'm afraid. And as far as I could tell it looks like Odoo just goes by with that limitation. An example would be the code I linked before...

pedrobaeza: Uhm, this is ugly... @HydrionBurst isn't there any way to initialize a field with all the translated strings?

pedrobaeza (approved):
Let's fix this while @HydrionBurst tells us anything.

/ocabot merge patch

PR #965, opened 2024-08-20
[15.0][FIX] website_sale_checkout_skip_payment: translate payment message

chienandalu:
- The default option in the field isn't detected as term to translate.
- Also, in a multilingual site we should provide that message in each language.

cc @Tecnativa TT50516

please review @pedrobaeza @victoralmau 

fw of 
- #964


pedrobaeza (approved):
The same as #964

/ocabot merge patch

PR #970, opened 2024-09-12
[16.0][FIX] website_sale_product_attribute_value_filter_existing: Hide attribute parent container

pilarvargas-tecnativa:
If only the attribute container is hidden, the parent container still occupies a space that is visible in the attribute list. To avoid this, apply the condition to the parent container and hide it completely.

This occurs in the filtering of attributes in the mobile view.

Attributes defined for the test:
image

Attributes 2, 3 and 4 will remain hidden as no product has been assigned to them.

Before the fix:
image

After the fix:
image

@chienandalu @carlos-lopez-tecnativa please review

ping @pedrobaeza 

cc @Tecnativa TT50704


pedrobaeza (approved):

chienandalu (approved):
:+1:

PR #978, opened 2024-10-23
[18.0][MIG] website_sale_product_attribute_value_filter_existing: Migration to 18.0

pilarvargas-tecnativa:
cc @Tecnativa TT51372

@CarlosRoca13 @chienandalu please review


CarlosRoca13 (approved):

pedrobaeza:
/ocabot migration website_sale_product_attribute_value_filter_existing

pedrobaeza (approved):
Tested in runboat

/ocabot merge nobump

Arozeo:
@pilarvargas-tecnativa Thanks for your work on updating this **website_sale_product_attribute_value_filter_existing** module !

I'm trying to retrofit the v18 version to a 17.4 as I cannot update our database yet and I'm facing this exact problem

All possible values for a given attribute are displayed in front end filters even though only 2 or 3 are used on actual displayed product, which is nonsensical for eCommerce

This is very explicit for the Brand attribute for example, and should be standard IMO. Why display "Category A's brands" when looking at Category B

Would you have any input on how to do that retrofit, as the v17 version of this module was skipped 🙂?

I can import the module but then I suppose I end up with an empty attr_values_used because only the names of my attributes are displayed. Tried to tweak it a bit with ChatGPT as I am not a dev at all but without any success unfortunately

Many thanks

Eliott

pedrobaeza:
I'm afraid you can't do it, as SaaS doesn't allow to import modules with, only data modules, which are a few. Most of the modules require extra code to work. You should go out of the SaaS (with the costs it implies), and be on a .0 version.

Arozeo:
Hi Pedro

Thanks for your very prompt answer !

I am currently using the "Applications" module from the Backoffice and them "Import an app" button in the atop menu
There I import the whole source code folder as a .zip

So in essence you're implying that even though the module actually gets imported without any error, some of the code inside is disabled because I am in a SaaS version ?

Thanks

pedrobaeza:
Exactly.

Arozeo:
WOW OK it explains so much things ! Thanks a lot and sorry for disturbing 🙂

Have a great weekend

PR #979, opened 2024-10-25
[16.0][FIX] website_sale_product_attribute_value_filter_existing: Add attribute ‘Test yellow'

pilarvargas-tecnativa:
It is being checked that the attribute ‘Test yellow’ is not present in the filter panel as it is not being used in any product. Actually, no real check is being made if the attribute is not defined. It is a fact, if it is not defined, it does not exist and will not be present at all.

cc @Tecnativa 


CarlosRoca13 (approved):
:+1:

pedrobaeza (approved):
/ocabot merge patch

PR #986, opened 2024-11-20
[16.0][FIX]website_sale_hide_price: fix button visibility issue

LuisAlejandroS:
Description:

In the website_sale_hide_price module, button visibility for the "Quick Add to Cart" button in the ecommerce product list (Kanban or List View) was not behaving as expected when the stock was zero and the "Allow out of stock order" option was unchecked. This PR resolves the issue by replacing view inheritance with logic in Python.

Cases covered with this improvement:

- Ensures that the "Quick Add to Cart" button disappears when stock is zero and "Allow out of stock order" is unchecked.
- Resolves compatibility issues with website_sale_hide_price module.

Reproduction Steps (in Runboat):

1. Create a product in the Sales > Products menu and set the stock to 0.
2. Make sure the "Allow out of stock order" option is unchecked.
3. Enable the website_sale_hide_price module.
4. In the ecommerce frontend, go to the product listing or Kanban view.
5. Verify that the "Quick Add to Cart" button is hidden when stock is zero.

Solution: 
The issue was fixed by modifying the _website_show_quick_add method in the product.template model to include logic for checking stock availability and the "Allow out of stock order" option in Python, instead of relying on view inheritance.

This approach ensures better maintainability and resolves the dependency issues that caused the button visibility error.
FL-556-4678

### DIFF
--- a/website_sale_hide_price/models/product_template.py
+++ b/website_sale_hide_price/models/product_template.py
@@ -1,6 +1,7 @@
 # Copyright 2022 Tecnativa - David Vidal
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from odoo import fields, models
+from odoo.tools import config
 
 
 class ProductTemplate(models.Model):
@@ -58,3 +59,14 @@ class ProductTemplate(models.Model):
                     }
                 )
         return results_data
+
+    def _website_show_quick_add(self):
+        show_price = (
+            self.env["website"].get_current_website().website_show_price
+            and not self.website_hide_price
+        )
+        if config["test_enable"] and not self.env.context.get(
+            "website_sale_hide_price_test"
+        ):
+            return show_price
+        return show_price and super()._website_show_quick_add()
--- a/website_sale_hide_price/views/website_sale_template.xml
+++ b/website_sale_hide_price/views/website_sale_template.xml
@@ -82,14 +82,6 @@
             </attribute>
         </xpath>
     </template>
-    <!-- Hide add to cart button if "Add to cart" option is enabled and not website_show_price -->
-    <template id="products_add_to_cart" inherit_id="website_sale.products_add_to_cart">
-        <xpath expr="//a[hasclass('a-submit')]" position="attributes">
-            <attribute name="t-if">
-                website.website_show_price and not product.website_hide_price
-            </attribute>
-        </xpath>
-    </template>
     <template id="website_search_box" inherit_id="website.website_search_box">
         <xpath expr="//input[@name='search']" position="attributes">
             <attribute


chienandalu (commented):
Is it ready for review or did you open it in draft by mistake?

Pablocce (approved):
LGTM, thanks

danielduqma (approved):

chienandalu (commented):
@pilarvargas-tecnativa could take a look?

pilarvargas-tecnativa (commented):
I don't see any difference in the behaviour of the module with this FIX, can you put an indicative image of the problem? I can't reproduce the problem you are referring to outside this PR.

LuisAlejandroS:
image
A previous feature ensured that when the "Continue Selling" checkbox was not checked and a product was out of stock, the "Add to Cart" button wouldn't display in the kanban or list views. This update restores that functionality, resolving conflicts caused by this module.

pilarvargas-tecnativa:
Ok, now I understand the problem. Maybe with this solution you don't have to apply that condition to other buttons at template level. For example those in some snippets.
Could you check if in the next versions there is this problem and bring the commit? 
Thanks!

pilarvargas-tecnativa (approved):

LuisAlejandroS:
I noticed this issue persists in version 17.0. To address it, I created a forward-port with the necessary adjustments. You can review the pull request here: https://github.com/OCA/e-commerce/pull/987.

pedrobaeza (approved):
Having a method that determines the visibility and override it makes it more inheritable that touching the QWeb template, so let's go with the change:

/ocabot merge patch

pedrobaeza:
@LuisAlejandroS it seems the tests are failing.

danielduqma:
Hi,

Tests that are failing are related to module `website_sale_stock_list_preview`. This module depends on `website_sale_hide_price` to be installed and improperly show the button.

In fact, if you just install `website_sale_stock_list_preview` (with no other modules from this repository) tests will also fail, because `website_sale_stock` doesn't render _Add to cart_ button that `website_sale_stock_list_preview` expects to be shown but disabled. In forward port to 17.0, tests doesn't fail because `website_sale_stock_list_preview` is not yet migrated.

As you are the authors of this module, could you please take a look at this? Thanks!

Regards.

pedrobaeza:
Well, this was not failing before your patch, and that's because views are enable on demand on tests, so you should find a way to coexist both modules. One option is to incorporate on the code a check on `config["test_enable"]` and then only act if there's a special context that means that you are testing this module, not others.

If you don't find a solution, we will need to revert the 17.0 PR for not having the problem on the future.

danielduqma:
To be honest, I don't see that point. There is a clear bug in `website_sale_hide_price` that breaks a behaviour implemented in `website_sale_stock` from Odoo core, and it can be easily tested in both 16.0 and 17.0 versions. I think that we agree until here.

There is a third module, `website_sale_stock_list_preview`, that dependes on `website_sale_hide_price` to show a button that `website_sale_stock` hides. What you propose is to revert the fix in 17.0 (and willfully keep a bug) and include a `test_enable` check to make the module misbehave in tests just to make `website_sale_stock_list_preview` pass them. But `website_sale_stock_list_preview` goal will be affected anyway and has to be reviewed.

I didn't pretend you to fix `website_sale_stock_list_preview` it ASAP. I just asked to know if you were going to take a first look as Tecnativa is the author and all contributors are from Tecnativa.

pilarvargas-tecnativa:
A solution at view level would be to incorporate in the t-if the existing conditions in the odoo base view + the conditions added by this module that cause the error when overwriting the whole t-if. At code level it is a good solution but it seems to cause conflict in this version. Or what @pedrobaeza said, add a context to the tests.

danielduqma:
But we don't see why adding more conditions to the `t-if` is necessary, if it already has a method that is inheritable to ease integration with other modules, as it is done in `website_sale_stock`. I don't get why this PR is conflicting, just because it arises existing bugs in other modules and we are asked to mask them.

pedrobaeza:
As said, it's not a problem in the other module, as how tests are executed, they are correct. The status quo is broken with this change.

LuisAlejandroS:
@pedrobaeza Hi! I’ve made a small adjustment to ensure the tests for your module don’t break. Feel free to review it whenever you have the chance. Thanks in advance!

website_sale_hide_price/models/product_template.py
```diff
@@ -58,3 +59,12 @@ def _search_render_results(self, fetch_fields, mapping, icon, limit):
                     }
                 )
         return results_data
+
+    def _website_show_quick_add(self):
+        show_price = (
+            self.env["website"].get_current_website().website_show_price
+            and not self.website_hide_price
+        )
+        if config["test_enable"]:
```
pedrobaeza: You have to test also for a specific context (like `website_sale_hide_price_test`), for allowing to check the functionality in this module (if having tests).

LuisAlejandroS: @pedrobaeza  I’ve just made the changes. Could you please check if this works for you now? Thanks for your time!

pedrobaeza (approved):
/ocabot merge patch

PR #990, opened 2024-12-06
[18.0][MIG] product_template_multi_link: Migration to 18.0

thienvh332:

PR #992, opened 2024-12-06
[18.0][MIG] product_variant_multi_link: Migration to 18.0

thienvh332:
### Depends on:
- https://github.com/OCA/e-commerce/pull/990


product_variant_multi_link/tests/test_product_template_link.py
```diff
@@ -3,10 +3,10 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo.exceptions import ValidationError
-from odoo.tests.common import SavepointCase
+from odoo.tests.common import TransactionCase
 
 
-class TestProductVariantLink(SavepointCase):
+class TestProductVariantLink(TransactionCase):
```
sebalix: You can inherit from `BaseCommon` class and remove the `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))` line in `setUpClass` method

thienvh332: I updated it. Thanks :pray:

sebalix (approved):
LG, `DON'T MERGE` commit can be removed, dependency is merged

PR #999, opened 2024-12-20
[15.0][ADD] website_sale_menu_partner_top_selling: New module

pilarvargas-tecnativa:
cc @Tecnativa TT52287

@pedrobaeza @chienandalu please review


pilarvargas-tecnativa:
I know the tests are failing but replicating in the browser console those triggers don't fail, I have to investigate further why.

website_sale_menu_top_user_selling/controllers/main.py
```diff
@@ -0,0 +1,90 @@
...
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param("website_sale_menu_top_user_selling.limit", 10)
+        )
+        # Get best-selling products from the user
+        product_data = (
+            request.env["sale.order.line"]
+            .sudo()
+            .read_group(
+                [("order_id.partner_id", "=", request.env.user.partner_id.id)],
```
chienandalu: Maybe you should restrict the website or at least the company..

chienandalu (changes requested):
This is problematic as well

image

website_sale_menu_top_user_selling/controllers/main.py
pilarvargas-tecnativa: Better to limit both and leave no loose ends. The problem is that they were not well declared in the manifest, haste is not good. Actually, whether they are published or not is not a problem since they are running as administrator and will appear shaded.

chienandalu (approved):
Great :+1:

pilarvargas-tecnativa:
@carlos-lopez-tecnativa can you take a look please?

website_sale_menu_partner_top_selling/controllers/main.py
```diff
@@ -0,0 +1,110 @@
...
+                    "|",
+                    ("product_id.product_tmpl_id.website_id", "=", request.website.id),
+                    ("product_id.product_tmpl_id.website_id", "=", False),
+                    "|",
+                    (
+                        "product_id.product_tmpl_id.company_id",
+                        "=",
+                        request.website.company_id.id,
+                    ),
+                    ("product_id.product_tmpl_id.company_id", "=", False),
```
carlos-lopez-tecnativa: I think this can be simplified to:
```suggestion
                    ("product_id.product_tmpl_id.website_id", "in", [request.website.id, False]),
                    (
                        "product_id.product_tmpl_id.company_id",
                        "in",
                        [request.website.company_id.id, False],
                    ),
```

website_sale_menu_partner_top_selling/models/res_config_settings.py
```diff
@@ -0,0 +1,30 @@
...
+            ),
+        )
+        return res
+
+    def set_values(self):
+        res = super().set_values()
+        self.env["ir.config_parameter"].sudo().set_param(
+            "website_sale_menu_partner_top_selling.limit", self.regular_products_limit
+        )
+        return res
```
carlos-lopez-tecnativa: I think this can be simplified as follows or defined as a field in the website model instead of a parameter. What do you think?
https://github.com/odoo/odoo/blob/15.0/odoo/addons/base/models/res_config.py#L339C48-L339C78

```suggestion
    regular_products_limit = fields.Integer(default=10, config_parameter='website_sale_menu_partner_top_selling.limit')

```

carlos-lopez-tecnativa (approved):
Tested and working fine. I’ve left a suggestion; let me know if it can be considered. Thanks, great feature!

website_sale_menu_partner_top_selling/models/res_config_settings.py
pilarvargas-tecnativa: thanks!

website_sale_menu_partner_top_selling/readme/USAGE.rst
```diff
@@ -0,0 +1,6 @@
+Regular products can be viewed by accessing the URL /shop/my_regular_products on the
+website.
+This URL can be accessed from */shop* by clicking on *My regular products* in the
```
pedrobaeza, line 4: RST syntax forces to have a blank link in between paragraphs. You can check that is not working right in https://github.com/OCA/e-commerce/blob/8cbf5ba0d5c7eaabd6c36c99c49813569b16affd/website_sale_menu_partner_top_selling/README.rst

```suggestion

This URL can be accessed from */shop* by clicking on *My regular products* in the
```

(and the same for the rest)

pilarvargas-tecnativa: Done, thanks

PR #1004, opened 2025-01-14
[16.0][ADD] website_sale_commitment_date

adasatorres:
Hi, this addon allows selecting the delivery date when purchasing from the e-commerce platform, configuring excluded days of the week, minimum or maximum days to limit the selection
from the current date. All of this is configurable from the shipping method.


adasatorres:
To use this addon, you need to go to the shipping method and activate the "allow commitment date" option.
Then, you can configure various settings right there. Afterward,
go to the cart in the e-commerce platform, and you will see that if you select that shipping method,
the option to choose the delivery day will appear.
When using a shipping method that requires specifying the delivery date,
the button will not appear until a date is selected.

website_sale_commitment_date/controllers/main.py
```diff
@@ -0,0 +1,88 @@
...
+    def update_commitment_date(self, **post):
+        order = request.website.sale_get_order()
+        string_date = post.get("date")
+        lang = request.env["res.lang"]._lang_get(
+            request.env.context.get("lang", False) or request.env.user.lang
+        )
+        if order:
+            date_obj = datetime.strptime(
+                string_date,
+                lang.date_format if lang else DEFAULT_SERVER_DATE_FORMAT,
```
rrebollo, line 40: Perhaps you can use "from odoo.tools.misc import format_date" to the date manipulation according to language and encapsulate this behavior in a separated auxiliary method to reuse below too.

website_sale_commitment_date/models/delivery.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    def _get_calendar_context(self, start=False):
+        today = fields.Date.today()
+
+        if start:
+            year, month = start.year, start.month
+        else:
+            year, month = today.year, today.month
+            start = date(year, month, 1)
```
rrebollo:
```suggestion
            start = fields.Date.start_of(today, "month")
```
Relays on odoo built-in tooling

website_sale_commitment_date/models/delivery.py
```diff
@@ -0,0 +1,108 @@
...
+    name = fields.Char(
+        compute="_compute_name", store=True, readonly=False, string="Label"
+    )
+    value = fields.Date(required=True, string="Date")
+
+    @api.depends("value")
+    def _compute_name(self):
+        for record in self:
+            if record.value:
+                record.name = record.value.strftime("%d/%m/%Y")
```
rrebollo: Why ""%d/%m/%Y" as format? Wouldn't "%Y/%m/%d" be "safer"  instead?

rrebollo (commented):
Code Review: Great work! The code looks good to me (LGTM). Thank you for your contribution! I've provided a few suggestions for your consideration—feel free to address them as you see fit. Btw, nice tests.

website_sale_commitment_date/models/delivery.py
adasatorres: Hi @rrebollo, I could use the date format provided by the language, so the format will be created based on the user's language.

website_sale_commitment_date/controllers/main.py
adasatorres: Hi @rrebollo, I believe that format_date cannot be used to create a date object from a string.

website_sale_commitment_date/models/delivery.py
rrebollo: Yes that'll be a better approach in deed

website_sale_commitment_date/controllers/main.py
rrebollo: Of course, my suggestion implicitly mean to do some refactoring around to rely upon odoo's built-in wrappers to work with python dates. But it was only a suggestion. Again, great work

website_sale_commitment_date/models/delivery.py
```diff
@@ -0,0 +1,151 @@
...
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    allow_commitment_date = fields.Boolean(
+        string="Allow commitment date",
+        help="Allow to set a commitment date on the sale order from to the e-commerce",
+    )
+
+    min_commitment_days = fields.Float()
+    max_commitment_days = fields.Float()
```
Honeyxilia, line 24: Since these fields contain a number of days, wouldn't it be better if this was an Integer field ?

website_sale_commitment_date/controllers/main.py
```diff
@@ -0,0 +1,88 @@
...
+from odoo.tools import DEFAULT_SERVER_DATE_FORMAT, format_date
+
+from odoo.addons.website_sale_delivery.controllers.main import WebsiteSaleDelivery
+
+_logger = logging.getLogger(__name__)
+
+
+class WebsiteSaleCommitmentDate(WebsiteSaleDelivery):
+    @http.route()
+    def payment(self, **post):
```
Honeyxilia, line 19: There's no controller method in WebsiteSaleDelivery or its parent classes, named `payment` :(
Would it be possible that this is meant to override `shop_payment` (since the prototype arguments are the same) ?

Honeyxilia (changes requested):
Hello - I installed the module locally, and I had a couple of remarks I wanted to make c:

PR #1014, opened 2025-01-29
[ADD] website_sale_browse_mode

victor-champonnois:
internal task : https://gestion.coopiteasy.be/web#id=13307&action=475&active_id=478&model=project.task&view_type=form&menu_id=536


remytms (commented):
Great, nice feature and very simple. :)

Some points:

- Not sure if the name is clear enough. Perhaps something like `website_sale_browse_only` or `website_sale_disable_shopping` or `website_sale_no_shopping`, etc. But maybe it’s just me that did catch directly the meaning.
- I think that the case when the cart is not empty when the website shift to "browse only" should be take into account. In such case the user will be able to finish the payment of his/her cart. Maybe button on the cart view should be removed also.

victor-champonnois:
@remytms 
Thanks for the review
- I hid the cart icon. It does not prevent someone who just add "/cart" to the url to access their card but it's a quick fix. 
- I renamed the module website_sale_browse_only

website_sale_browse_only/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+    "website": "https://github.com/OCA/e-commerce",
+    "license": "AGPL-3",
+    "depends": ["website_sale_stock"],
+    "data": [
+        "views/website.xml",
+        "views/res_config_settings.xml",
+        "views/website_sale_template.xml",
+    ],
+    "installable": True,
+}
```
huguesdk: the `summary` key is missing (and the `installable` key can be removed, as it’s as its default value).

victor-champonnois: fixed

victor-champonnois:
There is an issue with this PR : a banner is added when the website is closed,  but it "shifts" the whole website page down, so a part of the footer is no longer visible. As a consequence, visitors cannot change the language when the website is closed. 
I don't have time to fix it now, I set the PR as draft.

PR #1025, opened 2025-02-20
[17.0][MIG] website_sale_stock_available: Migration to version 17.0

carlos-lopez-tecnativa:
TT52396

@Tecnativa @pedrobaeza @pilarvargas-tecnativa @chienandalu could you please review this?


website_sale_stock_available/readme/CONFIGURE.md
```diff
@@ -0,0 +1,10 @@
+To configure this module, you need to:
+
+1.  Go to *Inventory \> Master Data \> Products* and edit a product that
```
pilarvargas-tecnativa:
```suggestion
1.  Go to *Inventory \> Products \> Products* and edit a product that
```

website_sale_stock_available/readme/CONFIGURE.md
```diff
@@ -0,0 +1,10 @@
...
+    you only want to sell in the eCommerce if there is enough stock.
+2.  Navigate to *Out-of-stock (continue selling)* field in the *Sales*
+    tab and unset it.
+3.  Go to *Inventory \> Configuration \> Settings*, navigate to *Stock
+    available to promise* section and set the desired option (you might
```
pilarvargas-tecnativa: I think this also needs to be updated

website_sale_stock_available/readme/CONFIGURE.md
carlos-lopez-tecnativa: Good catch! I’ve updated the README. Thanks!

website_sale_stock_available/readme/CONFIGURE.md
carlos-lopez-tecnativa: Do you remember which module adds this section? I don’t recall. I removed this section. Is that okay for you?

pilarvargas-tecnativa: I don't know which module adds it, I think it should be clarified which module it belongs to but if we don't know it is better to delete the part that doesn't correspond to the module as it can lead to confusion, what do you think?

pilarvargas-tecnativa (approved):

website_sale_stock_available/readme/CONFIGURE.md
carlos-lopez-tecnativa: This view was removed during the migration to V17 please see https://github.com/OCA/stock-logistics-availability/pull/20#discussion_r1587187600
As a result, I have removed this instruction from `CONFIGURE.md.`

carlos-lopez-tecnativa:
@chienandalu @CarlosRoca13 could you please review this?

CarlosRoca13 (approved):

PR #1027, opened 2025-02-25
[18.0][MIG] website_sale_checkout_country_vat: Migration to version 18.0

pilarvargas-tecnativa:
cc @tecnativa TT54387

@carlos-lopez-tecnativa  @CarlosRoca13 please review


carlos-lopez-tecnativa (changes requested):
When I click on the `Continue Checkout` button, I get this error.

image


Is it possible to search in this list? It's a bit complicated to find the country manually.
image

website_sale_checkout_country_vat/readme/CONFIGURE.md
```diff
@@ -0,0 +1,9 @@
+The default country that will appear is the one of the current user, or
+the one of the Public User assigned to the website, or the one of the
+company that owns the website. Change it in the corresponding place to
+get a sane default.
+
+If you want to have a sane default also in the country address, check
+the
+[website_sale_default_country](https://github.com/OCA/e-commerce/tree/10.0/website_sale_default_country)
+module.
```
carlos-lopez-tecnativa: This module has not been migrated from V11 onwards. Does it need to be referenced here?

pilarvargas-tecnativa:
> When I click on the `Continue Checkout` button, I get this error.

You can check again

> Is it possible to search in this list? It's a bit complicated to find the country manually. 

This is a matter of the base module and the way it has always worked. But it actually does the search when you select the country in the address. I think the module is meant for when you define new addresses, not when you modify them.

CarlosRoca13 (approved):
Thanks!

carlos-lopez-tecnativa (approved):
LGTM

PR #1028, opened 2025-02-26
[16.0][FIX] website_sale_order_type: Remove journal_id definition in tests to avoid side effects

victoralmau:
Remove `journal_id` definition in tests to avoid side effects.

```
odoo.exceptions.UserError: Incompatible companies on records:
- 'Test Sale Order Type' belongs to company 'My Company (San Francisco)' and 'Billing Journal' (journal_id: 'Customer Invoices') belongs to another company.
```

@Tecnativa

### DIFF
--- a/website_sale_order_type/tests/test_website_sale_order_type.py
+++ b/website_sale_order_type/tests/test_website_sale_order_type.py
@@ -28,9 +28,6 @@ class TestFrontend(HttpCase):
                 "padding": 3,
             }
         )
-        self.journal = self.env["account.journal"].search(
-            [("type", "=", "sale")], limit=1
-        )
         self.warehouse = self.env.ref("stock.warehouse0")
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Why the patch is needed here? It's not creating any invoice.

victoralmau:
Sorry for the confusion, the problem was different, it has been modified.

pedrobaeza (commented):
I think that you should remove the journal stuff (the search and assignation), as it's not needed for this test, and it's creating maintenance burden.

pedrobaeza (commented):
The principle "less is more" can be applied to tests. The test should check what the module does with the minimum dataset. Nothing more.

/ocabot merge patch

pedrobaeza:
I would even say that `warehouse` and the other optional fields may be also emptied without problems.

PR #1039, opened 2025-03-19
18.0 mig website sale product attachment

jado95:


pedrobaeza:
I think Odoo now includes this option, isn't it?

jado95:
> I think Odoo now includes this option, isn't it?

I don't find this option in Odoo standard.

PR #1040, opened 2025-03-19
[17.0][FIX]website_sale_hide_price:replace t-attf-class with hasclass()

cem-adhoc:
Fix XPath error inheriting website_sale.product_buy_now view

PR #1041, opened 2025-03-25
[FIX] website_sale_hide_price: change manifest

matiasperalta1:
Some Odoo databases with this module installed display this error. This is because a change was made to the view without modifying the manifest version.
PR without the change in manifest https://github.com/OCA/e-commerce/pull/1040

Error: psycopg2.errors.UndefinedColumn: column res_partner.website_show_price does not exist
LINE 1: ...", "res_partner"."require_purchase_order_number", "res_partn...

### DIFF
--- a/website_sale_hide_price/__manifest__.py
+++ b/website_sale_hide_price/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Website Sale Hide Price",
-    "version": "17.0.1.1.2",
+    "version": "17.0.1.2.2",
     "category": "Website",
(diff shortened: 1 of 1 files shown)


matiasperalta1:
@pedrobaeza Could you do a review on this change?

chienandalu:
The version is bumped post-merge:

https://github.com/OCA/e-commerce/commit/b67315182c1030fce3e9b6f5d3841b4a1ab0fc6e#diff-907b5b2df6397855bf75d02ed1e989930aa09f4e596dc76063de8ac745d04048

Anyway, I can't see how that change might be related to that error :thinking:

matiasperalta1:
@chienandalu You're right, the errors occurred during the days the manifest wasn't updated. I've already made the corresponding update and everything has been resolved. We've closed the PR.

cc @pedrobaeza

Issue #855, opened 2023-10-11
[16.0] website_sale_hide_price - Show price in comparison tool and wishlist


sergiuga:
## Module
website_sale_hide_price

## Describe the bug
After restricting access to public user, the price is still available in wishlist and comparison tool and products can be added to cart.

## To Reproduce
Version affected - Odoo 16 CE
1. Restrict public user to see prices in shop - working on shop page and product page
2. Add products to wishlist and to comparison tool
3. Product prices and "Add to cart" button available and working

**Expected behavior**
No "Add to cart" and product price available to restricted products
**Additional context**
![website_hide_price_nohide](https://github.com/OCA/e-commerce/assets/28352258/d8e21236-f0c6-46dd-81f0-0fbfc3cdc80f)

pedrobaeza:
Install https://github.com/OCA/e-commerce/tree/16.0/website_sale_comparison_hide_price for the first one and a new module `website_sale_wishlist_hide_price` is needed for the second.

PR #1043, opened 2025-04-08
[17.0] [ADD] website_sale_hide_price_wishlist

maciej-wichowski:
Based on issue #855 
Modules introduced here will extend functionality of `website_sale_hide_price` to fit into wishlist feature. Besides `website_sale_hide_price_wishlist`, I'm introducing `website_sale_stock_hide_price_wishlist`, so "Add to cart" buttons are hidden in all possible cases.

obraz


pilarvargas-tecnativa (changes requested):
There is already a module for this task, you just need to perform the migration.

https://github.com/OCA/e-commerce/tree/16.0/website_sale_wishlist_hide_price

PR #1045, opened 2025-04-10
[16.0][IMP-FIX] website_sale_tax_toggle: Set the value option to show tax also for the product page

pilarvargas-tecnativa:
@Tecnativa TT55895

@pedrobaeza @CarlosRoca13 please review


pedrobaeza (approved):
That's the correct way of injecting the value.

/ocabot merge patch

PR #1064, opened 2025-06-19
[18.0][MIG] website_sale_product_sort: Migration to version 18.0

pilarvargas-tecnativa:
@Tecnativa TT54394

@CarlosRoca13 @juancarlosonate-tecnativa please review


website_sale_product_sort/models/website.py
```diff
@@ -0,0 +1,27 @@
...
+    @api.model
+    def _get_product_sort_criterias(self):
+        """Extend to add more sort criterias"""
+        return [
+            ("website_sequence asc", _("Featured")),
+            ("create_date desc", _("Newest Arrivals")),
+            ("name asc", _("Name (A-Z)")),
+            ("name desc", _("Name (Z-A)")),
+            ("list_price asc", _("Price - Low to High")),
+            ("list_price desc", _("Price - High to Low")),
```
juancarlosonate-tecnativa:
```suggestion
            ("website_sequence asc", self.env._("Featured")),
            ("create_date desc", self.env._("Newest Arrivals")),
            ("name asc", self.env._("Name (A-Z)")),
            ("name desc", self.env._("Name (Z-A)")),
            ("list_price asc", self.env._("Price - Low to High")),
            ("list_price desc", self.env._("Price - High to Low")),
```

website_sale_product_sort/models/website.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2020 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import _, api, fields, models
```
juancarlosonate-tecnativa:
```suggestion
from odoo import api, fields, models
```

juancarlosonate-tecnativa (approved):
LGTM, non-blocking comments: Wouldn't it be possible to remove the empty option from the selector to avoid this validation error, and have it fall back to the default value or something like that?

image
image

pilarvargas-tecnativa:
> LGTM, non-blocking comments: Wouldn't it be possible to remove the empty option from the selector to avoid this validation error, and have it fall back to the default value or something like that?

An empty value is no longer allowed in the selection field. The reason for the error is that the message indicating that the field is required was not displayed; the ORM simply throws an error stating that it depends on another model. It is not a very explicit message, but in any case, it will not happen again.

Thanks!

juancarlosonate-tecnativa (approved):

pilarvargas-tecnativa:
@victoralmau can you take a look please?

website_sale_product_sort/views/res_config_settings_view.xml
```diff
@@ -0,0 +1,25 @@
...
+            >
+                <setting
+                    id="product_sort_criteria_setting"
+                    string="Sort Criteria"
+                    help="Default Sort Criteria"
+                >
+                    <field
+                        name="product_sort_criteria"
+                        class="o_light_label"
+                        required="1"
```
victoralmau:
```suggestion
```
You should not add a required here, as this would be problematic: https://github.com/OCA/management-system/pull/700

website_sale_product_sort/views/website_sale_templates.xml
```diff
@@ -0,0 +1,28 @@
...
+        <xpath expr="//t[@t-set='website_sale_sortable']" position="attributes">
+            <attribute name="t-value">website._get_product_sort_criterias()</attribute>
+        </xpath>
+    </template>
+
+    <template id="sort" inherit_id="website_sale.sort" priority="9999">
+        <xpath expr="//span[@t-field='website.shop_default_sort']" position="replace">
+            <t t-else="1">
+                <t t-if="is_view_active('website_sale_product_sort.products')">
+                    <span t-field="website.default_product_sort_criteria" />
```
victoralmau, line 22:
```suggestion
                    <span t-if="website.default_product_sort_criteria" t-field="website.default_product_sort_criteria" />
                    <span t-else="" t-field="website.shop_default_sort" />
```
Related to my previous comment, I believe this would help avoid any mistakes.

website_sale_product_sort/readme/CONFIGURE.md
```diff
@@ -0,0 +1,16 @@
+To configure this module, you need to:
+
+1.  Go to *Website \> Configuration \> Settings*
+2.  Select the website you want to configure.
+3.  In the *Product* section there's a *Sort Criteria* option that you
```
victoralmau:
```suggestion
3.  In the *Shop - Products* section there's a *Sort Criteria* option that you
```

victoralmau (changes requested):
Functional review OK

website_sale_product_sort/views/website_sale_templates.xml
pilarvargas-tecnativa: thanks!

victoralmau (approved):

PR #1067, opened 2025-06-24
[18.0][ADD] website_sale_product_public_category_product

jbaudoux:
* Allow to manage the list of products from the ecommerce category
* Add filter by eCommerce category on the products list


jbaudoux:
@OCA/e-commerce-maintainers Can you reopen and merge if there is no objection?

pedrobaeza:
I think `public` from the module name is useless and make it longer, as you are already saying `website_sale`, don't you think?

jbaudoux:
> I think `public` from the module name is useless and make it longer, as you are already saying `website_sale`, don't you think?

Thanks for the review. I like to be explicit and prevent any confusion as category and public category are 2 different things.

pedrobaeza:
But you are repeating 2 times `product` and the module purpose is not very explicit. If you want that, I would remove then `website_sale`...

PR #1069, opened 2025-06-24
[16.0][MIG] website_sale_affiliate: Migration to 16.0

fcayre:


website_sale_affiliate/models/sale_order.py
```diff
@@ -0,0 +1,21 @@
...
+        string="Affiliate request",
+        help="Affiliate request associated with sale order",
+    )
+
+    @api.model
+    def create(self, vals):
+        res = super().create(vals)
+        AffiliateRequest = self.env["sale.affiliate.request"]
+        res.affiliate_request_id = AffiliateRequest.current_qualified()
+        return res
```
Honeyxilia: This create method override doesn't include the batch create feature (`model_create_multi`) and it triggers a WARNING when installing the module
=> It would be better to modify the method in order to support batch creation of sale orders

Also the variables names are in Camel case, and should rather be Snake case to fit the PEP8 norm.

website_sale_affiliate/models/sale_affiliate_request.py
```diff
@@ -0,0 +1,94 @@
...
+
+from datetime import timedelta
+
+from odoo import fields, models
+from odoo.http import request
+
+
+class AffiliateRequest(models.Model):
+    _name = "sale.affiliate.request"
+    _order = "create_date desc"
```
Honeyxilia, line 12: this class and `sale.affiliate` have no`_description` value set - it would be better to include them in order to clarify their purposes.

Honeyxilia (changes requested):
Aside from these small non-blocking changes, the module runs well and the rest of the code is ported well - so after these changes, I would confirm the changes.

fcayre:
@Honeyxilia remarks taken into account

Honeyxilia (approved):
LGTM (code review + live testing)

Honeyxilia:
@gfcapalbo Hello, we would appreciate a review of this PR

PR #1072, opened 2025-06-25
[17.0][MIG] website_sale_comparison_hide_price: Migration to 17.0

mide-dynapps:


website_sale_comparison_hide_price/views/website_sale_template.xml
```diff
@@ -0,0 +1,44 @@
...
+        <xpath expr="//form[@action='/shop/cart/update']" position="attributes">
+            <attribute name="t-if">
+                website.website_show_price and not product.website_hide_price
+            </attribute>
+        </xpath>
+    </template>
+    <template
+        id="compare_popover_hide_price"
+        inherit_id="website_sale_comparison.product_product"
+    >
```
cgarcia-solvos, line 32: Hi,
I was testing this addon and found an issue: the discount is not being hidden in the comparator popover, as shown in the attached image:

imagen

Instead of hiding only the individual elements, wouldn’t it be better to hide the entire wrapping "&lt;div&gt;" to avoid this behavior?

dalonsod: hello @mide-dynapps would you address this?

ivantodorovich:
Superseeded by https://github.com/OCA/e-commerce/pull/1190

PR #1076, opened 2025-07-09
[16.0][MIG] website_sale_affiliate_product_restriction

Honeyxilia:
(Depends on https://github.com/OCA/e-commerce/pull/1069 )

PR #1077, opened 2025-07-11
[16.0][ADD+MIG] website_sale_affiliate_portal : Display affiliations to the portal user

Honeyxilia:
(Depends on https://github.com/OCA/e-commerce/pull/1076 and https://github.com/OCA/e-commerce/pull/1069 )


website_sale_affiliate_portal/tests/test_sale_affiliate.py
```diff
@@ -19,27 +20,39 @@ def setUpClass(cls):
         cls.pp3 = cls.env.ref("product.product_product_3")  # Desk Combination
 
     def create_sales(self):
-        sess = self.fake_session
-        sess["affiliate_request"] = self.create_affiliate_req("2018-01-05").id
-        self.create_sale([(self.pp1, 1)], "draft", create_date="2018-01-05")
+        with MockRequest(self.env) as mock_request:
+            mock_request.session["affiliate_request"] = self.create_affiliate_req(
```
fcayre: Maybe use `sess = mock_request.session` to:
- avoid repeating yourself
- keep lines short for better readability
- ease code review (comparison with previous version).

website_sale_affiliate_portal/models/sale_affiliate.py
```diff
@@ -57,9 +57,7 @@ def report_data(self, month_num=None):
 
         oldest = None
         if month_num is not None:
-            oldest = (
-                date.today().replace(day=1) - relativedelta(months=month_num)
-            ).strftime(fields.DATETIME_FORMAT)
+            oldest = datetime.today().replace(day=1) - relativedelta(months=month_num)
```
fcayre: It is useless to name the variable in the commit message. Proposal:

Fix invalid comparison between a date string and a datetime

We now use plain datetimes all the way.

(note there is typo in your message: compaired => compared)

fcayre: `datetime.today()` is the same as `datetime.now()` which is less confusing here and may be preferred.

website_sale_affiliate_portal/controllers/main.py
```diff
@@ -7,22 +7,18 @@
...
-        response.qcontext.update(
-            {
-                "affiliate_count": affiliate_count,
-            }
-        )
-        return response
+    def _prepare_home_portal_values(self, counters):
+        values = super()._prepare_home_portal_values(counters)
+        if "affiliate_count" in counters:
+            Affiliate = request.env["sale.affiliate"].sudo()
```
fcayre: Don't use `Affiliate` (capitalized) as a variable name, prefer e.g. `affiliate_model` or `affiliate_model_sudo`.

fcayre: In the commit message, you justify your design choice (`_prepare_home_portal_values` overloading) without naming the traditional way (using `qcontext` of the template rendering object), you should, to show you have considered it but it is not enough here.

website_sale_affiliate_portal/views/website_portal_sale_templates.xml
```diff
@@ -4,7 +4,7 @@
   <template
         id="portal_my_home_affiliations"
         name="Show Affiliations"
-        inherit_id="sale.portal_my_home_sale"
+        inherit_id="portal.portal_my_home"
```
fcayre: This is arguable, I would not do this in the migration. The `sale` module adds its own portal.portal_docs_entry sections to the o_portal_docs div, so we want to add affiliations after them probably, which is more explicit if you inherit the `sale.portal_my_home_sale` template, even if we may get the same result thanks to module load ordering.

fcayre (changes requested):

website_sale_affiliate_portal/controllers/main.py
Honeyxilia: For the Affiliate variable renaming : since it was also in the affiliations controller method, I'll do a commit dedicated to changing them both c:

website_sale_affiliate_portal/views/website_portal_sale_templates.xml
Honeyxilia: Understood - I will drop this edition

Honeyxilia:
I applied @fcayre's notes

fcayre (approved):
LGTM (code review)

Must wait for PRs https://github.com/OCA/e-commerce/pull/1076 and https://github.com/OCA/e-commerce/pull/1069 to be accepted however.

PR #1078, opened 2025-07-13
[18.0][ADD] website_sale_customer_reviews

kobros-tech:
image


kobros-tech:
@victoralmau 

I think I didn't need much effort to enable that feature in 18.0.

kobros-tech:
Screenshot from 2025-07-13 21-50-13

victoralmau:
Have you checked that this does not already exist?

Checking I see that it already exists since v16!

**v16**
v16

**v17**
v17


**v18**
v18


**master**
master

Maybe there is something I am not taking into account, but the functionality already exists, right?

kobros-tech:
yes, I just need to extend the function based on our needs.

I see it exists in all versions, thanks!

kobros-tech:
@victoralmau 

can you check on our latest update, we introduce interactive chart and statistics for product's popularity.

kobros-tech:
### Show statistics for public users

image

victoralmau:
Some comments:
- I think the name of the module could be confusing and lead people to believe that it adds reviews (which core already does), so we should find a more appropriate name that explains what it does (I can't think of one right now).
- Normally, a single commit is usually sufficient when adding a module.
- Perhaps it would be interesting to suggest a PR to Odoo with this improvement, especially if it is already in core, don't you think?

kobros-tech:
yeah, I am adding test cases and once finished I squash them all.

I don't have a name also, but let's keep it open.

maybe if we are done with the module I can PR to odoo.

PR #1083, opened 2025-07-17
[17.0][FIX] website_sale_hide_price: Hide price in product carousel

pilarvargas-tecnativa:
When the price of products was hidden at the website level, the product carousel continued to display it. To prevent this malfunction, the condition is adjusted to check at the website level whether it is permitted to display it.

@Tecnativa 

Before:

image

After:

image

@carlos-lopez-tecnativa @juancarlosonate-tecnativa please review

### DIFF
--- a/website_sale_hide_price/data/product_snippet_template_data.xml
+++ b/website_sale_hide_price/data/product_snippet_template_data.xml
@@ -9,7 +9,7 @@
         <xpath expr="//span[hasclass('fw-bold')]" position="before">
             <div
                 id="website_hide_price"
-                t-if="user_id.partner_id.website_show_price and not data.get('website_hide_price')"
+                t-if="website.website_show_price and not data.get('website_hide_price')"
             />
         </xpath>
         <xpath expr="//div[@id='website_hide_price']" position="inside">


website_sale_hide_price/data/product_snippet_template_data.xml
```diff
@@ -9,7 +9,7 @@
         <xpath expr="//span[hasclass('fw-bold')]" position="before">
             <div
                 id="website_hide_price"
-                t-if="user_id.partner_id.website_show_price and not data.get('website_hide_price')"
+                t-if="website.website_show_price and not data.get('website_hide_price')"
```
pedrobaeza, line 12: But isn't this a user decision, not website one?

pilarvargas-tecnativa: There are three ways to hide prices:
- For the entire website
- For a contact
- For a product

In this condition, the website condition was not being taken into account correctly. The contact is already taken into account in the compute method
https://github.com/OCA/e-commerce/blob/17.0/website_sale_hide_price/models/website.py#L24

pedrobaeza: `website_hide_price` and `website_show_price` are very confusing, but well... the product condition is being taken into account here?

pilarvargas-tecnativa: Yes, data contains product information.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):

PR #1079, opened 2025-07-15
[17.0][MIG] website_sale_product_matrix: Migration to version 17.0

pilarvargas-tecnativa:
@Tecnativa TT57082

@carlos-lopez-tecnativa @CarlosRoca13 please review

PR #1089, opened 2025-07-24
[17.0][FIX] website_sale_hide_price: repair regression as part of #1079

theangryangel:
#1079 made a change to website_sale_hide_price, which appears at least in our use case to have made the module behave differently (no product page at all, rather than no price or buy options)

Fixes GH184666


pedrobaeza (approved):
Confirmed that it was a mistake. Sorry for the inconveniences.

/ocabot merge patch

theangryangel:
No need to apologise Pedro, things happen, we’re all human - ❤️

PR #1092, opened 2025-08-01
[17.0][MIG] website_sale_secondary_unit_product_matrix: Migration to version 17.0

pilarvargas-tecnativa:
@Tecnativa TT57083

@carlos-lopez-tecnativa @CarlosRoca13 please review


website_sale_secondary_unit_product_matrix/templates/product_template.xml
```diff
@@ -0,0 +1,46 @@
...
+                t-value="product_matrix.get('secondary_unit_id')"
+            />
+            <t t-set="uom_name" t-value="product_matrix.get('uom_name')" />
+            <tr t-if="secondary_units">
+                <td t-att-colspan="len(product_matrix.get('header', []))">
+                    <div class="input-group" t-if="secondary_units">
+                        <select
+                            class="custom-select o_matrix_secondary_unit"
+                            id="secondary_unit"
+                        >
```
carlos-lopez-tecnativa: I think the class should be updated to render according to Odoo styles.
Additionally, a label describing the field would be great so that users can understand its purpose.
WDYT?
In V15
image

In V17

image

carlos-lopez-tecnativa (changes requested):
Tested locally and working fine.
Since it doesn't work as expected on Runboat, I think there is an incompatibility with another module, but I'm not sure which one.
Just a few minor comments I expect can be addressed.

website_sale_secondary_unit_product_matrix/templates/product_template.xml
pilarvargas-tecnativa: Done, thanks!

image

carlos-lopez-tecnativa (approved):
LGTM

CarlosRoca13 (approved):

PR #1038, opened 2025-03-18
[IMP]website_sale_product_assortment: Apply assortments to accessory & alternative products

LuisAlejandroS:
Issue

- Previously, product assortment restrictions were not correctly applied to accessory and alternative products displayed in website carousels. This created inconsistencies in product visibility and allowed users to add restricted products to their cart from the accessories section.

Solution

- Implemented proper filtering for accessory products based on assortment restrictions
- Extended the same restriction logic to alternative products in website carousels
- Fixed cart view to prevent purchasing restricted products

Reproduction Steps (in Runboat)

1. Configure product assortments with visibility and/or sale restrictions
2. Add accessory and alternative products to a main product
3. Visit the website and observe product carousels and cart behavior

Before fix: Restricted products remained visible in carousels, and accessory products could be added to cart despite restrictions.

After fix: Products are properly filtered according to assortment rules across all website sections, ensuring consistent visibility and purchasing restrictions.

Technical Approach

- Created helper method _get_allowed_show_products in product.template to ensure consistent filtering logic
- Enhanced the existing accessory/alternative product methods to incorporate assortment restrictions:
_get_website_accessory_product
_get_website_alternative_product
_get_products_alternative_products (website snippet filter)
- Added _show_quick_add_accesory_assortments method in product.product specifically for controlling "Add to cart" button visibility in the cart view for accessory products
- Modified cart template view to use this new method, preventing addition of restricted accessory products

FL-556-4432

PR #1093, opened 2025-08-06
[16.0][IMP] website_sale_product_assortment: Implement restrictions on the cart in _cart_update and shop_payment 

SuShanVoong:
### Description

A user experience issue was identified where customers could accidentally add restricted products to their cart. This occurred in cases where the product assortment was configured as "_Avoid selling not available products_" 

Due to slow product loading times, the "_add to cart_" button could be clicked on a restricted product before the restriction was fully applied, allowing the customer to purchase it.

It was identified that a product could be added to the cart without restrictions. If restrictions were subsequently applied to that product while it was still in the cart, the payment could still be completed even though the product was now restricted.

### Cases Covered with This Improvement
This update introduces two layers of security to prevent the sale of restricted products:

1. Immediate Cart Restriction: A new security layer has been implemented to block restricted products from being added to the cart, even if the "add to cart" button is clicked due to a slow-loading product page. This ensures that only valid products can be added.

2. Checkout Restriction: An additional security check has been added to the checkout process. If a product in the cart was previously available but has since been restricted, the customer will receive an error message preventing them from completing the purchase.

### Implemented Solution
This issue was resolved by implementing assortment restrictions within the `_cart_update` method of `sale_order.py` and the `shop_payment` method of `website_sale.py`. 

The solution utilizes the `_show_quick_add_accesory_assortments` method, which was previously implemented in this pull request: https://github.com/OCA/e-commerce/pull/1038

FL-556-5722

### DIFF
--- a/website_sale_product_assortment/models/product_template.py
+++ b/website_sale_product_assortment/models/product_template.py
@@ -85,3 +85,31 @@ class ProductTemplate(models.Model):
             domain.append(allowed_product_domain)
         res["base_domain"] = domain
         return res
+
+    def _get_allowed_show_products(self, products):
+        if not products:
+            return products
+        restricted_products_dict = self.get_product_assortment_restriction_info(
+            products.ids
+        )
+        if not restricted_products_dict:
+            return products
+
+        allowed_show_products = products.filtered(
+            lambda p: p.id not in restricted_products_dict
+            or not any(
+                assortment.website_availability == "no_show"
+                for assortment in restricted_products_dict[p.id]
+            )
+        )
+        return allowed_show_products
+
+    def _get_website_accessory_product(self):
+        res = super()._get_website_accessory_product()
+        visible_products = self._get_allowed_show_products(res)
+        return visible_products
+
+    def _get_website_alternative_product(self):
+        res = super()._get_website_alternative_product()
+        visible_products = self._get_allowed_show_products(res)
+        return visible_products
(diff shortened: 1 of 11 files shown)


SuShanVoong:
Evidence:
[Grabación de pantalla desde 2025-08-06 17-05-10.webm](https://github.com/user-attachments/assets/8514961a-5244-48f9-ae85-ee9d01af8ee8)

website_sale_product_assortment/models/product_template.py
```diff
@@ -85,3 +85,31 @@ def _search_get_detail(self, website, order, options):
...
+
+    def _get_website_accessory_product(self):
+        res = super()._get_website_accessory_product()
+        visible_products = self._get_allowed_show_products(res)
+        return visible_products
+
+    def _get_website_alternative_product(self):
+        res = super()._get_website_alternative_product()
+        visible_products = self._get_allowed_show_products(res)
+        return visible_products
```
Pablocce:
```suggestion

   def _get_allowed_show_products(self, products):
      if not products:
          return products

      restricted_products_dict = self.get_product_assortment_restriction_info(products.ids)
      if not restricted_products_dict:
          return products

      return products.filtered(
          lambda p: p.id not in restricted_products_dict
          or not any(
              assortment.website_availability == "no_show"
              for assortment in restricted_products_dict[p.id]
          )
      )

  def _get_website_accessory_product(self):
      return self._get_allowed_show_products(super()._get_website_accessory_product())

  def _get_website_alternative_product(self):
      return self._get_allowed_show_products(super()._get_website_alternative_product())
``` que te parece algo asi?

SuShanVoong: Bien visto, lo añado, gracias!!

Pablocce (approved):
LGTM

PR #1095, opened 2025-08-12
[16.0][ADD] website_sale_account_financial_risk: New module

pilarvargas-tecnativa:
@Tecnativa TT57364

@pedrobaeza please review


pedrobaeza:
Check that there are 2 commits. I would say the module can be named just `website_sale_financial_risk`.

pilarvargas-tecnativa:
@pedrobaeza please, take a look

website_sale_financial_risk/templates/website_sale_templates.xml
```diff
@@ -0,0 +1,87 @@
...
+                                /></li>
+                            <li
+                                t-if="commercial_partner.risk_account_amount_unpaid_include and commercial_partner.risk_account_amount_unpaid != 0"
+                                class="list-group-item bg-transparent border-0"
+                            ><em>Other Unpaid Amount </em><span
+                                    t-field="commercial_partner.risk_account_amount_unpaid"
+                                    t-options="{'widget': 'monetary', 'display_currency': risk_currency_id}"
+                                /></li>
+                        </ul>
+                    </li>
```
carlos-lopez-tecnativa, line 62: I think that displaying this information may require more glue modules, for example with:
https://github.com/OCA/credit-control/tree/16.0/sale_financial_risk
https://github.com/OCA/credit-control/tree/16.0/account_payment_return_financial_risk

These modules add new fields to the partner and, based on this, modify the computation of the remaining credit. So, if I have some of these modules installed, the summary displayed on the website may be inconsistent.

image

website_sale_financial_risk/models/payment_provider.py
```diff
@@ -0,0 +1,26 @@
...
+        providers = super()._get_compatible_providers(
+            *args, sale_order_id=sale_order_id, website_id=website_id, **kwargs
+        )
+        order = self.env["sale.order"].browse(sale_order_id).exists()
+        if (
+            not self.env.user.credit_limit
+            or self.env.user.risk_remaining_value < order.amount_total
+        ):
+            providers = providers.filtered(lambda p: not p.code == "credit")
+        return providers
```
carlos-lopez-tecnativa, line 26: What happens when you don’t pay a `sale.order` and instead try to pay an `invoice` or generate a payment link?

I see the payment available, and from my point of view, this should be hidden.

I suggest inverting the logic: always remove the provider, but only add the provider to the list when the condition is satisfied. What do you think?

image

image

carlos-lopez-tecnativa (commented):
Code and functional review looks good to me, but I left some comments in case you consider them applicable to prevent misuse by the users.

pedrobaeza:
I'm thinking this should go to `OCA/credit-control` to have all the financial_risk suite together.

website_sale_financial_risk/templates/website_sale_templates.xml
pedrobaeza: We can have in the base module a method `_get_financial_risk_amount_fields` or similar, and iterate it through it for getting the detail, and `sale_financial_risk` or others to expand such method with the new field name.

website_sale_financial_risk/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+Set a credit limit and configure financial risk settings as described in the 
+account_financial_report module.
```
pedrobaeza, line 2: Incorrect module name.

pedrobaeza (commented):
Functional review (although it's pending to be moved to the other repo):

imagen

It's also convenient to put more info in the payment acquirer description (not the title). The credit information should be shown in better place IMO.

website_sale_financial_risk/templates/website_sale_templates.xml
pilarvargas-tecnativa: Since the view is exactly the same, you could have the view in the base module with the necessary parameters to be able to call it from any module (in this case, call it from website_sale_financial_risk) and it can be extended from any module that inherits only the base module.

pilarvargas-tecnativa:
> imagen

By default, this payment method is disabled. If it appears in test mode, it is because whoever configured the payment method has activated it in test mode.
image

> It's also convenient to put more info in the payment acquirer description (not the title). The credit information should be shown in better place IMO.

Continue in https://github.com/OCA/credit-control/pull/491

PR #1114, opened 2025-09-25
[16.0][ADD] website_sale_order_shipping_modification: New module

pilarvargas-tecnativa:
@Tecnativa TT57928

@CarlosRoca13 @juancarlosonate-tecnativa @pedrobaeza please review


website_sale_order_shipping_modification/models/website.py
```diff
@@ -0,0 +1,41 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+from odoo.http import request
+
+
+class Website(models.Model):
+    _inherit = "website"
+
+    def sale_get_order(self, force_create=False, update_pricelist=False):
```
pedrobaeza, line 11: Should you override this general method and do all that stuff here? It seems a bit overkilling, and in any case, no comments have been added saying what's going on.

pedrobaeza (changes requested):
The idea of the module is not to "modify" the current address, but to select another address into the existing ones, so please switch the behavior to this one.

website_sale_order_shipping_modification/models/website.py
pilarvargas-tecnativa: This method changes the order partner to the current user because it is designed for the flow of an order. If you want to use the existing address wizard and take advantage of the logic, you must retrieve the order partner data and restore it, as is done in the method itself, but in this case for the existing order. Otherwise, you would have to define all the new logic and new controllers to edit the order address, taking advantage only of the view. I have left the explanation documented in the method.

pedrobaeza:
You have to still change the approach to what I have mentioned.

pilarvargas-tecnativa:
> You have to still change the approach to what I have mentioned.

Done

website_sale_order_shipping_modification/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,27 @@
...
+    >
+        <xpath
+            expr="//div[@id='invoicing_shipping_address']/div/strong"
+            position="before"
+        >
+            <a
+                t-if="sale_order.partner_shipping_id == sale_order.partner_invoice_id and sale_order.state =='sent' and not sale_order.is_expired"
+                class="float-end no-decoration"
+                t-attf-href="/shop/checkout?portal_order_id=#{sale_order.id}&amp;access_token=#{access_token or ''}"
+            ><i class="fa fa-edit" /> Edit</a>
```
pedrobaeza:
```suggestion
            ><i class="fa fa-edit" /> Edit shipping address</a>
```

website_sale_order_shipping_modification/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,27 @@
...
+    <template
+        id="sale_order_portal_content"
+        inherit_id="sale.sale_order_portal_content"
+    >
+        <xpath
+            expr="//div[@id='invoicing_shipping_address']/div/strong"
+            position="before"
+        >
+            <a
+                t-if="sale_order.partner_shipping_id == sale_order.partner_invoice_id and sale_order.state =='sent' and not sale_order.is_expired"
```
pedrobaeza, line 14: Why not allowing this in `draft` state?

website_sale_order_shipping_modification/views/sale_portal_templates.xml
pilarvargas-tecnativa: Done
image

website_sale_order_shipping_modification/views/sale_portal_templates.xml
pilarvargas-tecnativa: Not shown in draft https://github.com/odoo/odoo/blob/16.0/addons/sale/controllers/portal.py#L37

website_sale_order_shipping_modification/readme/USAGE.rst
```diff
@@ -0,0 +1,8 @@
+* Go to *Quotations* in the portal and open a quotation.
+* Next to the delivery address, an **Edit** button is shown.
```
pedrobaeza: Change this to the new label and mention about only being shown in sent quotations.

pilarvargas-tecnativa: No other quotations are actually shown on the portal, but I have already added the changes to the readme file. The part about creating a new address was missing.

juancarlosonate-tecnativa (approved):
LGTM functional review OK

PR #998, opened 2024-12-18
[MIG] website_sale_product_assortment: Migration to 18.0

PieterPaulussen:

PR #975, opened 2024-10-15
[16.0][IMP] website_sale_product_assortment

Pablocce:
Description:

When more than one assortment is applied to the customer on e-commerce, the functionality doesn't work correctly, and the configurations are not applied properly, for example, when combining "no restriction" with "no show" or "no purchase."

This PR is essential for the functionality of this module because the products are not shown or cannot be bought when they should.

Cases covered with this improvement:

1. Assortment A with product 1 and configuration "Avoid to show non available products" and assortment B with product 2 and configuration "Avoid to show non available products". Before, both products were displayed with the label "Not available". With this PR both products are displayed and they are allowed to be purchased.
2. Assorment A with product 1 and configuration "Avoid to show non available products" and assortment B with product 2 and configuration "No restrictions". Before, only the products from de assortment A were displayed, restricting the products from de B assortment. With this PR, the restrictions are not applied to the assortments with the configuration "No restrictions" when simultaneous assortments are applied to the same client or website.

Solution:
To resolve this problem, we have made the following modifications:

1. The method that returns the products to display is adjusted so that, in the absence of restrictive stock, it correctly returns the values.
3. Additionally, the method that returns the products and their restrictions is adjusted to make the assortments compatible.

PR #1120, opened 2025-10-17
[18.0][MIG] website_sale_product_assortment: Migration to 18.0

CarlosRoca13:
cc @Tecnativa TT54537

This PR superseds https://github.com/OCA/e-commerce/pull/998 because author is not giving response

ping @pedrobaeza @pilarvargas-tecnativa


pilarvargas-tecnativa:
Is #975 necessary here?

PieterPaulussen:
@carlosdauden I've tested this module and found the following missing features:
- The product domains are not respected when navigating the shop. You can test this by using a demo user and limit a product assortment to it. Log in with the demo user and then validate that the exact number of products are listed. (which does not happen)
- The Visibility filter is also not applied correctly.

mariancuadranetkia:
@CarlosRoca13 can you rebase? We like to check functionality.
Thanks!

mariancuadranetkia:
@CarlosRoca13 thanks for rebase!

As @PieterPaulussen commented before, visibility in website is not applied correctly. We can see assortments defined as not visible with assortment. Only add to cart button is disabled enough we have availability on website as 'avoid to show non available products':
imagen

I'm not sure if it could be because I set FURN_0097 as restricted product which is a variant of a template with more variants... If you think so, I can test with 1-1 product_template-product_product to check if is respected this visibility with different assortments
Thanks in advance!

CarlosRoca13:
Hi! I have tested and I'm not seeing the problem, when a user tries to go to a product that is hided, it will be impossible for him, just will find it when the product has multiple variants. In this case the variant is showed, but the client will be not able to purchase that product.

mariancuadranetkia (approved):
Functional review, LGTM
Thanks!!

pilarvargas-tecnativa (approved):

PR #1119, opened 2025-10-17
[18.0][MIG] website_sale_filter_product_brand: Migration to version 18.0

eduezerouali-tecnativa:
cc @Tecnativa TT58476
@pilarvargas-tecnativa @carlos-lopez-tecnativa

PR #1121, opened 2025-10-20
[18.0][IMP] website_sale_product_brand: merged website_sale_filter_product_brand

eduezerouali-tecnativa:
cc @Tecnativa TT58476

Merged website_sale_filter_product_brand from 14.0 
This comes from https://github.com/OCA/e-commerce/pull/1119

### DIFF
--- a/website_sale_product_brand/__manifest__.py
+++ b/website_sale_product_brand/__manifest__.py
@@ -16,6 +16,7 @@
         "data/website_menu.xml",
         "views/product_brand.xml",
         "views/product_brand_views.xml",
+        "views/templates.xml",
     ],
     "demo": [
         "demo/product_brand_demo.xml",
@@ -27,6 +28,7 @@
         ],
         "web.assets_tests": [
             "/website_sale_product_brand/static/src/js/tour.esm.js",
+            "/website_sale_product_brand/static/src/js/test_website_sale_filter_brand.esm.js",
         ],
     },
     "installable": True,
--- a/website_sale_product_brand/controllers/main.py
+++ b/website_sale_product_brand/controllers/main.py
@@ -9,6 +9,68 @@ from odoo.addons.website_sale.controllers.main import QueryURL, WebsiteSale
 
 
 class WebsiteSale(WebsiteSale):
+    def _get_shop_domain(
+        self, search, category, attrib_values, search_in_description=True
+    ):
+        domain = super()._get_shop_domain(
+            search=search,
+            category=category,
+            attrib_values=attrib_values,
+            search_in_description=search_in_description,
+        )
+        # add selected brands to product search domain
+        brands_list = self._get_brand_ids(request.httprequest.args)
+        return self._update_domain(brands_list, domain)
+
+    def _update_domain(self, brands_list, domain):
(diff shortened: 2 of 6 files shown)


pilarvargas-tecnativa:
Hi @sbidoul !

I'm trying to start the runboat, but it stays in a constant “starting” state. I've also tried it on a runboat from the repo, and the same thing happens. 

image

sbidoul:
Hi. I clicked reset and it worked. No idea what happened.

pilarvargas-tecnativa:
The behavior is as expected, except that the filter bar should always display the filters related to the products currently in the catalog. In other words, if they have been filtered, they should continue to be displayed in the filter panel, as is the case with the size filter, for example:

image

I have noticed that this happens when products do not have attributes. You can try with two products that only have a brand. When you filter by brand, it disappears from the panel when you use it as a filter. If you add an attribute to one of those branded products, it will remain in the panel once the filter is applied.

carlos-lopez-tecnativa (changes requested):
The filters by brand are working, but please check my comment here https://github.com/OCA/e-commerce/pull/1119#issuecomment-3415595592 it’s not working correctly in this case

pilarvargas-tecnativa (approved):
LGTM

carlos-lopez-tecnativa (changes requested):
Please check this use case:
When I select more than one brand, the products are not displayed correctly.

image

In the backend, the products with brands appear as follows:
image

I noticed that only the first selected brand is being used, which is incorrect.

eduezerouali-tecnativa:
@carlos-lopez-tecnativa  will have a look on that, neither was working in previous verison.

pilarvargas-tecnativa:
According to the behavior of the attributes, even though this is the brand, it should not be possible to select multiple brands. When an attribute is selected, the rest are hidden. Only the attributes present in the products that appear on the screen are displayed, so the brands should behave in the same way. This is why, when you click, the event is triggered and displays the products with the first brand that was selected, ignoring clicks on the following ones.

carlos-lopez-tecnativa:
> According to the behavior of the attributes, even though this is the brand, it should not be possible to select multiple brands. When an attribute is selected, the rest are hidden. Only the attributes present in the products that appear on the screen are displayed, so the brands should behave in the same way. This is why, when you click, the event is triggered and displays the products with the first brand that was selected, ignoring clicks on the following ones.

Based on your comment, when I select Brand 1, the remaining brands should be hidden, but this does not occur.
image

Also, if I can select only a single brand, I think it’s better to use a radio button instead of a checkbox. What do you think?

pilarvargas-tecnativa:
> Also, if I can select only a single brand, I think it’s better to use a radio button instead of a checkbox. What do you think?

Well, the checkbox can be kept, but the other unused marks must be hidden.

eduezerouali-tecnativa:
@pilarvargas-tecnativa @carlos-lopez-tecnativa  changes dones. Now when you select a brand others are hidden.

carlos-lopez-tecnativa (approved):
LGTM

pilarvargas-tecnativa (approved):

PR #1047, opened 2025-04-17
[18.0][MIG] website_sale_charge_payment_fee: migration to 18.0

miguel-S73:
migration from 14.0 to 18.0.

PR #1123, opened 2025-10-22
[18.0][MIG] website_sale_charge_payment_fee: Migration to version 18.0

eduezerouali-tecnativa:
cc @Tecnativa TT58475

Migrated from https://github.com/OCA/e-commerce/pull/1047

Supersedes https://github.com/OCA/e-commerce/pull/1047


website_sale_charge_payment_fee/tests/test_website_sale_fee.py
```diff
@@ -1,73 +1,90 @@
...
-            "/shop", "website_sale_order_payment_acquirer_tour", login="admin"
-        )
-        created_order = self.env["sale.order"].search(
-            [("id", "not in", existing_orders.ids)]
-        )
-        price = 10 / 100 * 49.5
+        with RecordCapturer(self.env["sale.order"], []) as capture:
+            self.start_tour("/shop", "payment_fee_tour", login="portal", step_delay=100)
+        created_order = capture.records
+        price = 10 / 100 * 99.0
```
pilarvargas-tecnativa, line 81: Why does this value change?

eduezerouali-tecnativa: It does not change the value, actually is calculate the expected value of `created_order.amount_payment_fee`. That's why next step it and `assetEqual`.

pilarvargas-tecnativa: I mean, why does the calculation change? Before, it was done with `price = 10 / 100 * 49.5`

eduezerouali-tecnativa: Selección_001
Selección_002

change demo price form 14.0 to 18.0
16.5 * 3 = 49.5 and 33 * 3 = 99

website_sale_charge_payment_fee/models/sale_order.py
```diff
@@ -13,17 +13,20 @@ class SaleOrder(models.Model):
...
-            lambda l: not l.payment_fee_line
-        )
+        res = super()._compute_website_order_line()
+        website_order_line = self.env["sale.order.line"]
+        for order in self:
+            for line in order.website_order_line:
+                if not line.payment_fee_line:
+                    website_order_line |= line
+            order.website_order_line = website_order_line
+        return res
```
carlos-lopez-tecnativa, line 27: Why change this code? I think the older code is clearer and more compact.

website_sale_charge_payment_fee/tests/test_website_sale_fee.py
```diff
@@ -1,73 +1,90 @@
...
                 "charge_fee": True,
-                "charge_fee_product_id": self.product_product_service.id,
+                "charge_fee_product_id": cls.product_product_service.id,
                 "charge_fee_type": "percentage",
                 "charge_fee_percentage": 10.00,
+                "state": "enabled",
+                "is_published": True,
+            }
+        )
+        transfer_provider._compute_charge_fee_description()
```
carlos-lopez-tecnativa: You don’t need the `_compute_charge_fee_description`  function; it should be called automatically.

website_sale_charge_payment_fee/models/sale_order.py
```diff
@@ -0,0 +1,95 @@
...
+                .show_line_subtotals_tax_selection
+                == "tax_excluded"
+            ):
+                for line in order.order_line:
+                    if line.payment_fee_line:
+                        amount_payment_fee += line.price_subtotal
+            else:
+                for line in order.order_line:
+                    if line.payment_fee_line:
+                        amount_payment_fee += line.price_total
```
carlos-lopez-tecnativa: The same here: Why change this code? I think the older code is clearer and more compact.

website_sale_charge_payment_fee/static/src/js/website_sale_fee.esm.js
```diff
@@ -0,0 +1,33 @@
...
+            this.target
+                .closest(".oe_website_sale")
+                .querySelector('button[name="o_payment_submit_button"]').disabled =
+                true;
+            browser.location.href =
+                "/shop/payment?provider_id=" +
+                providerId +
+                "&payment_option_id=" +
+                paymentOptionId;
+        }
```
carlos-lopez-tecnativa, line 31: I see that this redirect reloads the page, and I get this message when switching between different payment providers. Is this correct? I can’t continue from this step.
image

carlos-lopez-tecnativa (changes requested):

eduezerouali-tecnativa:
@carlos-lopez-tecnativa  thanks for your review! changes you requested done. Couldn't replicate what your error. Could you give more details about that?

website_sale_charge_payment_fee/models/sale_order.py
```diff
@@ -43,36 +47,38 @@ def _compute_amount_payment_fee(self):
                 )
             else:
                 order.amount_payment_fee = sum(
-                    order.order_line.filtered("payment_fee_line").mapped("price_total")
+                    order.order_line.filtered("payment_fee_line").mapped(
+                        "price_subtotal"
```
carlos-lopez-tecnativa: Why price_subtotal??

website_sale_charge_payment_fee/models/payment_provider.py
```diff
@@ -5,22 +5,20 @@
...
         "payment method",
     )
-    charge_fee_description = fields.Text("Fee Description")
-    charge_fee_product_id = fields.Many2one("product.product", string="Fee Product")
-    charge_fee_fixed_price = fields.Float(
-        "Fixed Price", digits=dp.get_precision("Product Price")
+    charge_fee_description = fields.Text(
+        "Fee Description", compute="_compute_charge_fee_description"
     )
+    charge_fee_product_id = fields.Many2one("product.product", string="Fee Product")
```
carlos-lopez-tecnativa: I think this product should have a domain that allows selecting only records of type `service`. What do you think?

website_sale_charge_payment_fee/static/src/js/website_sale_fee.esm.js
carlos-lopez-tecnativa: I attached a video. I believe the problem occurs when you have two or more payment providers. In this case, I installed `Demo` and `Wire Transfer` and selected the second payment option. The window flickers, and internally the sale order changes when another payment method is selected. So, I think the window is not being refreshed correctly.

https://github.com/user-attachments/assets/f917a74e-4d90-48ca-aaa5-ef4d88fd2e4a

carlos-lopez-tecnativa (changes requested):

website_sale_charge_payment_fee/models/sale_order.py
eduezerouali-tecnativa: It should be `price_total`

website_sale_charge_payment_fee/models/payment_provider.py
eduezerouali-tecnativa: I agree with you. Makes sense, as I can't see a possible scenario where that fee could be a good

website_sale_charge_payment_fee/tests/test_website_sale_fee.py
carlos-lopez-tecnativa: Could you avoid depending on demo data to prevent external changes from altering the behavior here?

website_sale_charge_payment_fee/controllers/main.py
```diff
@@ -0,0 +1,43 @@
...
+from odoo.http import request
+
+from odoo.addons.payment.controllers import portal as payment_portal
+
+
+class WebsiteSaleFee(payment_portal.PaymentPortal):
+    @http.route(
+        "/shop/payment", type="http", auth="public", website=True, sitemap=False
+    )
+    def shop_payment(self, **post):
```
carlos-lopez-tecnativa, line 16: This class `PaymentPortal`, doesn’t have the `shop_payment` function in the `payment` module. This method is added in the `website_sale` module. Why did you change the class to inherit from it instead of inheriting from the `WebsiteSale` class?

carlos-lopez-tecnativa (commented):
Working fine, just a doubt about the class change. I just want to understand the change. Other than that, LGTM.

eduezerouali-tecnativa:
@carlos-lopez-tecnativa Changes done, no it does not depends on demo data for test.

Reyes4711-S73:
@eduezerouali-tecnativa I tested and it is not working well when you try to pay directly with X payment.
image
If you press the button pay with demo, it doesn't apply the fee charge.
image
Otherwise, if you continue with the checkout and you fill all the billing addres and shipping address, the fee is charged.
image
image

website_sale_charge_payment_fee/views/payment_provider_views.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="provider_form_fee_inherit" model="ir.ui.view">
+        <field name="name">payment_provider.form.fee.inherit</field>
+        <field name="model">payment.provider</field>
+        <field name="inherit_id" ref="payment.payment_provider_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page string="Charge payment fee">
```
carlos-lopez-tecnativa: > @eduezerouali-tecnativa I tested and it is not working well when you try to pay directly with X payment. image 


@Reyes4711-S73 The case you mean is when the payment provider supports `express checkout`, but in this case, it’s out of scope for this module.
@eduezerouali-tecnativa You can add it to the `ROADMAP` and hide the page in this case.
I don’t know which payment providers support `express checkout` apart from Demo, which is never used in the real world.

image

image


```suggestion
                <page string="Charge payment fee" invisible="allow_express_checkout">
```

website_sale_charge_payment_fee/tests/test_website_sale_fee.py
```diff
@@ -0,0 +1,117 @@
...
+            {
+                "name": "Discount wire tranfer",
+                "standard_price": 70.0,
+                "list_price": 79.0,
+                "sale_ok": True,
+            }
+        )
+        cls.product_product_optional = cls.env["product.template"].create(
+            {
+                "name": "Oprional",
```
carlos-lopez-tecnativa:
```suggestion
                "name": "Optional",
```

website_sale_charge_payment_fee/tests/test_website_sale_fee.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        cls.product_product_buy = cls.env["product.template"].create(
+            {
+                "name": "Test-1",
+                "standard_price": 33.0,
+                "list_price": 33.0,
+                "is_published": True,
+                "sale_ok": True,
+                "taxes_id": [Command.clear()],
+                "optional_product_ids": [Command.set(cls.product_product_optional.ids)],
```
carlos-lopez-tecnativa: Is the optional product really necessary? Why?

eduezerouali-tecnativa: not really necessary,  i did that,just to not adjust the tour. Anyways i did adjust the tour. :)

website_sale_charge_payment_fee/views/payment_provider_views.xml
eduezerouali-tecnativa: I implemented your suggestions, anyways i think that stripe used it too. Also add it to roadmap. Thanks :)

carlos-lopez-tecnativa (approved):
LGTM

website_sale_charge_payment_fee/static/src/js/website_sale_fee.esm.js
Alexgars73: @carlos-lopez-tecnativa @eduezerouali-tecnativa we added this function in the migration to resolve the problem, could you add and test, thanks: 
_prepareTransactionRouteParams() {
        const transactionRouteParams = this._super(...arguments);
        // eslint-disable-next-line no-undef
        const cartValueEl = document.querySelector(".oe_currency_value");
        const cartValue = cartValueEl?.textContent || "0";
        transactionRouteParams.amount = Number(cartValue);
        return transactionRouteParams;
    },

eduezerouali-tecnativa: @Alexgars73 That problem was already solve or you are talking about express checkout?

Alexgars73: @eduezerouali-tecnativa i'm getting the same error that @carlos-lopez-tecnativa mentioned "The cart has been updated. Please refresh the page".

eduezerouali-tecnativa: @Alexgars73 try on runboat, it is working fine.
Selección_012
Selección_013

eduezerouali-tecnativa:
@pilarvargas-tecnativa please could you review

ivantodorovich (approved):

PR #1084, opened 2025-07-18
[14.0][ADD] website_sale_acquirer_confirm_order: New module

pilarvargas-tecnativa:
@Tecnativa TT57241

@CarlosRoca13 @carlosdauden please review

PR #1127, opened 2025-10-27
[18.0][MIG] website_sale_acquirer_confirm_order: Migration to version 18.0

pilarvargas-tecnativa:
Migrated from https://github.com/OCA/e-commerce/pull/1084

@Tecnativa TT58470

@carlos-lopez-tecnativa @victoralmau please review


website_sale_acquirer_confirm_order/views/payment_provider_views.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="payment_provider_form" model="ir.ui.view">
+        <field name="name">provider.form.inherit.website</field>
+        <field name="model">payment.provider</field>
+        <field name="inherit_id" ref="payment.payment_provider_form" />
+        <field name="arch" type="xml">
+            <field name="allow_express_checkout" position="after">
+                <field name="confirm_order" />
```
carlos-lopez-tecnativa: Functionally tested and working, but I think a `CONTEXT.md` file is necessary to explain in which cases this module can be used.

Every module `payment_`**** that adds the option to pay on the website and integrates with external providers must be in charge of handling when the transaction is approved, declined, or canceled. So, the common case is that these modules, in any way, end up confirming the order.
Therefore, the only module I can see that could use this one is when using a `Wire Transfer`. If that’s the case, this field should be visible only in those cases.

pilarvargas-tecnativa: done, thanks!

carlos-lopez-tecnativa (approved):
LGTM

carlos-lopez-tecnativa (changes requested):
Wait, when I use a payment link, the order is not confirmed, it just stays in the Sent state. Please review this case.
image

carlos-lopez-tecnativa (approved):
> Wait, when I use a payment link, the order is not confirmed, it just stays in the Sent state. Please review this case.

According to the internal chat, this module is intended for e-commerce, and a sales order from the backend is out of scope for this module. So, LGTM.

victoralmau (approved):

PR #1130, opened 2025-10-29
[18.0][MIG] website_sale_cart_expire: Migration to version 18.0

pilarvargas-tecnativa:
This migration only ports the cart expiration logic to v18 and deliberately
omits the frontend countdown timer in the cart.

The countdown timer has not been migrated because there is no underlying
reservation mechanism implemented in this module. Without a functional
reservation of seats, resources, or stock at quotation time, displaying a
timer in the cart would provide no real effect beyond visual urgency.

A countdown would only make sense in scenarios where the draft order
actually holds a temporary reservation (e.g., event seat allocation,
booking/rental slots, or other custom reservation flows). Since none of
those use cases are part of the current module and cannot be properly
validated here, the timer has been intentionally left out of the v18
migration.

The functional value of the module in standard deployments comes from the
automatic cancellation of stale carts through the cron job, which remains
fully supported and migrated. This keeps the module lean, predictable, and
focused on its core purpose.

The structure of the module remains open for future contributions, so
anyone needing a countdown UI in combination with a reservation mechanism
can reintroduce or adapt that part as required for their specific use case.

@Tecnativa TT58474

@carlos-lopez-tecnativa @CarlosRoca13  please review


website_sale_cart_expire/views/templates.xml
```diff
@@ -0,0 +1,23 @@
...
+-->
+<odoo>
+    <template
+        id="header_cart_link"
+        inherit_id="website_sale.header_cart_link"
+        name="Cart Expiration Timer"
+        active="False"
+        customize_show="True"
+    >
+        <xpath expr="//sup[hasclass('my_cart_quantity')]" position="after">
```
carlos-lopez-tecnativa: image

website_sale_cart_expire/readme/CONFIGURE.md
```diff
@@ -0,0 +1,4 @@
+Go to Website \> Settings and set a delay for Expire Carts settings.
+
+A cart expiration timer can be displayed on the website by enabling the
+Cart Expiration Timer setting in the Customize menu.
```
carlos-lopez-tecnativa: This can be updated because the `Customize` as Menu no longer exists.

carlos-lopez-tecnativa (changes requested):

website_sale_cart_expire/readme/CONFIGURE.md
pilarvargas-tecnativa: All done, thanks

website_sale_cart_expire/static/src/js/website_sale_cart_expire.esm.js
```diff
@@ -0,0 +1,129 @@
...
+        const remainingStr = `${minutes}:${seconds}`;
+        if (remainingStr !== this.$el.text()) {
+            this.$el.text(remainingStr);
+        }
+    },
+    /**
+     * Updates the expiration date by reading from backend
+     */
+    _refreshExpirationDate: async function () {
+        const expireDate = await this._rpc({route: "/shop/cart/get_expire_date"});
```
carlos-lopez-tecnativa: When the timer reaches zero, I get this error.
image

carlos-lopez-tecnativa (changes requested):

website_sale_cart_expire/static/src/js/website_sale_cart_expire.esm.js
```diff
@@ -0,0 +1,131 @@
...
+            this._stopTimer();
+            this._refreshExpirationDate();
+        }
+    },
+    /**
+     * Updates the remaining time on the dom
+     */
+    _renderTimer: function (remainingMs) {
+        const remainingMsRounded = Math.ceil(remainingMs / 1000) * 1000;
+        // Don't show the timer if remaining time is less than 1 hour
```
carlos-lopez-tecnativa:
```suggestion
        // Don't show the timer if remaining time is great than 1 hour
```

carlos-lopez-tecnativa (approved):
LGTM, just a minor non-blocking comment.
I think it would be great to add an option to notify the user when the cart has expired or is about to expire, because I waited for the time to run out but, as a user, I didn’t receive any notification. Also, while the cron hasn’t been executed yet, my cart remains unchanged, but once the cron runs, I only see my cart empty.

carlos-lopez-tecnativa (approved):

pilarvargas-tecnativa:
@eduezerouali-tecnativa please review

eduezerouali-tecnativa (approved):
Code review + runboat

PR #1131, opened 2025-11-03
[18.0][MIG] website_sale_barcode_search: Migration to version 18.0

pilarvargas-tecnativa:
@Tecnativa TT58473

@carlos-lopez-tecnativa @carlosdauden please review

In this migration, the search function has been adapted to the current
product search function in any search engine on the website.


carlos-lopez-tecnativa (changes requested):
When I tried to test and search, I got this error. Please review

image

I think this code needs to be adapted to use something like `product_variant_ids.barcode`.
https://github.com/odoo/odoo/blob/54f10b56f577ad9ed5575bd396dba7d20d22fc2e/addons/website_sale/models/product_template.py#L776

pilarvargas-tecnativa:
> When I tried to test and search, I got this error. Please review

I don't replicate that error locally. It's a calculated field, but when there are no variants, the barcode goes in the product template. Otherwise, it's calculated from the variants, but as I said, I don't replicate it. I've pushed again to restart runboat in case it was a runbot error, and I haven't replicated it either.
My tests on a clean database are:
- Search for a random code without setting a barcode for any product. The result is that it does not find any products, which is expected.
- Set a barcode for a product template without variants and search. It finds the product.
- In the previous template, delete the barcode and set it on the only variant. It finds the product.
- Set barcodes on two variants of the same product. It finds the product.
- Search for any other code and it finds nothing, as expected.

Is there a specific test where the error appears?

carlos-lopez-tecnativa:
> > When I tried to test and search, I got this error. Please review
> 
> I don't replicate that error locally. It's a calculated field, but when there are no variants, the barcode goes in the product template. Otherwise, it's calculated from the variants, but as I said, I don't replicate it. I've pushed again to restart runboat in case it was a runbot error, and I haven't replicated it either. My tests on a clean database are:
> 
> * Search for a random code without setting a barcode for any product. The result is that it does not find any products, which is expected.
> * Set a barcode for a product template without variants and search. It finds the product.
> * In the previous template, delete the barcode and set it on the only variant. It finds the product.
> * Set barcodes on two variants of the same product. It finds the product.
> * Search for any other code and it finds nothing, as expected.
> 
> Is there a specific test where the error appears?

I just type any word in the search panel. I noticed that when I type more than three words, this error is raised. It’s weird, but you can reproduce it in Runboat. Looking deeper, I imagine this is related to PostgreSQL’s trigram index:
https://www.postgresql.org/docs/current/pgtrgm.html


https://github.com/user-attachments/assets/a2bc3421-5653-42bb-9313-39e93348069b

carlos-lopez-tecnativa (approved):
LGTM

carlosdauden (approved):
It may not be the ideal module name and repository for this functionality as it stands after migration to this version, but I don't think it's important enough to prevent merging the PR.

PR #1137, opened 2025-11-24
[18.0][IMP] website_sale_product_attribute_value_filter_existing: improve performance using ID sets

pilarvargas-tecnativa:
The overhead of extending the entire "shop()" method is eliminated and the logic is moved to "_get_additional_extra_shop_values()", the hook provided by Odoo to add data to the context without duplicating core code.

In addition, the filtering of unused values is optimized by avoiding the intersection of recordsets in QWeb ("attr_values_used & v"). Instead, the IDs of used values are precomputed as a set ("attr_values_used_ids") and the templates check "v.id in attr_values_used_ids", which drastically reduces the cost per iteration.

image

@Tecnativa


carlosdauden (approved):

sergio-teruel (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1138, opened 2025-11-24
[18.0][MIG] website_sale_order_shipping_modification: Migration to version 18.0

pilarvargas-tecnativa:
@tecnativa TT59372

@juancarlosonate-tecnativa @victoralmau please review


website_sale_order_shipping_modification/controllers/main.py
```diff
@@ -0,0 +1,43 @@
...
+    def _check_cart(self, order_sudo):
+        portal_order_id = request.session.get("portal_order_id", False)
+        portal_access_token = request.session.get("portal_access_token", False)
+        sale_order_id = request.session.get("sale_order_id", False)
+        if portal_order_id and portal_access_token and sale_order_id:
+            return
+        return super()._check_cart(order_sudo=order_sudo)
+
+    @http.route()
+    def shop_address(self, **kw):
```
eduezerouali-tecnativa: In `shop_address` there are parameters missing. Please could you check?
 https://github.com/OCA/OCB/blob/4899174a7b4adde72c05b5590dce3497473da7f8/addons/website_sale/controllers/main.py#L1112

eduezerouali-tecnativa (changes requested):

website_sale_order_shipping_modification/controllers/main.py
pilarvargas-tecnativa: thanks!

victoralmau (approved):
Functional review OK

Would it be possible for the Discard button to return to the quotation?

pilarvargas-tecnativa:
> Functional review OK
> 
> Would it be possible for the Discard button to return to the quotation?

Where does the discard button appear? In the new address view? Can you mark it as required changes until I review it, please?

victoralmau:
> > Functional review OK
> > Would it be possible for the Discard button to return to the quotation?
> 
> Where does the discard button appear? In the new address view? Can you mark it as required changes until I review it, please?

ejemplo

pilarvargas-tecnativa:
> Would it be possible for the Discard button to return to the quotation?

I've been checking because I thought the button to discard changes led to a link that deviated from this logic, but it actually behaves as expected. When you discard adding a new address, the normal thing is to return to the place where you decided to add it, i.e., the address panel. This is the basic behavior of Odoo and is the behavior that was already present in v16. I believe that, for the moment, it is not necessary to further modify Odoo's behavior beyond the purpose of this module, which is to be able to select or add another delivery address.

eduezerouali-tecnativa (approved):
LGTM

PR #1144, opened 2025-12-03
[18.0][IMP] website_sale_checkout_skip_payment: Fix duplicate order name display + allows to confirm or not the order

rousseldenis:
image


rousseldenis:
@pilarvargas-tecnativa You changed the way this module behaves there : https://github.com/OCA/e-commerce/commit/38d6f684e7ea2e8095127be4254cabab097133b4

The order is confirmed now. But a mail as 'Quotation' is still sent. Which is not good.

My plan is to add a configuration (global) parameter :

- Either we skip payment and set to quotation sent
- Either we skip payment and we confirm the order

What do you think ?

rousseldenis:
A good improvement should be also :

- To store the 'skip_payment' information on sale order level (if this changes on partner level)
- To use a specific mail template

pilarvargas-tecnativa:
> @pilarvargas-tecnativa You changed the way this module behaves there : [38d6f68](https://github.com/OCA/e-commerce/commit/38d6f684e7ea2e8095127be4254cabab097133b4)

Well, I didn't really change any behavior, I just recovered something that was already in 15.0.

https://github.com/OCA/e-commerce/commit/0a0bfc2035d42815de364d884289425a222cfc9f

> The order is confirmed now. But a mail as 'Quotation' is still sent. Which is not good.
> 
> My plan is to add a configuration (global) parameter :
> 
> * Either we skip payment and set to quotation sent
> * Either we skip payment and we confirm the order
> 
> What do you think ?

The purpose of this module is aimed at B2B customers who are allowed to pay later but whose order must be confirmed and who receive that email. If the problem is that a quote email is sent, perhaps the solution is to send the email after confirming the order.

The purpose of this module is clear: if the partner has it configured, skip payment, confirm the order, and receive an email as feedback for that order.

rousseldenis:
@OCA/e-commerce-maintainers Could you reopen this ? Thanks

PR #1160, opened 2026-02-10
[18.0][ADD] website_sale_checkout_address_no_save: add new module

eduezerouali-tecnativa:
cc @Tecnativa TT59949
This modulo allows to not store the address on checkout. Once transaction is confirm it will archive the contact. It is meant for customers that are dropshippers and have lots of addresses. 
ping @juancarlosonate-tecnativa @sergio-teruel 

pending:

- [x] add test


pedrobaeza:
I don't know the exact implementation of this, but the name doesn't seem the best. A more accurate one may be `website_sale_checkout_address_no_save` or maybe `website_sale_checkout_temp_address`, but as said, it depends on the implementation.

@eduezerouali-tecnativa please recheck all the sentences, as they contain incorrect English and even your name is not correctly spelled. It's good to read things twice in these cases.

eduezerouali-tecnativa:
@pedrobaeza you are right. I changed the name to `website_sale_checkout_address_no_save`; it seems more accurate.
This module is intended to avoid having too many addresses during checkout. For example, if my customer serves other customers, they might not need to have that address saved, as it may not be a regular customer. In that case, when editing the address, they can choose to archive it after the checkout is completed.

Selección_073

Additionally, when a user enters the details for another address, if the phone, email, ZIP code, and street match an archived contact, the module reactivates that contact instead of creating a new one. Al thought I do not think this will happened very often.

I did add the test but the tour to work depends on this PR https://github.com/OCA/e-commerce/pull/1162.

sergio-teruel (approved):

PR #1165, opened 2026-02-16
[19.0][ADD] website_sale_ribbon

etobella:
Odoo defines product.ribbon, but the model and field are hidden to common users from backend

The only way to configure this kind of fields is using debug mode or by the website editor.

With this module, we allow other users to configure this field from backend.


pedrobaeza:
AFAIK, this is handled from website editor.

etobella:
Yes, it is accessible from the products view. However, this change was applied in odoo 19 to simplify the work:

https://github.com/odoo/odoo/commit/b2c5e4ef0b0254549a4ddde21379a6ba1a008014

For giving context of how it is done right now.

image

luisDIXMIT (approved):
Tested on runboat and LGTM!

pedrobaeza:
OK, I see. I sincerely put to everyone `base_technical_features` and enable for all the users, as this is not the only field you are going to find hidden because they say "it's technical", and doing all these modules for just removing the technical group is too much IMO.

etobella:
Well, I understand your position, but it has no sense to me for safety reasons. It is true, that users could handle the field in the standard odoo way, and as Odoo changed the behavior un odoo 19, I thought it was a good idea to add the same features in previous versions.

pedrobaeza:
OK, as you prefer. That module is totally safe and it's not disclosing any field that compromises the security. On contrary, it prevents to start debug mode for showing business fields and having the debug dropdown with risky options.

PR #1166, opened 2026-02-19
[16.0][MIG]:website_sale_restrict_by_pricelist

SiscoIngeos:
Standard migration to v16


marcos-mendez (changes requested):
**Automated Review -- Tests Failed**

### 1. Root Cause

The test failure is caused by a missing `website_sale_restrict_by_pricelist` module in the `website_sale` dependency list, which prevents the pricelist restriction logic from being properly initialized during tests. This leads to an `AttributeError` when trying to access restricted product data.

### 2. Suggested Fix

Add `website_sale_restrict_by_pricelist` to the `depends` list in `website_sale/__manifest__.py`:

```python
# In website_sale/__manifest__.py
'depends': [
    'website_sale_restrict_by_pricelist',  # Add this line
    # ... other dependencies
],
```

### 3. Additional Code Issues

- **Missing module dependency**: The `website_sale_restrict_by_pricelist` module is not declared as a dependency in `website_sale`, causing runtime errors when its functionality is expected.
- **Incomplete test coverage**: Tests do not validate the interaction between `website_sale` and `website_sale_restrict_by_pricelist` modules, particularly around pricelist-based product visibility logic.

### 4. Test Improvements

Add the following test cases to ensure proper integration:

1. **TransactionCase for pricelist restriction logic**:
   - Create a test case inheriting from `TransactionCase`
   - Test product visibility based on pricelist restrictions
   - Validate that products are hidden when user's pricelist doesn't allow access

2. **SavepointCase for data integrity**:
   - Test data cleanup after pricelist restriction logic execution
   - Validate that database state remains consistent

3. **Tagged tests for pricelist functionality**:
   - Add `@tag('pricelist')` to tests covering pricelist-based product access
   - Include tests for both allowed and restricted product scenarios

Example test structure:
```python
from odoo.tests import TransactionCase, tagged

@tagged('pricelist')
class TestWebsiteSaleRestrictByPricelist(TransactionCase):
    def test_product_visibility_by_pricelist(self):
        # Test logic here
        pass
```

---
🔍 No reviews yet. PSC members: this PR could use your attention.

> _Every ignored PR is a contributor who might not come back. Review time matters. ([OCA Aging Report](https://git.pop.coop/pop/oca-reviewer))_

---
### Reciprocal Review Request



Hi everyone! I found some test failures on this PR and left detailed feedback above. I am happy to discuss or help debug. In the meantime, if any of you get a chance, I would appreciate a look at my open PR(s):

**My open PRs across OCA:**
- [server-tools#3554 [MIG] datetime_formatter: Migration to 18.0](https://github.com/OCA/server-tools/pull/3554)
- [server-tools#3548 [18.0][MIG] base_kanban_stage: Migration to 18.0](https://github.com/OCA/server-tools/pull/3548)
- [hr-attendance#262 [16.0][ADD] Hr_attendance_idsecure: iDSecure (ControliD) attendance integration](https://github.com/OCA/hr-attendance/pull/262)
- [stock-logistics-workflow#2276 [16.0][ADD] stock_move_line_devaluation](https://github.com/OCA/stock-logistics-workflow/pull/2276)
- [stock-logistics-workflow#2275 [16.0][ADD] Stock move line analytic account](https://github.com/OCA/stock-logistics-workflow/pull/2275)
- [stock-logistics-workflow#2268 [16.0][ADD] stock_move_line_picking_partner](https://github.com/OCA/stock-logistics-workflow/pull/2268)
- [purchase-workflow#2694 [16.0][IMP]Purchase workflow added to review state & exception fix](https://github.com/OCA/purchase-workflow/pull/2694)

Reviewing each other's work helps the whole community move forward. Thank you!

---
_Environment via [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer): Minikube + K8s Job + oca-ci/py3.10-odoo16.0 | Odoo 16.0_
_Automated review by [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer) + qwen3-coder:30b_

SiscoIngeos:
Hi @marcos-mendez , thanks for the feedback. Regarding the depends list in the manifest, I believe the current order is correct. My module extends features from website_sale.
If I move it as you suggested, the inheritance will fail or the Odoo ORM won't find the parent fields. Could you clarify if there is a specific OCA pattern you are referring to for this case?

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2025 APSL-Nagarro Antoni Marroig
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import http
+from odoo.http import request
+from odoo.tools.func import lazy
+
+from odoo.addons.http_routing.models.ir_http import slug
+from odoo.addons.website_sale.controllers.main import TableCompute, WebsiteSale
```
marcos-mendez, line 9: Suggestion: this import (and its usage on line 79) would not be needed if the controller used `_get_search_domain()` instead of post-processing. See general comment below.

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class WebsiteSaleRestrictPartnerProductController(WebsiteSale):
```
marcos-mendez, line 12: The class name still references the old module name (`RestrictPartnerProduct`). Should be aligned with the current module name, e.g. `WebsiteSaleRestrictByPricelistController`.

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
...
+    def _get_portal_restricted_products(self, search_product, product_count):
+
+        restrict_products_for_all_users = (
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param(
+                "website_sale_restrict_by_pricelist.restrict_products",
+                default="",
+            )
```
marcos-mendez, line 21: Bug: `get_param()` returns a string. When the `config_parameter` stores "True"/"False", 
`bool("False")` evaluates to `True` (line 24). Once a user enables and disables this setting, 
the restriction stays permanently active.

Fix: use `odoo.tools.str2bool()` and change `default=""` to `default="False"`.

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+        if request.env.user.has_group("base.group_portal") and (
+            request.env.user.restrict_products or bool(restrict_products_for_all_users)
```
marcos-mendez, line 24: See comment on line 15 — `bool()` on a string is always `True` for any non-empty value, 
including the string "False".

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
...
+        ):
+            pricelist = request.env.user.property_product_pricelist
+            if pricelist:
+                products = pricelist.item_ids.mapped("product_tmpl_id")
```
marcos-mendez, line 28: Bug: this only resolves items where `applied_on == '1_product'` (i.e. items with a direct 
`product_tmpl_id`). Pricelist items with `applied_on` values '3_global' (all products), 
'2_product_category', or '0_product_variant' have no `product_tmpl_id` set — they will be 
silently ignored, resulting in zero visible products for the most common pricelist configurations.

All four `applied_on` types need to be handled.

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
...
+                products = products.filtered(lambda p: p.is_published)
```
marcos-mendez, line 29: Performance: `filtered()` loads all product templates into memory, then filters in Python. 
For large catalogs this is wasteful. The `is_published` check is already part of the base 
`website_sale` search domain, so if you use `_get_search_domain()` this becomes unnecessary.

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
...
+            else:
+                products = request.env["product.template"].browse()
+            return products, len(products)
+        return (
+            search_product,
+            product_count,
+        )
+
+    @http.route()
```
marcos-mendez, line 38: Architecture: overriding `shop()` to call `super()` and then redo pagination, bins, and price 
computation is fragile and conflicts with any other module that also extends `shop()`.

The correct hook in Odoo 16.0 is `_get_search_domain()`. Adding an 
`("id", "in", allowed_product_ids)` leaf to the domain preserves all existing search, category, 
price range filtering, pagination, and pricing logic — and reduces this entire method to zero lines.

This also fixes the bug where search/category/price filters are silently discarded (the current 
`_get_portal_restricted_products` replaces `search_product` with all pricelist products regardless 
of what the user searched for).

website_sale_restrict_by_pricelist/controllers/main.py
```diff
@@ -0,0 +1,97 @@
...
+            # Get all filtered products (search_product contains the filtered results)
+            filtered_products = res.qcontext["search_product"]
+            # Recalculate prices only for the filtered products
+            products_prices = lazy(
+                lambda: filtered_products._get_sales_prices(pricelist)
+            )
+            # Update the function to use the recalculated prices
+            res.qcontext["get_product_prices"] = lambda product: lazy(
+                lambda: products_prices[product.id]
+            )
```
marcos-mendez, line 95: This entire block exists only because `shop()` is being post-processed. With `_get_search_domain()`, 
the price computation from the parent `shop()` would already be correct, making this unnecessary.

website_sale_restrict_by_pricelist/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    restrict_products_for_all_users = fields.Boolean(
+        help="Enable to restrict products base on the assigned pricelist of all users.",
```
marcos-mendez, line 11: Typo: "base on" → "based on"

website_sale_restrict_by_pricelist/models/res_partner.py
```diff
@@ -0,0 +1,27 @@
...
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    restrict_products = fields.Boolean(
+        help="Enable to restrict products base on the assigned pricelist."
+    )
+    restrict_products_for_all_users = fields.Boolean(
+        compute="_compute_restrict_products_for_all_users", store=False
+    )
+
```
marcos-mendez, line 16: Design issue: this is a global config parameter projected as a computed field on every 
res.partner record. Since the value is the same for all partners regardless of context, 
this field has no reason to live on res.partner — it adds overhead on every partner read 
for no benefit. The controller already reads `ir.config_parameter` directly, so this 
field appears to only serve the view (to show/hide the per-partner checkbox?). 

If that's the case, consider using `attrs` with a `res.config.settings` field or simply 
always showing the per-partner checkbox.Design issue: this is a global config parameter projected as a computed field on every 
res.partner record. Since the value is the same for all partners regardless of context, 
this field has no reason to live on res.partner — it adds overhead on every partner read 
for no benefit. The controller already reads `ir.config_parameter` directly, so this 
field appears to only serve the view (to show/hide the per-partner checkbox?). 

If that's the case, consider using `attrs` with a `res.config.settings` field or simply 
always showing the per-partner checkbox.

website_sale_restrict_by_pricelist/models/res_partner.py
```diff
@@ -0,0 +1,27 @@
...
+    @api.depends_context("uid")
+    def _compute_restrict_products_for_all_users(self):
```
marcos-mendez, line 18: The computed value does not depend on `uid` — it reads a global `ir.config_parameter` 
that is the same for all users. `depends_context("uid")` forces unnecessary recomputation 
on every user switch without actually reflecting any user-dependent logic.

marcos-mendez:
@etobella this is not fake you must consider in you statistic all my original reviews. Not just a shadow ban for everything.

Everything made by AI was tagged as is

etobella:
@marcos-mendez the problem of tagging inside a comment as you did is that it relies in regex or filtering of a big body, so it hard to detect. Also, you do it one way, other people does it in another way... quite complicated to make an automation here.... IMO, all AI reviews should be done with specific users and that is what I will expect in the future, mainly because it was not done by you directly, it was a bot and github does not split it correctly, the ony way would be to have a MarcosMendezBot. Actually, It is not a big deal to have several github users, and it makes sense to split it this way (in a similar way, it was done in OCA for OCA-git-bot, OCA-translate-bot ....

I don't know if I will have the time to review how to split your original work from your AI work in the future review of statistics. I can try at some point

marcos-mendez:
PR Closed by inactivity right? The bot wasn't so dumb as was said

PR #1172, opened 2026-02-26
[19.0][ADD] website_sale_product_multiple_qty

yankinmax:
This module extends the eCommerce flow to support **Sales Multiples**
(packaging quantities) directly on the product page, in the cart,
and in the product configurator.

When a product (or variant) has a *Sales Multiple* configured,
the quantity entered by the customer on the website is automatically
rounded to a valid multiple according to the interaction type.

The rounding logic is applied dynamically when the customer:

- Opens the product page
- Changes the product variant
- Clicks the "+" (increase) button
- Clicks the "–" (decrease) button
- Manually enters a quantity
- Presses **Enter** inside the quantity input
- Changes quantities in the cart


yankinmax:
This PR is moved from OCA/sale-workflow:
- [ ] https://github.com/OCA/sale-workflow/pull/4165


Hello @alexey-pelykh, I address your suggestions on the closed PR here.
Thanks again

ivantodorovich (approved):
Thanks @yankinmax !

vvrossem (approved):
Great work. 
Note: I overlooked the JS part.

ivantodorovich:
**Depends on:**
- https://github.com/OCA/sale-workflow/pull/4143

BhaveshHeliconia (approved):
LGTM!

alexey-pelykh (commented):
Thanks for moving this over and working through the feedback from OCA/sale-workflow#4165, @yankinmax — I went back through each of my earlier points against the current diff:

- **`_get_sale_multiple_vals` duplication** — resolved. `controllers/product_configurator.py` now calls `self.env["product.template"]._get_sale_multiple_vals(...)` instead of carrying its own copy, so there's a single source of truth.
- **Tests** — resolved. `tests/test_website_sale_product_multiple_qty.py` plus the two tours (`website_shop_product_multiple_qty` / `website_shop_checkout_product_multiple_qty`) cover page load, +/- buttons and manual-input rounding.
- **`int(multiple_uom.factor)` as the step** — this was my main open question, and it's correct for 19.0. The UoM refactor made `factor` the *absolute* quantity (`uom.uom._compute_factor`: `relative_factor * relative_uom_id.factor`), so a "Pack of 5" yields `factor == 5` and `int(multiple_uom.factor)` is the right step. My original concern was against the pre-18.0 semantics where `factor` was the inverse ratio (`0.002` for a pack of 500) — that no longer applies here.
- **Rounding epsilon** — good: `step * 1e-9` is now scaled to the step and carries an explanatory comment.
- **`_get_additionnal_combination_info` spelling** — matches the upstream `super()` name, so nothing to change.

The only outstanding item is the one you already flagged: `Detect unreleased dependencies` is red because `test-requirements.txt` pins `sale_product_multiple_qty` to `refs/pull/4143/head`. That clears once OCA/sale-workflow#4143 lands — nothing to do on your side until then.

Nice work.

Co-Reviewed-By: Claude Opus 4.8 <noreply@anthropic.com>

PR #1180, opened 2026-03-12
[19.0][MIG] website_sale_cart_expire: Migration to 19.0

yankinmax:


yankinmax:
Hello @ivantodorovich can you pls take a look?

website_sale_cart_expire/models/website.py
```diff
@@ -0,0 +1,63 @@
...
+        self.ensure_one()
+        expire_date = fields.Datetime.now() - timedelta(hours=self.cart_expire_delay)
+        return (
+            Domain("website_id", "=", self.id)
+            & Domain("state", "=", "draft")
+            & Domain("write_date", "<=", expire_date)
+            # We don't want to cancel carts that are already in payment.
+            & (
+                Domain("transaction_ids", "=", False)
+                | Domain(
```
ivantodorovich: This leaf is probably useless now with the new "not any" operator.

Essentially we want to ignore orders with transactions in any of those states. Not any should return true when no transactions, I guess.

To be checked

```suggestion
            & Domain(
```

ivantodorovich (approved):
Thanks @yankinmax !

website_sale_cart_expire/models/website.py
yankinmax: I tend to agree with you, but it's not explicitly written in the Odoo docs that any/not any handle the empty-relation case.
Let's keep it like this.
To be checked by functional reviews.

ivantodorovich: It should be easy to verify with a shell though. Can you check?

yankinmax: The result is the same with or without that leaf.

ivantodorovich: Excellent, thanks!

vvrossem (approved):
Migration commit LG

imlopes (approved):

BhaveshHeliconia (approved):

PR #1188, opened 2026-03-27
[17.0][IMP] website_snippet_product_category: add website-specific snippet categories

pilarvargas-tecnativa:
@Tecnativa TT61624


website_snippet_product_category/hooks.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Tecnativa - Pilar Vargas
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+def post_init_hook(env):
+    env.cr.execute(
+        """
+        INSERT INTO product_public_category_snippet_website_rel (
```
pedrobaeza: I think this is not needed, as the restriction of the categories is already handled in the website mixin.

website_snippet_product_category/templates/snippets.xml
```diff
@@ -23,31 +23,40 @@
...
-                        class="categ_link w-100"
-                        t-field="category.name"
-                        t-att-title="category.name"
-                    />
-                </div>
-            </div>
             <t
-                t-call="website_snippet_product_category.s_product_category_items_recursive"
+                t-if="
+                    category.published_in_product_category_snippet
```
pedrobaeza: Why not handle this in the `t-set` sentence for only having in `categories` the available ones?

pilarvargas-tecnativa: Done, thanks!

website_snippet_product_category/templates/snippets.xml
```diff
@@ -18,36 +18,43 @@
...
-                    />
-                </div>
-            </div>
+        <t t-if="categories and cur_level &lt;= 4">
             <t
-                t-call="website_snippet_product_category.s_product_category_items_recursive"
-            >
-                <t t-set="categories" t-value="category.child_id" />
-                <t t-set="cur_level" t-value="cur_level+1" />
+                t-set="categories"
```
pedrobaeza: But it should be at this level, or in https://github.com/Tecnativa/e-commerce/blob/e58e564bcfca968d90177939a9d7952d3bac3b6c/website_snippet_product_category/templates/snippets.xml#L55 ?

pilarvargas-tecnativa: You are right, it is better at that level. Root categories are already filtered in the controller, so handling child categories in that t-set is enough and keeps the template cleaner. Thanks

pedrobaeza (approved):

victoralmau (approved):

PR #1191, opened 2026-03-31
[17.0][FIX] website_sale_hide_empty_categories: respect product website visibility

pilarvargas-tecnativa:
The current behavior does not take the current website into account when deciding whether a category should be considered non-empty.

Because of this, in multi-website setups a category may still appear on a website even when none of its products are published there, simply because it contains products published on another website.

The expected behavior is that categories should only be considered non-empty on a website if they contain products published on that specific website.

@Tecnativa TT61762


victoralmau (approved):

pedrobaeza (approved):
Code review

/ocabot merge patch

PR #1194, opened 2026-04-11
[16.0][ADD] website_sale_digital_improved

hbrunn:
This module extends the functionality of website_sale_digital:

-  allow downloading digital products without portal account
-  allow attaching digital products to the invoice mail

Usage
=====

Access to digital products for customers without a portal account works
out of the box. Note that the links to digital products are designed to
be valid for an hour, so in case you receive complaints from your
customers that they don't work, ask them to refresh the page showing the
sale order.

To have Odoo mail digital products along with the invoice, enable
``Attach to invoice mail`` on a digital attachment of a product. This
only works as expected if you're using an instant payment method, and
have enabled ``Automatic Invoice`` in the website settings.


multimar3 (approved):

ivantodorovich (changes requested):
IMO the name `"*_improved"` is too vague. It could mean anything.

By reading `USAGE.md`, I see there are two features implemented in this module.

I recommend splitting the module in two, then, each named after the specific feature its implementing.

For example:

- `website_sale_digital_attach_to_invoice`
- `website_sale_digital_temporary_download_link`

PR #1206, opened 2026-05-11
[18.0][MIG] product_alias (was shopinvader_product_alias)

sebalix:
Migrate `shopinvader_product_alias` (which was hosted in a PR on [shopinvader/odoo-shopinvader](https://github.com/shopinvader/odoo-shopinvader)) to 18.0 and rename it `product_alias` as it doesn't depend at all on Shopinvader.

PR #1208, opened 2026-05-11
[18.0][MIG] product_alias_multi_link

sebalix:
Depends on:
- #1206


paradoxxxzero (approved):
LGTM

product_alias_multi_link/tests/test_product_alias_multi_link.py
```diff
@@ -0,0 +1,286 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(
+            context=dict(
+                cls.env.context,
+                tracking_disable=True,
+            )
+        )
```
metaminux:
```suggestion
```
BaseCommon is already disabling tracking

metaminux (approved):

product_alias_multi_link/tests/test_product_alias_multi_link.py
sebalix: Well, `BaseCommon` introduces also tons of other things, and it appears it adds overhead when running tests. At some point we are probably not improving tests performance on CI by using it, but we gain clarity/maintainability.

metaminux: @sebalix I searched when to use TransactionCase instead of BaseCommon too and I totally agree with you. But this one is inheriting BaseCommon already, so we can remove the context update or inherit TransactionCase...

sebalix: My mistake :) Done

metaminux (approved):

sebastienbeau:
@sebalix can you rebase to remove the dependency merged ?

sebastienbeau (changes requested):
need to remove the unmerge dependency

sebalix:
@sebastienbeau no issue with current module, but a timeout with Odoo tour in `website_sale_stock_list_preview` tests, `OCA/OCB` went fine however. To re-trigger :crossed_fingers:

PR #1213, opened 2026-05-18
[16.0][FIX] website_sale_stock_available

SuShanVoong:
## Description

  It was identified that within the `website_sale_stock_available` module, when computing stock quantities for a recordset of multiple products with the `website_sale_stock_available` context flag enabled, the `_compute_quantities_dict` override iterated the recordset one product at a time, reading `product.immediately_usable_qty` on each single-record recordset.

## Cases Covered with This Improvement

  - Rendering of /shop and `/shop/category/... pages` where multiple product cards are displayed and the stock cascade is evaluated for each one.
  - Any caller invoking `_compute_quantities_dict` on a multi-product recordset with website_sale_stock_available=True in context (for example bulk queries from API endpoints or batched reports).

## Implemented Solution

 - The `_compute_quantities_dict` override has been refactored to invoke `_compute_available_quantities` once on the full recordset before iterating. This pre-warm populates the ORM cache for `immediately_usable_qty` for every product at once, so the subsequent per-record reads inside the loop hit the cache instead of re-triggering the compute and the underlying cascade.

 - The behavior is functionally equivalent: the returned free_qty for each product still equals its `immediately_usable_qty`, just as before. Only the number of times the cascade executes changes — from N+1 to a single invocation on the full recordset.

FL-666-8676

### DIFF
--- a/website_sale_stock_available/models/product_product.py
+++ b/website_sale_stock_available/models/product_product.py
@@ -13,11 +13,18 @@ class Product(models.Model):
         res = super()._compute_quantities_dict(
             lot_id, owner_id, package_id, from_date, to_date
         )
-        if self.env.context.get("website_sale_stock_available"):
-            products_no_ctx = self.with_context(website_sale_stock_available=False)
-            products_no_ctx._compute_available_quantities()
-            for product in products_no_ctx:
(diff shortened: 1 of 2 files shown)


IVA-BO:
LGTM!

LuisAlejandroS (approved):

PR #1215, opened 2026-05-19
[19.0] Copier update to v1.42

JasminSForgeFlow:
Upstream Odoo's ChromeBrowser.stop() (odoo/tests/common.py) sends
  Browser.close + SIGTERM to chrome, then warns if any process in the
  chrome process tree is still listed after 15s. Under Docker/CI, chrome's
  renderer subprocesses are frequently still in 'zombie' state (already
  exited, not yet reaped) when that timeout elapses, producing:

      WARNING ... Killing chrome descendants-or-self of <pid>: 5 remaining
      - chrome (zombie) x5

  The tests pass (0 failed, 0 error(s)), but checklog-odoo treats any
  non-INFO/DEBUG line as a build failure. Whitelist this specific message
  so OCA CI is not gated on a flaky upstream cleanup race.

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,8 +1,7 @@
 # Do NOT update manually; changes here will be overwritten by Copier
-_commit: v1.35
+_commit: v1.42
 _src_path: git+https://github.com/OCA/oca-addons-repo-template
 additional_ruff_rules: []
-ci: GitHub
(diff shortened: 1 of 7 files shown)


pedrobaeza:
Please update copier template instead.

pedrobaeza (approved):

PR #1219, opened 2026-05-19
[19.0][FIX] website_sale_checkout_skip_payment: ensure correct order state

NICO-SOLUTIONS:
This is to ensure the sale order is confirmed before the confirmation email is generated.

Previously, the confirmation email was sent before calling action_confirm(), resulting in email content being rendered with an outdated order state.

### DIFF
--- a/website_sale_checkout_skip_payment/README.rst
+++ b/website_sale_checkout_skip_payment/README.rst
@@ -78,18 +78,23 @@ Authors
 Contributors
 ------------
 
-- `Tecnativa <https://www.tecnativa.com>`__
+-  `Tecnativa <https://www.tecnativa.com>`__
 
-  - Sergio Teruel
-  - David Vidal
-  - Alexandre Díaz
-  - Carlos Roca
-  - Pilar Vargas
+   -  Sergio Teruel
+   -  David Vidal
+   -  Alexandre Díaz
+   -  Carlos Roca
+   -  Pilar Vargas
 
-- Martin Wilderoth <martin.wilderoth@linserv.se>
-- `Studio73 <https://www.studio73.es>`__:
+-  Martin Wilderoth <martin.wilderoth@linserv.se>
+-  `Studio73 <https://www.studio73.es>`__:
 
-  - Miguel Gandia <miguel@studio73.es>
+   -  Miguel Gandia <miguel@studio73.es>
+
+-  `NICO SOLUTIONS - ENGINEERING &
+   IT <https://www.nico-solutions.de>`__:
+
+   -  Nils Coenen <nils.coenen@nico-solutions.de>
 
 Maintainers
 -----------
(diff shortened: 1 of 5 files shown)


NICO-SOLUTIONS:
@miguel-S73 
@sergio-teruel 
@pilarvargas-tecnativa 

I noticed that the confirmation email was being rendered prematurely. At the time the email was generated, the sale order had not yet reached the sale state, which led to missing or inconsistent confirmation content in the email (e.g. “confirmed” sections not appearing correctly).

This was caused by sending the email directly via _send_order_confirmation_mail() before calling action_confirm(). As a result, the QWeb template did not receive the final order state.

I refactored the flow to use the send_email context flag inside action_confirm() instead. This ensures that:
• the order is fully confirmed before the email is sent
• the email template receives the correct sale state
• the confirmation content is rendered consistently

Let me know if you see any edge cases or concerns with this approach.

NICO-SOLUTIONS:
@CRogos
if you have the time...

pilarvargas-tecnativa (approved):
LGTM

pilarvargas-tecnativa:
@chienandalu what do you think?

NICO-SOLUTIONS:
@chienandalu 
Hey, do you mind taking a look when you have a moment? Would be great.

I’m confident that the problem is correctly identified and that the fix should be the right approach, but I’d really appreciate a second review to ensure there are no edge cases I may have missed so we can move forward.

MiquelRForgeFlow (approved):

pedrobaeza:
@NICO-SOLUTIONS please redo the commit message to put your second comment into it, that makes more sense than the bullets content you have put initially, which doesn't correspond to the code.

NICO-SOLUTIONS:
@pedrobaeza 
Is this fine like this?

pedrobaeza:
One last check: if there's an exception like with `sale_financial_risk`, the mail is sent?

NICO-SOLUTIONS:
> One last check: if there's an exception like with `sale_financial_risk`, the mail is sent?

@pedrobaeza 
In the backend, when sale_financial_risk is triggered, action_confirm() returns a wizard and the confirmation is not completed, so no email is sent.

In the website skip-payment flow, this result is not properly handled, so the checkout process may fail silently without confirming the order and without providing appropriate user feedback. Since the order is not confirmed in this case, no confirmation email is sent.

Do we need any additional handling for this scenario in the website flow? It might be worth considering dedicated handling (e.g. a glue module or specific website integration) since the website flow does not properly handle wizard return from action_confirm().

NICO-SOLUTIONS:
@pedrobaeza 
We could replace the current try/except block with the following approach:
`result = order.with_context(
    send_email=True,
    mark_so_as_sent=True,
).action_confirm()

if isinstance(result, dict) or order.state != "sale":
    return request.render(
        "website_sale_checkout_skip_payment.confirmation_order_error"
    )`
            
This would properly handle the sale_financial_risk case (wizard return) and ensure that unsuccessful confirmations are correctly reflected in the frontend instead of being stated as successful.

Please let me know if this approach is suitable for you, and I will proceed with pushing the changes.

pedrobaeza:
@NICO-SOLUTIONS thanks for checking. Yeah, that sounds good to have a fallback flow for anything interrupting the normal confirmation if no specific glue module exists.

pedrobaeza (approved):
/ocabot merge patch

NICO-SOLUTIONS:
@pedrobaeza 
thx a lot for your input and getting this merged!  :pray:

PR #639, opened 2022-05-13
[13.0][IMP] website_sale_hide_price: hide price at product level

chienandalu:
- [x] Now we can hide the price only for the products we want to.
- [x] Adding this new flag to the comparison compatibility module
- [x] Improved readme
- [x] Fixed product price visible in recently viewed products

cc @Tecnativa TT36129

Ready to review: @CarlosRoca13 @victoralmau

PR #1224, opened 2026-05-26
[15.0][IMP] website_sale_hide_price: add an option to show website_hide_price_message to public users

smorita7749:
Currently, non-logged-in users cannot see the website_hide_price_message, so they have no way to understand why product prices are hidden.

This improvement adds a website-level option "Show hidden price message to public users". When enabled, the website_hide_price_message is also shown to public (non-logged-in) users.
@qrtl QT6775


yostashiro (approved):
Looks reasonable to me, but maybe @chienandalu could explain the rationale of the original design which was introduced in https://github.com/OCA/e-commerce/pull/639?

chienandalu:
> Looks reasonable to me, but maybe @chienandalu could explain the rationale of the original design which was introduced in #639?

The idea was something along these lines: when you're visiting the site as public user, you can't see any price in the eCommerce, but when you do it as portal user, only some selected products have their prices hidden (for commercial reasons or whatever) and the site manager can show an extra message explaining why.

This was in a b2b scenario where these prices are very strategic

smorita7749:
> > Looks reasonable to me, but maybe @chienandalu could explain the rationale of the original design which was introduced in #639?
> 
> The idea was something along these lines: when you're visiting the site as public user, you can't see any price in the eCommerce, but when you do it as portal user, only some selected products have their prices hidden (for commercial reasons or whatever) and the site manager can show an extra message explaining why.
> 
> This was in a b2b scenario where these prices are very strategic

Thank you for your detailed explanation!
To preserve this b2b behavior, the new option is disabled by default, so nothing changes for existing sites: public users still don't see the message, and the message is shown only when the option is explicitly enabled.

Could you please take a look when you have time? Thanks a lot.

Thank you for your support! @yostashiro

AungKoKoLin1997 (approved):
LG 👍

ivantodorovich (approved):
/ocabot merge minor

PR #1226, opened 2026-05-27
[18.0][IMP] website_sale_hide_empty_category: Use _read_group and cache to performance

sergio-teruel:
@Tecnativa

### DIFF
--- a/website_sale_hide_empty_category/models/product_public_category.py
+++ b/website_sale_hide_empty_category/models/product_public_category.py
@@ -16,12 +16,27 @@ class ProductPublicCategory(models.Model):
     @api.depends("product_tmpl_ids", "child_id.has_product_recursive")
     @api.depends_context("website_id")
     def _compute_has_product_recursive(self):
-        for category in self:
-            website = self.env["website"].get_current_website()
-            website_domain = website.sale_product_domain()
-            has_products = bool(
-                category.product_tmpl_ids.filtered_domain(website_domain)
+        website = self.env["website"].get_current_website()
+        website_domain = website.sale_product_domain()
+        data = self.env["product.template"]._read_group(
+            domain=website_domain,
+            groupby=["public_categ_ids"],
+        )
+        used_category_ids = set()
+        for group in data:
+            category = group[0]
+            if not category or (category.website_id and category.website_id != website):
+                continue
+            used_category_ids.update(
+                int(category_id)
+                for category_id in category.parent_path.split("/")
+                if category_id
             )
-            category.has_product_recursive = has_products or any(
-                child.has_product_recursive for child in category.child_id
-            )
+        # TODO: Filter categories before split in c to avoid cache because compute is
+        #  called with only one category
+        categories = self.search(website.website_domain()) | self
+        self.env.cache.update_raw(
+            categories,
+            self._fields["has_product_recursive"],
+            [category.id in used_category_ids for category in categories],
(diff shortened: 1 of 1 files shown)


website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,12 +16,27 @@ class ProductPublicCategory(models.Model):
...
-            website = self.env["website"].get_current_website()
-            website_domain = website.sale_product_domain()
-            has_products = bool(
-                category.product_tmpl_ids.filtered_domain(website_domain)
-            )
-            category.has_product_recursive = has_products or any(
-                child.has_product_recursive for child in category.child_id
+        website = self.env["website"].get_current_website()
+        website_domain = website.sale_product_domain()
+        data = self.env["product.template"]._read_group(
```
ivantodorovich, line 21: I'm not sure the read_group is the right choice here, because we don't really care about the grouped products.. so that's just overhead.

How about this instead?

```py
categories = self.search([("product_tmpl_ids", "any", website_domain)])
```

website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,12 +16,27 @@ class ProductPublicCategory(models.Model):
...
+            domain=website_domain,
+            groupby=["public_categ_ids"],
+        )
+        used_category_ids = set()
+        for group in data:
+            category = group[0]
+            if not category or (category.website_id and category.website_id != website):
```
ivantodorovich, line 28: I would add these filtering conditions to the search domain directly.
If you followed my other comment recommendation, it would be like:

```py
categories = self.search(
    [
        ("website_id", "in", [False, website.id]),
        ("product_tmpl_ids", "any", website_domain),
    ]
)
```

website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,12 +16,27 @@ class ProductPublicCategory(models.Model):
...
+                continue
+            used_category_ids.update(
```
ivantodorovich, line 30: I'm not sure if this works (nested `"any"` in domain) but possibly this can be part of the search domain too, like this:

```py
categories = self.search(
    [
        ("website_id", "in", [False, website.id]),
        "|",
        ("product_tmpl_ids", "any", website_domain),
        ("child_ids", "any", ("product_tmpl_ids", "any", website_domain)),
    ]
)
```

website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,12 +16,27 @@ class ProductPublicCategory(models.Model):
...
+                int(category_id)
+                for category_id in category.parent_path.split("/")
+                if category_id
             )
+        # TODO: Filter categories before split in c to avoid cache because compute is
+        #  called with only one category
+        categories = self.search(website.website_domain()) | self
+        self.env.cache.update_raw(
```
ivantodorovich, line 38: I'm not sure about this respects the ORM, tbh.

For instance if you have other computed fields depending on them, I don't think their recomputations will be triggered like this.

Maybe the cache can be used but not like this.

Perhaps we could check if we already have a cached `has_product_recursive` value for any of the children categories - -and if we do we set to `True` directly bypassing the SQL-based computation.

But I'd say first check the performance with the recommendations I provided above, maybe it's not even needed

PR #1228, opened 2026-05-29
[18.0][MIG] website_sale_affiliate migration from 16.0 to 18.0

jans23:


jans23:
@fcayre please review.

jans23:
@pilarvargas-tecnativa could you review this please?

pilarvargas-tecnativa (changes requested):
Hello,
Thank you for your contribution. Please check the commit history as it is incomplete.
Regards!

KDRdev:
> Hello, Thank you for your contribution. Please check the commit history as it is incomplete. Regards!

@pilarvargas-tecnativa thank you for the review. Could you please check if the current PR version has been prepared as expected? Thanks!

jans23:
@pilarvargas-tecnativa Updated accordingly. Please check.

website_sale_affiliate/models/sale_affiliate.py
```diff
@@ -10,118 +9,119 @@
...
+        compute="_compute_conversion_rate",
+        help="Conversion count / Request count",
     )
     sales_per_request = fields.Float(
         string="Average Sales per Request",
         digits=(12, 4),
-        compute='_compute_sales_per_request',
-        help='Sale count / Request count',
+        compute="_compute_sales_per_request",
+        help="Sale count / Request count",
```
pilarvargas-tecnativa, line 63: Changing ' to " doesn't help with the migration; it just makes the diff bigger

website_sale_affiliate/tests/test_sale_affiliate.py
```diff
@@ -28,12 +27,14 @@ def test_compute_conversion_rate_with_requests(self):
...
-            'valid_sales': 1,
-        })
+        test_affiliate = self.env["sale.affiliate"].create(
+            {
+                "name": "test_affiliate",
+                "company_id": self.demo_company.id,
+                "valid_hours": 24,
+                "valid_sales": 1,
+            }
+        )
```
pilarvargas-tecnativa, line 37: This change doesn't make any difference either

website_sale_affiliate/tests/test_sale_affiliate.py
```diff
@@ -82,50 +87,57 @@ def test_find_from_kwargs_aff_ref_absent(self):
...
-        kwargs = {'aff_ref': 'not_int'}
+        kwargs = {"aff_ref": "not_int"}
         affiliate = self.Affiliate.find_from_kwargs(**kwargs)
         self.assertIsNone(affiliate)
 
     def test_get_request_aff_key_present_and_request_exists(self):
         """Returns existing affiliate request record matching aff_key
         from kwargs when match already exists"""
-        kwargs = {'aff_key': self.demo_request.name}
+        kwargs = {"aff_key": self.demo_request.name}
```
pilarvargas-tecnativa, line 97: ''

website_sale_affiliate/controllers/main.py
```diff
@@ -1,31 +1,34 @@
...
 # Copyright 2017 LasLabs Inc.
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+import logging
 
-from odoo.addons.website_sale.controllers.main import WebsiteSale as Base
 from odoo.http import request, route
 
+from odoo.addons.website_sale.controllers.main import WebsiteSale as Base
+
+_logger = logging.getLogger(__name__)
```
pilarvargas-tecnativa: Is it necessary?

website_sale_affiliate/controllers/main.py
```diff
@@ -1,31 +1,34 @@
...
 class WebsiteSale(Base):
     def _store_affiliate_info(self, **kwargs):
-        Affiliate = request.env['sale.affiliate']
-        affiliate = Affiliate.sudo().find_from_kwargs(**kwargs)
-        try:
-            affiliate_request = affiliate.get_request(**kwargs)
-            request.session['affiliate_request'] = affiliate_request.id
-        except (AttributeError, ValueError):
-            pass
+        affiliate = request.env["sale.affiliate"].sudo().find_from_kwargs(**kwargs)
```
pilarvargas-tecnativa: This change doesn't help with the migration either; it just adds noise to the diff

website_sale_affiliate/controllers/main.py
```diff
@@ -1,31 +1,34 @@
...
+        if not affiliate:
+            return
+
```
pilarvargas-tecnativa: It is best to avoid leaving spaces between lines of code

website_sale_affiliate/controllers/main.py
```diff
@@ -1,31 +1,34 @@
...
         return res
 
     @route()
-    def product(self, product, category='', search='', **kwargs):
-        res = super(WebsiteSale, self).product(product, category='',
-                                               search='', **kwargs)
-        self._store_affiliate_info(**kwargs)
+    def product(self, *args, **post):
+        res = super().product(*args, **post)
+        self._store_affiliate_info(**post)
```
pilarvargas-tecnativa: why?

website_sale_affiliate/controllers/main.py
```diff
@@ -1,31 +1,34 @@
...
+            return  # pragma: no cover
+
+        request.session["affiliate_request"] = affiliate_request.id
 
     @route()
-    def shop(self, page=0, category=None, search='', ppg=False, **post):
-        res = super(WebsiteSale, self).shop(page, category,
-                                            search, ppg, **post)
+    def shop(self, *args, **post):
+        res = super().shop(*args, **post)
```
pilarvargas-tecnativa: why?

pilarvargas-tecnativa (changes requested):

website_sale_affiliate/models/sale_affiliate.py
KDRdev: According to the OCA module migration technical guidelines (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0#technical-method-to-migrate-a-module-from-170-to-180-branch), pre-commit needs to be run during migration. Running pre-commits causes these changes, so I think it's correct.

website_sale_affiliate/tests/test_sale_affiliate.py
KDRdev: Caused by pre-commit, which is recommended during module migration according to OCA

website_sale_affiliate/tests/test_sale_affiliate.py
KDRdev: Caused by pre-commit, which is recommended during module migration according to OCA

website_sale_affiliate/controllers/main.py
KDRdev: Probably appeared after running pre-commit, which is expected during module migration.

website_sale_affiliate/controllers/main.py
KDRdev: In this case I'd argue that this change can stay since we're separating different logical blocks (line 16 is a return statement).

website_sale_affiliate/controllers/main.py
KDRdev: It's added to allow generating affiliate links with '/shop/...'. Then we can track if order source uses affiliate links by checking the field `affiliate_request_id`on sale order.

website_sale_affiliate/controllers/main.py
KDRdev: It's added to allow generating affiliate links with '/shop/product...'. Then we can track if order source used affiliate links by checking the field `affiliate_request_id`on sale order.

website_sale_affiliate/controllers/main.py
KDRdev: No, it's not needed. Thanks!

KDRdev:
@pilarvargas-tecnativa I've replied to your comments, could you please have a look. Thank you!

website_sale_affiliate/models/sale_affiliate.py
pilarvargas-tecnativa: In any case, it is advisable to run `pre-commit` and save any changes you make in a separate commit, rather than including them in the migration commit. This ensures that the functional changes made by the migration are clearly separated from the automatic formatting adjustments, making it easier to review and track the commit history.

```
pre-commit run -a
git add -A
git commit -m "[IMP] affiliate: pre-commit auto fixes" --no-verify
```

website_sale_affiliate/controllers/main.py
pilarvargas-tecnativa: I understand that pre-commit may have changed the formatting, but this diff does not look like a pure formatting change to me.

It removes the try/except block and the assignment to request.session["affiliate_request"], so it seems to change the behavior, not just the style. Could you please check whether this change was intentional and, if so, explain why it is needed for the migration?

website_sale_affiliate/models/sale_affiliate.py
pedrobaeza: @KDRdev following strictly the migration guide you have referred, that changes are isolated in its own commit, not on the migration commit.

website_sale_affiliate/controllers/main.py
pilarvargas-tecnativa: I understand the purpose, but this is a functional change, not a migration one.

If /shop affiliate links were not supported before, adding support for them should not be part of the migration commit. For a 16.0 to 18.0 migration, we should keep the existing behavior working in 18.0 and avoid introducing new tracking flows unless they are required by a framework/API change.

Could you please move this to a separate commit/PR, or explain which 18.0 change makes it necessary for the migration?

website_sale_affiliate/models/sale_affiliate.py
KDRdev: Ok, so migration itself (syntactic, Odoo-standard business logic-dependent changes) should be a separate commit, pre-commit should be a separate commit, correct?. If there are some extra changes/additional functions, each should be a separate commit, but can still be included in this PR, right?

pedrobaeza: Just follow the guide...

website_sale_affiliate/controllers/main.py
pilarvargas-tecnativa: Finally, this change does not appear to stem from a pre-commit or an automatic version update. The flow of _store_affiliate_info is being modified, removing the original try/except block and changing how cases without an affiliate or without an affiliate_request are handled. That may be correct, but it should then be justified as a functional change and covered by tests. If it is not strictly necessary for the migration, it would be better to revert it to reduce noise and risk.

KDRdev: @pilarvargas-tecnativa the flow has been modified to fix pylint error `[W8138(except-pass), WebsiteSale._store_affiliate_info] pass into block except. If you really need to use the pass consider logging that exception`. Current execution fixes this error and doesn't change the functionality and doesn't remove original session header assignment. I think it's better than just logging the error as per pylint suggestion. It's covered by tests in 'test_website_sale.py' and is a very minor change that doesn't produce extra noise. I hope this explanation justifies the change and it can be included in the migrated module.

KDRdev: @pilarvargas-tecnativa I've reconsidered my opinion and removed the change as it may have caused some tests to fail. I've added a fix for pylint error as per linter suggestion.

website_sale_affiliate/controllers/main.py
KDRdev: This change has been removed.

website_sale_affiliate/controllers/main.py
KDRdev: This is now changed only due to change in Odoo standard /shop endpoint changes and to make sure unit tests pass

website_sale_affiliate/tests/test_sale_affiliate.py
KDRdev: Moved to a separate commit from migration commit.

website_sale_affiliate/tests/test_sale_affiliate.py
KDRdev: Moved to a separate commit from migration commit.

KDRdev:
@pilarvargas-tecnativa I've made changes as per your suggestions, could you please have another look? Thanks!

KDRdev:
@pilarvargas-tecnativa just a friendly reminder for you. Thanks!

website_sale_affiliate/models/sale_order.py
```diff
@@ -13,7 +13,7 @@ class SaleOrder(models.Model):
         help="Affiliate request associated with sale order",
     )
 
-    @api.model
+    @api.model_create_multi
```
pilarvargas-tecnativa, line 16: Is this change to @api.model_create_multi really needed for the migration?

As it is now, the method is decorated as multi-create but still receives vals as a single dict and processes order as a single record. So it looks half-adapted and could be misleading or problematic if several records are created at once.

If there is no specific need to support multi-create here, I think it would be cleaner to keep the previous @api.model signature for the migration.

website_sale_affiliate/models/sale_affiliate.py
```diff
@@ -91,19 +91,17 @@ def _default_sequence_id(self):
...
-    @api.model_cr_context
     def find_from_kwargs(self, **kwargs):
         """Find affiliate record based on kwargs"""
         try:
             affiliate_id = int(kwargs["aff_ref"])
             return self.search([("id", "=", affiliate_id)], limit=1)
-        except KeyError:
-            pass
+        except KeyError as err:
+            _logger.debug(err)
```
pilarvargas-tecnativa: Is there any reason to log this exception?

Previously, the absence of aff_ref was silently ignored, which seems to be the expected and common case. I'm wondering whether this debug log adds any value, or if it would be better to keep the previous behavior.

website_sale_affiliate/controllers/main.py
```diff
@@ -13,12 +17,21 @@ def _store_affiliate_info(self, **kwargs):
         try:
             affiliate_request = affiliate.get_request(**kwargs)
             request.session["affiliate_request"] = affiliate_request.id
-        except (AttributeError, ValueError):
-            pass
+        except (AttributeError, ValueError) as err:
+            _logger.debug("Could not store affiliate info: %s", err)
```
pilarvargas-tecnativa: Same question here: is there any specific reason for adding this log?

Since this is a migration, I'd prefer to avoid introducing new logging unless it addresses a known issue or provides information that is actually useful in production.

jans23:
@pilarvargas-tecnativa Thank you very much for the review. I just addressed it in the code.

website_sale_affiliate/models/sale_order.py
jans23: I fixed this.

website_sale_affiliate/models/sale_affiliate.py
jans23: I fixed this.

website_sale_affiliate/controllers/main.py
jans23: I fixed this.

jans23:
@pedrobaeza this got two reviews and it could be merged I think.

pedrobaeza:
But there aren't approvals.

KDRdev (approved):

jans23:
@pilarvargas-tecnativa All your comments are addressed. Please have another look.

pilarvargas-tecnativa:
> @pilarvargas-tecnativa All your comments are addressed. Please have another look.

You don’t need a separate commit to migrate the module as a FIX for the migration, simply include the changes in the migration commit

jans23:
@pilarvargas-tecnativa Squashed as requested.

jans23:
@pilarvargas-tecnativa please have a look.

PR #1230, opened 2026-06-02
[19.0][ADD] website_sale_search_display_name

MiquelRForgeFlow:
Website Sale Search Display Name
================================

This module makes the website product search render the result title from ``display_name`` instead of ``name``.

By default the website search (the autocomplete dropdown and the ``/website/search`` results page) shows the product template ``name``, whereas the rest of the eCommerce (shop grid, product page) uses ``display_name``, which prepends the internal reference (``default_code``) when the ``display_default_code`` context flag is set.

As a result, a product reached through the search bar drops its reference even though ``default_code`` and ``variants_default_code`` are already searchable. This module makes the search results consistent with the rest of the shop and exposes the product reference.


pedrobaeza:
I don't think this is applicable, as you are searching here for templates, not variants, so both the default code and the possible attributes are not yet applicable.

MiquelRForgeFlow:
> I don't think this is applicable, as you are searching here for templates, not variants, so both the default code and the possible attributes are not yet applicable.

I don't know what you mean, as it works perfectly.

# Before:
Selection_5961

# After:
Selection_5962

pedrobaeza:
Then basically you are adding the default code when the template has only one variant... This should be added to `website_sale_product_reference_displayed` IMO.

MiquelRForgeFlow:
> This should be added to `website_sale_product_reference_displayed` IMO.

Done in https://github.com/OCA/e-commerce/pull/1234.

PR #1233, opened 2026-06-03
[19.0][ADD] website_sale_partner_sale_contact

JasminSForgeFlow:
When an order is placed from the website, the logged-in user is usually a
child contact of a company. Standard `website_sale` sets that contact as the
order customer (`partner_id`).

This module mirrors the backend `sale_partner_sale_contact` behaviour on the
website: the **parent company** becomes the order customer and the **contact**
is stored in the **Sale Contact** field (`sale_contact_partner_id`). Billing
and shipping addresses are left untouched.

Depends on:

- [ ] https://github.com/OCA/sale-workflow/pull/4225

@ForgeFlow


AaronHForgeFlow (approved):
Functional 👍

website_sale_partner_sale_contact/models/res_partner.py
```diff
@@ -0,0 +1,32 @@
...
+            return commercial
+        if not self.is_company and self.company_name:
+            company = self.search(
+                [("is_company", "=", True), ("name", "=", self.company_name)],
+                limit=1,
+            )
+            if not company:
+                company = self.create({"name": self.company_name, "is_company": True})
+            if self.parent_id != company:
+                self.parent_id = company
```
ivantodorovich, line 30: This if-condition seems pointless. We've just created `company` so it's safe to assume it's not the contact's parent

```suggestion
            self.parent_id = company
```

ivantodorovich (approved):

BhaveshHeliconia (approved):
Tested locally functional review LGTM!

PR #1256, opened 2026-06-26
[19.0][MIG] website_sale_wishlist_hide_price

DavidJForgeFlow:
Supersedes #1147 

Made by @JasminSForgeFlow


ivantodorovich:
Hello 👋🏻 Thanks!

Could you check the CI, though? it's red

website_sale_wishlist_hide_price/views/website_sale_template.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="wishlist_page_hide_price"
+        inherit_id="website_sale_wishlist.product_wishlist"
+    >
+        <xpath expr="//div[hasclass('o_wish_price')]/span" position="attributes">
```
cristina-hidalgo-tecnativa, line 7:
```suggestion
        <xpath expr="//div[hasclass('o_wish_price')]" position="attributes">
```
Hi! Thanks for the work you did, It really looks nice! 

I tested it locally and I found a small edge case: 

Image

When the product has an active discount, the base template renders the original price as a sibling `<small><del>` next to this `<span>`, so it stays unaffected by this `t-if` (no discount = no `<small>`, so easy to miss in testing).

Moving the `t-if` one level up, to the `div.o_wish_price` wrapper, covers both

TT64170

website_sale_wishlist_hide_price/tests/test_tour.py
```diff
@@ -0,0 +1,29 @@
...
+                "email": "wishlist_hide_price_user@example.com",
+            }
+        )
+
+    def test_tour(self):
+        self.start_tour(
+            "/shop",
+            "website_sale_wishlist_hide_price_tour",
+            login="wishlist_hide_price_user",
+            debug=True,
```
cristina-hidalgo-tecnativa, line 28:
```suggestion
            
```

This `debug=True` is what's failing CI: it forces `watch=True` in `start_tour`, which launches Chrome non-headless (fullscreen + devtools), and `--no-sandbox` is only added in headless mode — hence "Running as root without --no-sandbox is not supported". It's meant for local debugging only.

cristina-hidalgo-tecnativa (changes requested):

website_sale_wishlist_hide_price/tests/test_tour.py
```diff
@@ -0,0 +1,29 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+# Copyright 2024 Tecnativa - Pilar Vargas
+
+from odoo.tests import HttpCase, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestWebsiteSaleProductAttachmentTourl(HttpCase):
```
cristina-hidalgo-tecnativa, line 8: Looks like this got copy-pasted from another module's test ("ProductAttachment" + a stray l).

cristina-hidalgo-tecnativa (changes requested):

pedrobaeza:
Superseded by #1323

PR #1258, opened 2026-06-30
[19.0][ADD] website_sale_tax_selection: website sale tax display partner preference

yankinmax:
Allow partners to specify how prices are displayed on the e-commerce website: tax excluded, tax included, or standard website behavior when no preference is set.
The website tax display field is made non-stored so it can be computed dynamically from the current website user’s partner, with contact preferences taking precedence over commercial entity preferences.


website_sale_tax_selection/models/website.py
```diff
@@ -0,0 +1,55 @@
...
+        """Return the partner used to resolve website tax display preference.
+
+        :return: Current HTTP request user partner, or environment user partner
+            when running outside an HTTP request.
+        :rtype: res.partner
+        """
+        try:
+            return request.env.user.partner_id
+        except RuntimeError:
+            return self.env.user.partner_id
```
ivantodorovich:
```suggestion
        if request and request.env:
            return request.env.user.partner_id
        else:
            return self.env.user.partner_id
```

nitpicking: this pattern https://github.com/OCA/odoo/blob/5cfaa4999856950b053ad3461ea0a277714491b8/addons/website/models/res_users.py

website_sale_tax_selection/models/website.py
```diff
@@ -0,0 +1,55 @@
...
+        The contact value takes precedence over the commercial entity value.
+
+        :return: Preferred tax display value, or False when none is set.
+        """
+        partner = self._get_current_website_tax_selection_partner()
+        if not partner:
+            return False
+        if partner.website_show_line_subtotals_tax_selection:
+            return partner.website_show_line_subtotals_tax_selection
+        commercial_partner = partner.commercial_partner_id
```
ivantodorovich: Instead, add the field to the `_commercial_fields()` and then it's configurable company-wide, which makes more sense -- and you don't have to explicitly check both.

In fact if you do that, you can drop this method and simply read `partner.website_show_line_subtotals_tax_selection`.

ivantodorovich (commented):
Thanks!

website_sale_tax_selection/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
```
Ricardoalso:
```suggestion
```

website_sale_tax_selection/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
```
Ricardoalso:
```suggestion
```

website_sale_tax_selection/models/website.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class Website(models.Model):
+    _inherit = "website"
+
+    # OVERRIDE: make the tax display dynamic so the current website partner can
+    # specify a value that takes precedence over the standard website behavior.
+    show_line_subtotals_tax_selection = fields.Selection(
+        store=False,
+    )
```
Ricardoalso, line 15: Will this not cause any change in settings to appear succeeding but never reaching the DB ? 🤔  

If we really want this, we should probably hide it in the configurations view and setting a default value in res.partner.website_show_line_subtotals_tax_selection

ivantodorovich: Yes you're right, we should hide it from the settings

website_sale_tax_selection/models/website.py
yankinmax: Of course, thanks

website_sale_tax_selection/readme/DESCRIPTION.md
Ricardoalso, line 1: This description needs to be updated. Stating it will visually disable the website_tax_inclusion_setting is IMHO crucial

website_sale_tax_selection/models/website.py
```diff
@@ -0,0 +1,38 @@
...
+    @api.depends_context("uid")
+    def _compute_show_line_subtotals_tax_selection(self):
+        # OVERRIDE: apply the current website partner tax display preference,
+        # when set, on top of the standard website behavior.
+        partner = self._get_current_website_tax_selection_partner()
+        if partner_tax_selection := partner.website_show_line_subtotals_tax_selection:
+            for website in self:
+                website.show_line_subtotals_tax_selection = partner_tax_selection
+        else:
+            return super()._compute_show_line_subtotals_tax_selection()
```
Ricardoalso, line 27: This will never trigger if `partner.website_show_line_subtotals_tax_selection` gets updated during the session 🤔 

I would add something like this write override in `res.partner`: 

```
def write(self, vals):
    result = super().write(vals)
    if "website_show_line_subtotals_tax_selection" in vals:
        # The website's non-stored compute reads this field through the
        # current request/env user's partner, a path `@api.depends` can't
        # express, so its cache is never auto-invalidated on this write.
        self.env["website"].invalidate_model(
            ["show_line_subtotals_tax_selection"]
        )
    return result
```

website_sale_tax_selection/readme/DESCRIPTION.md
yankinmax: Good catch, thanks

website_sale_tax_selection/models/website.py
yankinmax: Agree with you

apf-c2c:
Functional test:
Left is contact with Tax Excl. configured and right is tax included (public user in this case):
image

Checkout tax applied:
image


:+1:

ivantodorovich (approved):

apf-c2c (approved):
Functional review 👍

Ricardoalso (approved):
Thanks 🙏

BhaveshHeliconia (approved):
LGTM!

PR #1260, opened 2026-07-03
[19.0][MIG] website_sale_require_legal: Migration to version 19.0

pilarvargas-tecnativa:
@Tecnativa TT63427

@eduezerouali-tecnativa @juancarlosonate-tecnativa  please review


pilarvargas-tecnativa:
The error is caused by the website_sale_charge_payment_fee module, this needs to be checked

website_sale_require_legal/templates/website_sale.xml
eduezerouali-tecnativa, line 27: Shouldn't this line be removed?

eduezerouali-tecnativa (changes requested):

website_sale_require_legal/templates/website_sale.xml
pilarvargas-tecnativa: I left it in because, although it is no longer sufficient for the view to appear in the Website Builder, `customize_show` still marks the view as a switchable customisation and is still used by `get_switchable_related_views()`. I haven’t found that keeping it causes any harm, and I think it better reflects the intention of the view.

juancarlosonate-tecnativa (approved):
LGTM

pedrobaeza:
/ocabot migration website_sale_require_legal

The README is still not clarifying why this module is needed over the standard legal terms.

pilarvargas-tecnativa:
> The README is still not clarifying why this module is needed over the standard legal terms.

Isn’t the detailed explanation provided in context at the time clear enough? Should it be added elsewhere in the README?

pedrobaeza:
That explanation is not reflected in the module itself, and now I don't remember it.

pilarvargas-tecnativa:
> That explanation is not reflected in the module itself, and now I don't remember it.

I've added a short sentence to the description to make that distinction explicit. I thought the context section already conveyed that idea, but I agree it's useful to mention it earlier as well. Do you think it's clear enough now?

https://github.com/OCA/e-commerce/blob/5c6a96b77f3d3d2293731cc043447f2d9fcd1e44/website_sale_require_legal/README.rst

pedrobaeza:
Thanks, now I remember, although I think the standard is enough for most of the cases.

pilarvargas-tecnativa:
> Thanks, now I remember, although I think the standard is enough for most of the cases.

I agree, the standard option is enough for most cases. However, this functionality is still requested in real-world scenarios: requiring the acceptance of the legal terms when creating or editing an address, together with keeping metadata about the acceptance for audit or legal purposes.

eduezerouali-tecnativa (approved):
Code LGTM + tested on runboat

PR #1271, opened 2026-07-24
[19.0][MIG] website_sale_hide_empty_category: Migration to 19.0

prayagrk:
website_sale_hide_empty_category: Migration to 19.0


prayagrk:
@ivantodorovich @remytms @wpichler , Please have a look at this PR.

Thanks

website_sale_hide_empty_category/models/website_snippet_filter.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2017 LasLabs Inc.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+from odoo import api, models
+from odoo.http import request
+
+
+class WebsiteSnippetFilter(models.Model):
```
cristina-hidalgo-tecnativa: Hi! Nice job you did!

Quick question about `models/website_snippet_filter.py`: 

v19's `website_sale` already returns `'unpublished': not cat.has_published_products` per category in core's [_prepare_category_list_data()](https://github.com/odoo/odoo/blob/19.0/addons/website_sale/models/website_snippet_filter.py#L169) (`has_published_products` is computed recursively — same idea as this module's `has_product_recursive`).

Was there a reason to re-browse the categories and recompute `has_product_recursive` instead of just using the unpublished key core already provides?, e.g.:

```
<!-- hide_categories_snippet_recursive -->
<attribute name="t-if">(not is_parent or includeParent) and not category.get('unpublished')</attribute>
```
Asking because it'd drop this file (and its tests) entirely and avoid an extra ORM query per snippet render — but maybe I'm missing a case where the two diverge?

TT64158

prayagrk: Hi Cristina,

Great catch! You're right. v19's native has_published_products / 'unpublished' key performs the exact same recursive check via SQL.

I've refactored the XML templates to use the core flags and removed the Python files and tests. The module is now XML-only, avoiding the extra ORM queries during snippet rendering.

BhaveshHeliconia (approved):
Functional review LGTM!

website_sale_hide_empty_category/i18n/website_sale_hide_empty_category.pot
cristina-hidalgo-tecnativa, line 1: These i18n entries reference the removed `has_product_recursive` field/model — now that the module is XML-only, they're stale. Could you regenerate the `.pot`/`.po` files (should end up basically empty)?

cristina-hidalgo-tecnativa (changes requested):

cristina-hidalgo-tecnativa (approved):
Code reviewed and tested locally and on runboat, LGTM! 

Finally, could you squash the administrative commits to keep the git history tidier?

PR #1279, opened 2026-08-06
[19.0][ADD] website_sale_comparison_specification_variant

Ricardoalso:
On the eCommerce specs table, hide non-variant-defining attribute values that are excluded for the currently selected variant combination, and refresh the table when the combination changes.


ivantodorovich (commented):
May I suggest name `website_sale_comparison_specification_variant` --> cause the purpose is not only about the exclusion rules, but to show/update the spec of the selected variant

Ricardoalso:
Superseded by https://github.com/OCA/e-commerce/pull/1280

PR #1280, opened 2026-08-06
[19.0][ADD] website_sale_comparison_specification_variant

Ricardoalso:
On the eCommerce product page's "Specifications" display and on the comparison page (/shop/compare), hide non-variant-defining attribute values that are excluded for the currently selected variant combination(s), and refresh the display when the combination changes.


ivantodorovich:
I would expect this table to show only "White" and "Steel" here:
image

Same if I change the attribute combinations, I see no update on the table:

https://github.com/user-attachments/assets/771a5030-d3d7-4830-9dba-2d3531f120d2

Ricardoalso:
> I would expect this table to show only "White" and "Steel" here: image
> 
> Same if I change the attribute combinations, I see no update on the table:
> 
>  Screen.Recording.2026-08-06.at.11.47.41.AM.mov

As discussed, it should now reflect our business need more appropriately

ivantodorovich (approved):
Thanks!

Tested on runboat, works as expected 🚀

imlopes (approved):

BhaveshHeliconia (approved):
LGTM!

PR #1285, opened 2026-08-13
[19.0][FIX] website_sale_require_legal: Prevent an error when base_vat is installed

carlos-lopez-tecnativa:
The tour fails when the base_vat module is installed because the VAT number is invalid.

image
TT64060
@Tecnativa @pedrobaeza @pilarvargas-tecnativa could you please review this?


pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1288, opened 2026-08-17
[19.0][MIG] website_snippet_product_category: Migration to version 19.0

cristina-hidalgo-tecnativa:
- Rewrote the frontend widget from `sAnimation` (removed in 19.0) to the new `Interaction` API.
- Replaced the old snippet options system (`web_editor`, removed) with the new Builder `Plugin` API: the save-cleanup logic moves to `clean_for_save_handlers`, which also makes the `s_product_category_options` template (it inherited `website.snippet_options`, also removed) unnecessary and lets us drop it.
- `@route(type="json")` → `@route(type="jsonrpc")` (deprecated alias in 19.0).

@Tecnativa TT64172


pilarvargas-tecnativa:
Please reduce administrative commits

website_snippet_product_category/i18n/ca.po
```diff
@@ -0,0 +1,101 @@
...
+msgid "Wooden Box"
+msgstr "Caixa de fusta"
+
+#, python-format
+#~ msgid ""
+#~ "An error occured with this product categories block. If the problem "
+#~ "persists, please consider deleting it and adding a new one"
+#~ msgstr ""
+#~ "S'ha produït un error amb aquest bloc de categories de productes. Si el "
+#~ "problema continua, penseu en suprimir-lo i afegir-ne un de nou"
```
pilarvargas-tecnativa:
```suggestion

```

You can clean up the translations for each language to remove old leftover data and make the files look tidier

cristina-hidalgo-tecnativa:
> Please reduce administrative commits

Done!

ivantodorovich (approved):
LG

website_snippet_product_category/static/src/js/frontend.esm.js
```diff
@@ -1,78 +1,80 @@
...
+    }
+
+    start() {
+        if (this.html === null) {
+            this.onRenderError?.();
+            return;
+        }
+        const wrapperEl = document.createElement("div");
+        wrapperEl.innerHTML = this.html;
+        const contentEl = wrapperEl.firstElementChild;
```
pilarvargas-tecnativa: Why are we keeping only firstElementChild here? In v18 the whole returned HTML was inserted into the target. This seems to change the behavior beyond what is required for the migration.

website_snippet_product_category/static/src/js/frontend.esm.js
```diff
@@ -1,78 +1,80 @@
...
+        const count = contentEl?.querySelector("input[name='object_count']")?.value;
+        if (!count) {
+            this.insert(this.renderWarningEl());
+            return;
+        }
+        this.insert(contentEl);
```
pilarvargas-tecnativa: Is insert() required by the new Interaction API here? The previous implementation replaced the target contents with the returned HTML. We should keep the same behavior unless the v19 API requires otherwise.

website_snippet_product_category/static/src/js/frontend.esm.js
```diff
@@ -1,78 +1,80 @@
...
+        } catch {
+            this.html = null;
+        }
+    }
+
+    start() {
+        if (this.html === null) {
+            this.onRenderError?.();
+            return;
+        }
```
pilarvargas-tecnativa, line 39: Do we need this extra abstraction for the migration? In v18 the edit-mode error handling was done directly in the existing flow. Please keep the implementation as close as possible to the previous behavior unless this is required by the new Interaction API.

website_snippet_product_category/static/src/js/snippet.options.esm.js
```diff
@@ -1,13 +1,21 @@
...
+import {registry} from "@web/core/registry";
+
+export class ProductCategoryOptionPlugin extends Plugin {
+    static id = "websiteSnippetProductCategoryOption";
+    resources = {
+        clean_for_save_handlers: this.cleanForSave.bind(this),
+    };
+
+    cleanForSave({root}) {
+        for (const el of root.querySelectorAll(".js_product_category")) {
```
pilarvargas-tecnativa, line 13: The previous implementation only cleaned the current option target (this.$target.empty()). Why do we now clean every .js_product_category under root? Please avoid broadening the behavior unless required by the new builder API.

website_snippet_product_category/i18n/ca.po
cristina-hidalgo-tecnativa: removed the obsolete entries in `ca.po`, and also found the same leftover in `fr.po`, `es.po` and `it.po`, so I cleaned those up too.

website_snippet_product_category/static/src/js/frontend.esm.js
cristina-hidalgo-tecnativa: Switched to setting `this.el.innerHTML` directly with the full response, same as the original `$target.html($object_html)`

website_snippet_product_category/static/src/js/frontend.esm.js
cristina-hidalgo-tecnativa: Not required here, it was overkill for static content. Replaced with a direct `innerHTML` assignment and an explicit `destroy()` that empties the element, mirroring the original `$target.empty()` on destroy.

website_snippet_product_category/static/src/js/frontend.esm.js
cristina-hidalgo-tecnativa: Agreed, simplified. Found that `this.services["public.interactions"].editMode` is readable directly from the interaction, so this is now a single inline `if`, same shape as the original `if (_this.editableMode)` — no separate mixin/hook needed anymore.

website_snippet_product_category/static/src/js/snippet.options.esm.js
cristina-hidalgo-tecnativa: I found this one is required by the new API, not a behavior broadening: `clean_for_save_handlers` in the new Builder Plugin system is registered once globally and only receives the whole editable root, not a per-instance target like the old options system did. 

Iterating `root.querySelectorAll(...)` is the standard pattern the core itself uses for this exact case (e.g. [popup_visibility_plugin.js](https://github.com/OCA/OCB/blob/8d823a8f1b0354e6afd1edd01fcc9e384fad17a3/addons/website/static/src/builder/plugins/popup_visibility_plugin.js#L86-L92), [embed_code_option_plugin.js](https://github.com/OCA/OCB/blob/8d823a8f1b0354e6afd1edd01fcc9e384fad17a3/addons/website/static/src/builder/plugins/options/embed_code_option_plugin.js#L26-L34) both do the same). 

Without the loop, a page with more than one instance of this snippet would only get the first one cleaned on save.

website_snippet_product_category/readme/USAGE.md
```diff
@@ -0,0 +1,25 @@
...
+- `.categ_link` \> The link of the category row
+- `.categ_img` \> The container of the icon of the category row
+
+You can use 'data-tree-level' attribute to select the category row of a
+selected level (1-4 by design).
+
+Also you can replace the entire template replacing the `data-template`
+attribute value in the `s_product_category_options` view. You only need
+know that the name of the variable that have all the categories is
+called `object`.
```
pilarvargas-tecnativa: s_product_category_options no longer exists in v19. Please adapt these instructions to the new way of setting data-template.

website_snippet_product_category/static/src/js/frontend.esm.js
```diff
@@ -0,0 +1,89 @@
...
+            );
+        } catch {
+            this.html = null;
+        }
+    }
+
+    start() {
+        if (this.html === null) {
+            // Only inform the editor about the failure, as the original
+            // widget did through its `editableMode` flag.
```
pilarvargas-tecnativa:
```suggestion
```
Please avoid comments referring to the previous implementation/version. Comments should explain the current code only when needed.

website_snippet_product_category/static/src/js/frontend.esm.js
```diff
@@ -0,0 +1,89 @@
...
+            if (this.services["public.interactions"]?.editMode) {
+                this.renderError();
+            }
+            return;
+        }
+        // Parse the response first, just to inspect it, like the original
+        // widget did with `$(object_html)` before deciding what to do.
```
pilarvargas-tecnativa:
```suggestion
```

website_snippet_product_category/static/src/js/frontend.esm.js
```diff
@@ -0,0 +1,89 @@
...
+        );
+        this.el.append(errorEl);
+    }
+}
+
+registry
+    .category("public.interactions")
+    .add("website_snippet_product_category.product_category", ProductCategory);
+// Also render the categories while editing the page, as the original
+// widget did (disabledInEditableMode: false).
```
pilarvargas-tecnativa:
```suggestion
```

website_snippet_product_category/readme/USAGE.md
cristina-hidalgo-tecnativa: Done!

website_snippet_product_category/static/src/js/frontend.esm.js
cristina-hidalgo-tecnativa: Done!

website_snippet_product_category/static/src/js/frontend.esm.js
```diff
@@ -1,78 +1,83 @@
...
      * @override
      */
-    destroy: function () {
-        this.$target.empty();
-        this._super.apply(this, arguments);
-    },
-});
+    destroy() {
+        this.el.replaceChildren();
+    }
```
pilarvargas-tecnativa: Is overriding destroy() needed here? The previous implementation also called the parent destroy(). If clearing the generated content is still required, should we call super.destroy() as well?

cristina-hidalgo-tecnativa: It's not needed — save-time cleanup is already handled separately by `clean_for_save_handlers`. Removed the override.

And for reference: `Interaction.prototype.destroy()` is an empty no-op and its own [JSDoc](https://github.com/OCA/OCB/blob/8d823a8f1b0354e6afd1edd01fcc9e384fad17a3/addons/web/static/src/public/interaction.js#L165-L170) says `super.destroy()` isn't needed unless extending a "concrete" subclass. You can also see an example [here](https://github.com/OCA/OCB/blob/8d823a8f1b0354e6afd1edd01fcc9e384fad17a3/addons/website/static/src/snippets/s_dynamic_snippet/dynamic_snippet.js#L66-L71).

website_snippet_product_category/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,4 @@
+Adds a new snippet to show e-commerce categories.
+
+The snippet only displays categories explicitly enabled for it, and can
+optionally restrict them to selected websites.
```
carlos-lopez-tecnativa: I think this text should be clearer because, literally, it doesn't show categories that are explicitly enabled.

Currently, a category is displayed only if all of its parent categories are enabled. However, if I select a child category but its parent category is not enabled, the explicitly selected category is not displayed.

Image

Image

carlos-lopez-tecnativa (changes requested):
LGTM, but please clarify the README a bit.

website_snippet_product_category/readme/DESCRIPTION.md
cristina-hidalgo-tecnativa: You're right, updated the description to reflect that.

pilarvargas-tecnativa (approved):

pedrobaeza:
@carlos-lopez-tecnativa please review

carlos-lopez-tecnativa (approved):

cristina-hidalgo-tecnativa:
@pedrobaeza could you retry the merge please?

PR #1292, opened 2026-08-19
[18.0][FIX] website_sale_product_brand: ensure brand is detected in product  microdata

pilarvargas-tecnativa:
The meta approach was correctly detected during the initial validation, but it is not being interpreted consistently in the deployed environment. Use an explicit Brand schema structure instead to ensure reliable detection.

@Tecnativa TT64220

@victoralmau @pedrobaeza please review

### DIFF
--- a/website_sale_product_brand/views/templates.xml
+++ b/website_sale_product_brand/views/templates.xml
@@ -187,11 +187,12 @@
         <xpath expr="//span[@itemprop='image']" position="after">
             <span
                 t-if="product.product_brand_id"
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Adding span elements alters the visible elements (although putting `d-none`, their spaces "counts" AFAIK). Isn't there other way to put them?

pedrobaeza (approved):
Reviewed internally

victoralmau (approved):

PR #1295, opened 2026-08-19
[19.0][MIG] website_sale_product_attribute_filter_category: Migration to version 19.0

cristina-hidalgo-tecnativa:
- Reimplement the "Group by categories" toggle as an Owl Website Builder option (`ProductsListPageOption` extension) instead of the removed `we-checkbox`.
- Move the category-grouping xpath from `products_attributes` to `product_attribute_filters_form`, keeping the same `$0`-replace technique, with unique ids per category to avoid collisions between the sidebar and offcanvas renders.
- Update `shop()` override to match the new core signature and param names; use `request.env._()` per current translation convention.
- Update tour and tests accordingly.

@Tecnativa TT64160


pedrobaeza:
/ocabot migration website_sale_product_attribute_filter_category

website_sale_product_attribute_filter_category/static/src/website_builder/products_list_page_option.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+<t
```
pedrobaeza: Incorrect indentation, and don't leave empty lines

pedrobaeza (changes requested):
On a test in runboat, grouping by category duplicates children filters:

Image

Other thing about migrating DBs: those with this option enabled in previous version, will it be kept in this one, or a migration script should handle it?

cristina-hidalgo-tecnativa:
> On a test in runboat, grouping by category duplicates children filters:
> Image
> 
> Other thing about migrating DBs: those with this option enabled in previous version, will it be kept in this one, or a migration script should handle it?

Could you please review it again?

About the question, no migration script is needed. When updating an existing `<template>` record (not a fresh install), [odoo/tools/convert.py](https://github.com/odoo/odoo/blob/3dd41395e2e4205fa477eb474d4a4dba0a976154/odoo/tools/convert.py#L517-L524) (`_tag_template`) explicitly skips re-writing the active field if the view already exists.

So a DB that had "Group by categories" enabled in 18.0 keeps it enabled after upgrading — only `inherit_id` gets repointed to the new `product_attribute_filters_form`.

pedrobaeza (commented):
OK, now it works correctly on runboat, although 2 observations:

- The slider is not next to "Collapse sidebar sections", which is related.
- The new categories are collapsible by default, not respecting this other option.

And about migration, I'm still not sure, as if the previous view was active in 17/18, then when migrating, it's kept with noupdate=1, and not getting the new view spec. Let's see if @pilarvargas-tecnativa can confirm.

pilarvargas-tecnativa:
This is one of the views that are duplicated when they are specifically enabled for a website. Those copies associated with the website are not updated when the module is updated; only the original view is updated. Therefore, I believe a migration script with website context would indeed be necessary.

pedrobaeza:
I think this is a general problem for any activable view. We can try to mark all of those views to check if they are the same as the website_id=null ones, and if so, then update them after the migration with the new content of the agnostic view.

pedrobaeza:
The collapse/uncollapse arrows are not working well for second and successive attributes in the same category:

Peek 17-09-2026 09-10

cristina-hidalgo-tecnativa:
> The collapse/uncollapse arrows are not working well for second and successive attributes in the same category:
> Peek 17-09-2026 09-10

the `$0` copy per category was leaving several attribute accordions with the same collapse id, so opening one toggled them all. Couldn't fix it safely via xpath (the desktop collapsible markup only exists when "Collapse sidebar sections" is on, so patching it there risked crashing the page). Fixed it in JS instead: a small `Interaction` gives each accordion item its own id after render.

pedrobaeza (approved):

pilarvargas-tecnativa (changes requested):
Could we handle the duplicated IDs directly in the QWeb template instead of adding a JavaScript interaction to modify them after rendering? I think it would be cleaner to fix the IDs and their references at the template level, while keeping $0 to reuse the original Odoo template. If JavaScript is strictly necessary, could you explain why this cannot be handled in XML?

Also, the current layout looks quite different from the standard Odoo 19 filters. The category headings, spacing and nested attribute filters should be adjusted to better match the native sidebar design.

Image

cristina-hidalgo-tecnativa:
> Could we handle the duplicated IDs directly in the QWeb template instead of adding a JavaScript interaction to modify them after rendering? I think it would be cleaner to fix the IDs and their references at the template level, while keeping $0 to reuse the original Odoo template. If JavaScript is strictly necessary, could you explain why this cannot be handled in XML?
> 
> Also, the current layout looks quite different from the standard Odoo 19 filters. The category headings, spacing and nested attribute filters should be adjusted to better match the native sidebar design.
> Image

Those ids (`o_products_attributes_{{a.id}}`) come from core's templates, reused via `$0` so we don't duplicate that markup. `$0` copies the same node once per category, so the ids end up duplicated across categories.

Fixing it purely in XML would mean xpath'ing inside that copied node to rename the id/data-bs-target — but its structure changes depending on whether "Collapse sidebar sections" is active (a separate view inheriting the same template), so we'd need two conditional xpaths just for that, and one would crash whenever the other toggle state doesn't match. The JS interaction is smaller and works regardless of that toggle.

Also fixed the spacing — I was using `p-0` on the category accordion button, which stripped the vertical padding core keeps (`px-0`). Updated to match, should look closer to the native filters now.

PR #1298, opened 2026-08-20
[19.0][MIG] website_sale_secondary_unit: Migration to 19.0

CarlosRoca13:
cc @Tecnativa TT64166

This PR depends of:

- [x] https://github.com/OCA/sale-workflow/pull/4535


pedrobaeza:
/ocabot migration website_sale_secondary_unit

For being testable, I think you should include the temporary reference.

website_sale_secondary_unit/models/sale_order.py
```diff
@@ -0,0 +1,96 @@
...
+        along with the chosen ``secondary_uom_id``. When no secondary unit is
+        received (e.g. the "Add to cart" snippet or an express checkout), the
+        quantity is still expressed in secondary units for products that can't
+        be sold in their own unit, so it has to be converted here.
+        """
+        if "secondary_uom_id" not in kwargs:
+            product = self.env["product.product"].browse(product_id)
+            if not product.allow_uom_sell:
+                secondary_uom = (
+                    product.sale_secondary_uom_id or product.secondary_uom_ids[:1]
```
carlos-lopez-tecnativa: I’m not sure if this should be done in this part of the code or somewhere else, but I think you should filter to only include published records. I noticed that you have the `_get_website_secondary_uoms` function, which performs this filtering. Could this function be used in all places?

```suggestion
                    product.sale_secondary_uom_id or product._get_website_secondary_uoms()[:1]
```

website_sale_secondary_unit/readme/CONFIGURE.md
```diff
@@ -0,0 +1,3 @@
+To define the secondary units, you should activate *Units of Measure* in
```
carlos-lopez-tecnativa:
```suggestion
To define the secondary units, you should activate *Units of Measure & Packagings* in
```

website_sale_secondary_unit/security/website_sale_secondary_unit.xml
```diff
@@ -0,0 +1,19 @@
...
+    <record id="product_secondary_unit_public" model="ir.rule">
+        <field name="name">Public secondary unit</field>
+        <field
+            name="model_id"
+            ref="product_secondary_unit.model_product_secondary_unit"
+        />
+        <field name="domain_force">[('website_published', '=', True)]</field>
+        <field
+            name="groups"
+            eval="[(4, ref('base.group_public')), (4, ref('base.group_portal'))]"
```
carlos-lopez-tecnativa:
```suggestion
            eval="[Command.link(ref('base.group_public')), Command.link(ref('base.group_portal'))]"
```

website_sale_secondary_unit/tests/test_website_sale_secondary_unit.py
```diff
@@ -0,0 +1,53 @@
...
+                        [cls.secondary_unit_box_5.id, cls.secondary_unit_box_10.id],
+                    ),
+                ],
+            }
+        )
+        # Add group "Manage Multiple Units of Measure" to admin
+        admin = cls.env.ref("base.user_admin")
+        admin.group_ids |= cls.env.ref("uom.group_uom")
+        # Force VAT to avoid error in the module that makes it required.
+        admin.partner_id.vat = "TEST12345678"
```
carlos-lopez-tecnativa: I think it would be better to set a valid VAT number to prevent errors similar to https://github.com/OCA/e-commerce/pull/1285

carlos-lopez-tecnativa (changes requested):
LGTM. I left a comment and hope it can be implemented.

Before V19, the `Add to Cart` button in the `shop` was not enabled by default and could be enabled through the `Customize options`.

Image

But in V19, I noticed that it is enabled by default, and there is no option to hide it.

Image

With this option enabled, when a product has variants, a modal is displayed to select the variant. **In this context, no secondary units are displayed. It would be great if we could select the secondary units here as well.**

Imagine that in my product, I set `Allow to Sell in Unit of Measure` to `False` and use only secondary units. In this case, the first secondary unit is currently selected, but it does not check whether it is published or not. If the first secondary unit is `unpublished`, I think this is incorrect behavior.

For me, the expected behavior would be to select the first published secondary unit.

Image

Image

Image

pilarvargas-tecnativa:
Perhaps the simplest solution to what @carlos-lopez-tecnativa mentions, to address something that no one has requested over the course of several versions, is to bring back the option to hide the shopping cart button.

CarlosRoca13:
> Perhaps the simplest solution to what @carlos-lopez-tecnativa mentions, to address something that no one has requested over the course of several versions, is to bring back the option to hide the shopping cart button.

But I think it should be an extra module

pedrobaeza:
Isn't possible to add the secondary units to that screen that is also included in the main `website_sale` module? Or is it too complicated?

CarlosRoca13:
I'm working on adding it

pedrobaeza:
Please add in the main PR comment on which PR it depends as reference.

carlos-lopez-tecnativa (commented):
LGTM.

Just a side note, to avoid going out of scope for the migration, but I think this issue could also be handled in this module because the feature is added by website_sale.

When I finish a sale order with a line that has a secondary unit, then go to My Account and reorder the same order, the quantities are changed. I suspect that, in this flow, the backend uses the default secondary unit or the first secondary unit. However, I would expect it to use the same secondary unit as the original order.

Could you add this to the ROADMAP? Or do you think this behavior could be fixed in this PR?

Image

Image

In the backend:
Original order
Image

New order
Image

CarlosRoca13:
@carlos-lopez-tecnativa solved

carlos-lopez-tecnativa (approved):
LGTM, thanks for the improvements.

pilarvargas-tecnativa (approved):

PR #1299, opened 2026-08-20
[19.0][MIG] website_sale_product_multi_website: Migration to version 19.0

cristina-hidalgo-tecnativa:
* Adapt `website.website_domain()` override to the new signature (no `website_id` arg, returns `odoo.fields.Domain`).
* Adapt `can_access_from_current_website()` to use `request.env['website'].get_current_website()` instead of the removed context-based lookup.
* Update `website.snippet.filter` overrides to the new `**kwargs`/`**options` signatures.

@Tecnativa TT64165
@pedrobaeza  @pilarvargas-tecnativa


pedrobaeza:
Please check CI

pilarvargas-tecnativa:
Please check CI and reduce administrative commits

cristina-hidalgo-tecnativa:
> Please check CI and reduce administrative commits

Done, could you please review it again?

website_sale_product_multi_website/models/website.py
```diff
@@ -1,21 +1,35 @@
...
+
+            # Unlike website_id, website_ids has no SQL column of its own, so
+            # a plain Domain evaluated in-memory (e.g. filtered_domain(), used
+            # by core accessory/alternative product logic) would fetch it
+            # directly and raise an AccessError for products a public user
+            # can't otherwise read. Domain.custom() without a predicate makes
+            # that in-memory evaluation run through a real search() instead,
+            # which applies ir.rule the same safe way regular searches do.
+            return Domain.custom(to_sql=to_sql)
+        return super().website_domain()
```
pilarvargas-tecnativa, line 32: Could you add a test reproducing the AccessError case mentioned in the comment? It would help to verify that this workaround is actually required. If possible, could you also check whether the same case can be handled with a regular Domain instead of the custom SQL domain, keeping the implementation simpler?

cristina-hidalgo-tecnativa: Test added!.

Regarding a regular Domain: it can't be used here. Odoo's fetch() (`models.py`) decides between a real `search()` (which applies `ir.rule` silently) or a strict `check_access('read')` based on `field.column_type` — and many2many fields like `website_ids` never have a `column_type`, regardless of the operator used. Any plain Domain leaf on `website_ids` evaluated in-memory (`filtered_domain`, used by `_get_website_accessory_product`) would hit the strict path and raise.

The `Domain.custom()` workaround is necessary to route that in-memory evaluation through a real `search()` instead, same as core does in [analytic_mixin.py](https://github.com/odoo/odoo/blob/701c8cccffb4a262d7421681de368cc2af44f5ad/addons/analytic/models/analytic_mixin.py#L112-L118), [mrp_production.py](https://github.com/odoo/odoo/blob/701c8cccffb4a262d7421681de368cc2af44f5ad/addons/mrp/models/mrp_production.py#L908), [account_tax.py](https://github.com/odoo/odoo/blob/701c8cccffb4a262d7421681de368cc2af44f5ad/addons/account/models/account_tax.py#L270) and [purchase_order.py](https://github.com/odoo/odoo/blob/701c8cccffb4a262d7421681de368cc2af44f5ad/addons/purchase/models/purchase_order.py#L372-L381).

pilarvargas-tecnativa: thanks

pilarvargas-tecnativa (approved):
Code review. LGTM!

pedrobaeza (commented):
Code and functional review

/ocabot merge nobump

PR #1300, opened 2026-08-21
[19.0][FIX] website_sale_product_minimal_price: Fix price scale rendering and currency

cristina-hidalgo-tecnativa:
Fixes two v19 regressions in the price scale feature:

- **Rendering**: the `VariantMixin._onChangeCombination` monkey-patch is orphaned since core now copies it into `WebsiteSale.prototype` via `Object.assign` at import time. Switched to `patch(WebsiteSale.prototype, {...})`, and fixed the form selector/insertion point (the `<form>` has no `action` anymore and Terms/shipping/accordion now live inside it too).
- **Currency**: the price scale used `product.currency_id` instead of the pricelist's currency, causing a mismatch between the amount and its symbol.

@Tecnativa TT64164

### DIFF
--- a/website_sale_product_minimal_price/models/product_template.py
+++ b/website_sale_product_minimal_price/models/product_template.py
@@ -177,7 +177,7 @@ class ProductTemplate(models.Model):
                     {
                         "min_qty": min_qty,
                         "price": new_price,
-                        "currency_id": product.currency_id.id,
+                        "currency_id": combination_info["currency"].id,
                     }
                 )
                 last_price = new_price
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please remove the patch that is in #1273 (or use temporary reference if needed, but not to hard-copy it), and for the rest, please add an explanation of the what and why in the commit message itself.

cristina-hidalgo-tecnativa:
> Please remove the patch that is in #1273 (or use temporary reference if needed, but not to hard-copy it), and for the rest, please add an explanation of the what and why in the commit message itself.

Done, could you review it again please?

carlos-lopez-tecnativa (approved):
LGTM, thanks for the fix. How could this module be merged if it wasn't working... 😄

Anyway, I left a comment as a future improvement. In V19, you have a button to add the product to the cart from the product list, and this button cannot be disabled. This button opens a modal, so it could be interesting to add the price scale to this modal as well. What do you think?

Image
Image

Andrii9090-tecnativa (approved):
LGTM. Test on runboat 👍

PR #1301, opened 2026-08-21
[19.0][MIG] website_sale_product_matrix: Migration to version 19.0

cristina-hidalgo-tecnativa:
- The `add-to-cart` button moved out of `website_sale.product` into a new `website_sale.cta_wrapper` template, so the "Configure options" button/hidden `add-to-cart` xpaths were moved there too.
- The variant selector is no longer wrapped in a t`-placeholder="select"`, so the xpath hiding it was adjusted accordingly.
- `website_sale.add_to_cart_redirect` no longer exists in 19.0 (removed along with it), so the template and JS code relying on it (`product-add-mode` html attribute / `stayOnPageOption`) were dropped as dead code.
- `website.sale_get_order()` was replaced by `request.cart` / `website._create_cart()` in the controller and in the matrix modal template.
- The legacy `publicWidget.registry.WebsiteSale` widget was removed in 19.0, so the JS was rewritten as its own small widget instead of extending it.

@Tecnativa TT64162


website_sale_product_matrix/templates/product_template.xml
```diff
@@ -0,0 +1,206 @@
...
+                                                class="badge badge-pill badge-secondary"
+                                            >
+                                                <span
+                                                    class="variant_price_extra"
+                                                    style="white-space: nowrap;"
+                                                >
+                                                    + <t
+                                                    t-esc="column_header['price']"
+                                                    t-options="{'widget': 'monetary', 'display_currency': website.currency_id}"
+                                                />
```
pilarvargas-tecnativa: This indentation looks strange, did precommit do this?

cristina-hidalgo-tecnativa: Yes, this is `prettier` (with `plugin-xml`, run via our pre-commit hook), not a manual edit. It breaks the line right after the `+` (it treats it as a text node attached to the `<t>` tag) and misaligns the `<t>` attributes relative to the tag itself. 

I reproduced it standalone with a clean-indented version of the same snippet and pre-commit's prettier hook reformats it exactly the same way, so it's just a quirk of that plugin with mixed text+tag content.

pilarvargas-tecnativa: Thanks, I've noticed some similar changes, which is why I asked, but I can see now that pre-commit is messing with the indentation and making it look a little weird.

website_sale_product_matrix/templates/product_template.xml
```diff
@@ -0,0 +1,206 @@
...
+            <a
+                t-if="product.product_add_mode == 'matrix'"
+                href="#modalMatrix"
+                data-bs-toggle="modal"
+                data-bs-target="#modalMatrix"
+                aria-expanded="false"
+                aria-controls="modalMatrix"
+                role="button"
+                class="btn btn-primary my-1 me-1 px-5 py-2 flex-grow-1 text-uppercase"
+            ><i class="fa fa-cart-plus mr-2" /> Configure options</a>
```
pilarvargas-tecnativa:
```suggestion
            ><i class="fa fa-cart-plus me-2" /> Configure options</a>
```

website_sale_product_matrix/templates/product_template.xml
```diff
@@ -155,33 +140,31 @@
...
                                     t-att-value="request.csrf_token()"
                                 />
+                                <t t-call="website_sale_product_matrix.product_matrix">
                                     <t
-                                    t-call="website_sale_product_matrix.product_matrix"
-                                >
-                                        <t
                                         t-set="website_sale_order"
-                                        t-value="website.sale_get_order()"
+                                        t-value="request.cart or website._create_cart()"
```
pilarvargas-tecnativa: Is this really necessary here? Calling _create_cart() while rendering the product page creates a sale.order just by visiting it.

As far as I can see, the cart is only actually needed when adding the matrix quantities, and cart_update_from_matrix already creates it when needed.

website_sale_product_matrix/templates/product_template.xml
cristina-hidalgo-tecnativa: Modified.

website_sale_product_matrix/templates/product_template.xml
cristina-hidalgo-tecnativa: You're right, fixed — now it's just `request.cart`, matching the original `sale_get_order()` (no `force_create`). `_get_matrix()` works fine on an empty order.

cristina-hidalgo-tecnativa:
@pedrobaeza could you please review it?

website_sale_product_matrix/templates/product_template.xml
pedrobaeza: This indentation is correct. The incorrect one is for `+ <t`, that is indented one more level.

pedrobaeza (commented):
Testing it on runboat, one problem I see is if you wishlist the template, on the wishlist view, you don't have the matrix on it, but the regular configurator one, but I suppose this problem also exists in the previous version, and it will require a `website_sale_wishlist_product_matrix` glue module. Just mention it in the ROADMAP.

website_sale_product_matrix/templates/product_template.xml
cristina-hidalgo-tecnativa: I could wrap `+` in its own `<span>+</span>` (instead of bare text next to `<t>`) — verified it formats cleanly and stays stable across repeated pre-commit runs. No visual change.

pedrobaeza: I think you can just deindent that line.

cristina-hidalgo-tecnativa: Done, should be good now.

cristina-hidalgo-tecnativa:
> Testing it on runboat, one problem I see is if you wishlist the template, on the wishlist view, you don't have the matrix on it, but the regular configurator one, but I suppose this problem also exists in the previous version, and it will require a `website_sale_wishlist_product_matrix` glue module. Just mention it in the ROADMAP.

Added it to the ROADMAP.

pedrobaeza (approved):

CarlosRoca13 (approved):

PR #1302, opened 2026-08-22
[18.0][IMP] website_sale_product_minimal_price: performance improvements

jans23:
`_get_cheapest_info` iterated over every variant of every template, which forced a per-record compute of the non-stored `price_extra` field and a contextual price computation for each variant. This is very slow for products with many variants.

* When there are no per-variant pricelist rules, the price is monotonic in `price_extra`, so only the base, min-extra and max-extra variants can be the cheapest. Price only those instead of all variants.
* Read `price_extra` for all variants via a single grouped SQL query (`_get_variants_price_extra_map`) instead of the per-record ORM compute, and prefetch it once per page in `_get_sales_prices`.

No request-scoped caching is added because profiling showed `_get_cheapest_info` is called at most once per template per request, so a cache never yields a hit and only adds complexity.

### DIFF
--- a/website_sale_product_minimal_price/models/product_template.py
+++ b/website_sale_product_minimal_price/models/product_template.py
@@ -9,6 +9,37 @@ from odoo.osv import expression
 class ProductTemplate(models.Model):
     _inherit = "product.template"
 
+    def _get_variants_price_extra_map(self, variants):
+        """Return ``{variant_id: price_extra}`` computed in a single SQL query.
+
+        Avoids the slow per-record ORM compute of the non-stored ``price_extra``
+        field for products with many variants.
+        """
+        result = {}
+        if not variants:
+            return result
+        # Resolve m2m table/columns from the field to avoid hardcoded names.
+        field = self.env["product.product"]._fields[
+            "product_template_attribute_value_ids"
+        ]
+        rel, col_variant, col_value = field.relation, field.column1, field.column2
+        self.env.cr.execute(
+            f"""
+            SELECT rel.{col_variant} AS variant_id,
+                   COALESCE(SUM(ptav.price_extra), 0.0) AS price_extra
+            FROM {rel} rel
+            JOIN product_template_attribute_value ptav
+                 ON ptav.id = rel.{col_value}
+            WHERE rel.{col_variant} IN %s
+            GROUP BY rel.{col_variant}
+            """,
+            (tuple(variants.ids),),
+        )
+        queried = dict(self.env.cr.fetchall())
+        for variant in variants:
+            result[variant.id] = queried.get(variant.id, 0.0)
+        return result
+
     def _get_product_subpricelists(self, pricelist):
         base_domain = pricelist._get_applicable_rules_domain(
             self, fields.Datetime.now()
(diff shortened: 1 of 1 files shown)


website_sale_product_minimal_price/models/product_template.py
```diff
@@ -48,29 +79,41 @@ def _get_pricelist_variant_items(self, pricelist):
...
+            # the base, min-extra and max-extra variants can be cheapest/differ.
+            # price_extra is read in bulk (see _get_variants_price_extra_map).
+            extra_map = price_extra_map
+            if extra_map is None:
+                extra_map = self._get_variants_price_extra_map(self.product_variant_ids)
+            variants_with_extra = [
+                variant
+                for variant in self.product_variant_ids
+                if extra_map.get(variant.id)
+            ]
```
fkantelberg: You are converting the recordset to a list of records which can be quite annoying. The following does the same but also results in a recordset.

```python
variants_with_extra = self.product_variant_ids.filtered(lambda variant: extra_map.get(variant.id))
variants_without_extra_price = self.product_variant_ids - variants_with_extra

...
if variants_with_extra:
    variants_with_extra.sorted(lambda v: extra_map.get(v.id) or 0.0)
```

website_sale_product_minimal_price/models/product_template.py
```diff
@@ -48,29 +79,41 @@ def _get_pricelist_variant_items(self, pricelist):
...
             variants = variant_items.mapped("product_id")
             products = variants + (self.product_variant_ids - variants)[:1]
+            products |= self.product_variant_ids.filtered("price_extra")
         else:
+            # No per-variant rules: price is monotonic in price_extra, so only
+            # the base, min-extra and max-extra variants can be cheapest/differ.
+            # price_extra is read in bulk (see _get_variants_price_extra_map).
+            extra_map = price_extra_map
+            if extra_map is None:
+                extra_map = self._get_variants_price_extra_map(self.product_variant_ids)
```
fkantelberg: You don't need to rename the variable and can stay with `price_extra_map` because you are only assigning the dict and not adjusting it in the following code.

website_sale_product_minimal_price/models/product_template.py
```diff
@@ -9,6 +9,37 @@
...
+            JOIN product_template_attribute_value ptav
+                 ON ptav.id = rel.{col_value}
+            WHERE rel.{col_variant} IN %s
+            GROUP BY rel.{col_variant}
+            """,
+            (tuple(variants.ids),),
+        )
+        queried = dict(self.env.cr.fetchall())
+        for variant in variants:
+            result[variant.id] = queried.get(variant.id, 0.0)
```
fkantelberg: If you would use `result[variant]` some of the below code might get easier too.

E.g.
```python
self.product_variant_ids.filtered(extra_map.get)
# instead of
self.product_variant_ids.filtered(lambda variant: extra_map.get(variant.id))
```

fkantelberg (changes requested):

pedrobaeza:
You should check #1310

website_sale_product_minimal_price/models/product_template.py
jans23: changed as requested.

website_sale_product_minimal_price/models/product_template.py
jans23: changed as requested.

website_sale_product_minimal_price/models/product_template.py
jans23: changed as requested.

jans23:
#1310 (ported to 18.0) reduces loading time from ca. 37 seconds to ca. 24 seconds while this PR reduces it to less than 1 second.

fkantelberg (approved):

PR #1304, opened 2026-08-25
[19.0][MIG] website_sale_product_attribute_filter_order: Migration to version 19.0

cristina-hidalgo-tecnativa:
* In 19.0, `website_sale.products_attributes`'s combined `radio`/`pills`/`multi` branch was split into per-type sub-templates, and core now always sorts selected values first for `radio`/`multi`. So this module now only patches `website_sale.filter_pills_attributes`, the one type core still leaves unordered.
* `views/snippets/snippets.xml` inherited `website_sale.snippet_options`, removed in 19.0 (`we-checkbox` panel replaced by OWL builder options). Replaced by `static/src/website_builder/products_list_page_option.xml`, re-adding the "Filters Marked First" checkbox in the same spot.
* New asset registered under `website.website_builder_assets`.
* Tests updated to reference the renamed `filter_pills_attributes` view and use a `pills` attribute, the only display type still covered by the module.

@Tecnativa TT64161


pedrobaeza:
/ocabot migration website_sale_product_attribute_filter_order

adasatorres-tecnativa (approved):
Code review and tested on runboat.

pedrobaeza:
/ocabot migration website_sale_product_attribute_filter_order

website_sale_product_attribute_filter_order/templates/website_sale.xml
```diff
@@ -1,16 +1,13 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
     <template
-        id="products_attributes"
-        inherit_id="website_sale.products_attributes"
+        id="filter_pills_attributes"
```
pilarvargas-tecnativa: If we are going to change the XML ID, I think we should add a migration script to rename the existing ir.model.data entry instead of creating a new view.

Otherwise, Odoo will consider products_attributes removed and filter_pills_attributes a new record, so any existing configuration attached to the previous view, including website-specific copies created when enabling the option for a particular website, may not be preserved.

pilarvargas-tecnativa:
Please reduce administrative commits

website_sale_product_attribute_filter_order/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,5 @@
+Move active checkbox options to the first place of the list.
```
pilarvargas-tecnativa:
```suggestion
Move active attribute filter options to the first place of the list for the pills display type.
```

website_sale_product_attribute_filter_order/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,5 @@
...
+
+Since Odoo 19.0, the "radio" and "multi" attribute filter display types
+already do this by default in Odoo core, so this module now only affects
+the "pills" display type.
```
pilarvargas-tecnativa: I don't think we need to mention the Odoo 19 core behavior in the module description. This is useful context for the migration PR, but the README should describe what the module provides in the current version.

Since in 19.0 the module only applies this behavior to the pills display type, I would just adapt the description accordingly.

cristina-hidalgo-tecnativa: Modified.

website_sale_product_attribute_filter_order/readme/DESCRIPTION.md
cristina-hidalgo-tecnativa: Modified.

website_sale_product_attribute_filter_order/templates/website_sale.xml
cristina-hidalgo-tecnativa: We can keep `id="products_attributes"` unchanged and only update `inherit_id` to `website_sale.filter_pills_attributes` (plus the xpath), instead of adding a migration script. Since the `ir.model.data` entry stays the same, the existing record — and any website-specific copy created when the option was enabled — is preserved and just gets its `inherit_id/arch` updated on `-u`.

cristina-hidalgo-tecnativa:
> Please reduce administrative commits

Done.

pilarvargas-tecnativa (approved):

PR #1307, opened 2026-09-01
[19.0][MIG] website_sale_product_matrix_hide_price: Migration to version 19.0

cristina-hidalgo-tecnativa:
* The "Configure options" button (`href="#modalMatrix"`) moved out of `website_sale_product_matrix.produc`t into the new `website_sale_product_matrix.cta_wrapper` template (see #1301), so the xpath hiding it when prices are hidden was moved there too. 
* `slug()` is no longer importable from `odoo.addons.http_routing.models.ir_http`; the test now uses `self.env["ir.http"]._slug()` instead.

@Tecnativa TT64163


pedrobaeza (approved):
Code and functional review.

The button to configure the matrix is hidden.

Andrii9090-tecnativa (approved):
Code&functional review

pilarvargas-tecnativa (approved):

PR #1309, opened 2026-09-01
[19.0][MIG] website_sale_stock_product_matrix: Migration to version 19.0

cristina-hidalgo-tecnativa:
* Updated `_get_additionnal_combination_info()` override to match the new `website_sale`/`website_sale_stock` signature, which now takes an extra `uom` argument.
* Replaced the `type == 'product'` checks (model and template) with `is_storable`, since `product.template.type` no longer has a `product` value in 19.0 (storable tracking moved to the `is_storable` boolean).

@Tecnativa TT64168


website_sale_stock_product_matrix/templates/product_template.xml
```diff
@@ -43,9 +43,10 @@
                     t-elif="product.show_availability"
                     t-attf-class="availability_message_#{product.id}"
                 >
-                    <small class="text-success"><t t-esc="int(free_qty)" /> <t
-                            t-esc="product.uom_name"
-                        /></small>
+                    <small class="text-success">
+                        <t t-esc="int(free_qty)" />
```
pedrobaeza: Why converting to `int`? Quantities can be float.

pedrobaeza (commented):
Testing on runboat, I get an error:

Image

website_sale_stock_product_matrix/models/product_template.py
```diff
@@ -0,0 +1,32 @@
...
+        if (
+            self.product_add_mode == "matrix"
+            and self.is_storable
+            and product_or_template.is_product_variant
+            and not self.env.context.get("product_matrix_cell")
+        ):
+            res["free_qty"] = sum(
+                website._get_product_available_qty(variant)
+                for variant in self.sudo().product_variant_ids
+            )
```
pilarvargas-tecnativa: In 19.0, core only computes free_qty when website_sale_stock_get_quantity is set in the context, and it converts the quantity to the requested uom.

Could you adapt this override to preserve both behaviors when computing the aggregated free_qty?

https://github.com/odoo/odoo/blob/c1ec73be8e16b57c8cf7e4d0c3a915dd9fd6c78b/addons/website_sale_stock/models/product_template.py#L44

cristina-hidalgo-tecnativa: Adapted.

cristina-hidalgo-tecnativa:
ping @pilarvargas-tecnativa

website_sale_stock_product_matrix/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+    _inherit = "product.template"
+
+    @api.model
+    def _format_matrix_free_qty(self, qty):
+        """Format a matrix cell's free quantity for display.
+
+        Mirrors `SaleOrder._verify_updated_quantity()`'s own `format_qty()` in
+        `website_sale_stock`: avoid a spurious trailing ``.0`` for whole
+        numbers, while still showing genuinely fractional quantities as-is.
+        """
```
pilarvargas-tecnativa: Could we simplify this docstring? The reference to SaleOrder._verify_updated_quantity() seems too implementation-specific and may become outdated. A short description of what this helper does should be enough.

```suggestion
        """Format the matrix free quantity for display."""
```

website_sale_stock_product_matrix/models/product_template.py
```diff
@@ -0,0 +1,51 @@
...
+        res = super()._get_additionnal_combination_info(
+            product_or_template, quantity, uom, date, website
+        )
+        # For products sold through the variant matrix the variant selector is
+        # hidden, so the page-level availability message stays frozen on the first
+        # variant and wrongly shows "out of stock" when only the first size/color
+        # is depleted. The message must reflect the whole template instead: if ANY
+        # variant has stock the product isn't out of stock. The matrix grid cells
+        # still need the per-variant quantity, so those calls are flagged with the
+        # `product_matrix_cell` context and left untouched.
```
pilarvargas-tecnativa: Could we simplify this comment? The full explanation of the matrix behavior seems a bit too verbose for the code. Something short explaining why template-wide stock is used here should be enough.
```suggestion
        # Use template-wide stock for the matrix availability message.
```

website_sale_stock_product_matrix/templates/product_template.xml
```diff
@@ -0,0 +1,68 @@
...
+                name="t-if"
+            >cell.get('is_possible_combination') and not disallow_out_of_stock</attribute>
+        </xpath>
+        <!-- The "Not available" span used to be a plain t-else paired with the
+        css_quantity div's t-if. Since we insert content between them below,
+        that pairing would silently break (t-else would bind to whatever
+        element ends up immediately before it instead), so it needs its own
+        explicit, independent condition: only genuinely impossible
+        combinations are "not available" - being blocked by stock is a
+        different case, handled by the availability message. -->
```
pilarvargas-tecnativa: Could we simplify this comment? The important part is just explaining why the condition needs to be explicit.

```suggestion
    <!-- Keep this condition explicit as content is inserted after css_quantity. -->
```

pilarvargas-tecnativa (commented):
Just a few minor comments.

pilarvargas-tecnativa:
@carlos-lopez-tecnativa please, can you take a look?

pilarvargas-tecnativa (approved):

carlos-lopez-tecnativa (approved):

PR #1310, opened 2026-09-02
[17.0][FIX] website_sale_product_minimal_price: improve shop performance

pilarvargas-tecnativa:
Avoid computing variant prices and pricelist items one template at a time.

Batch the variant-specific pricelist item lookup and product price computation, and avoid calling `_get_additionnal_combination_info` for every product by computing the required sales prices in batch.

This significantly reduces the overhead introduced by the module on `/shop`.

@Tecnativa TT64353

@carlos-lopez-tecnativa @carlosdauden please review

### DIFF
--- a/website_sale_product_minimal_price/models/product_template.py
+++ b/website_sale_product_minimal_price/models/product_template.py
@@ -28,8 +28,11 @@ class ProductTemplate(models.Model):
         return self.env["product.pricelist"].browse(pricelist_ids)
 
     def _get_variants_from_pricelist(self, pricelist):
-        return pricelist.mapped("item_ids").filtered(
-            lambda i: i.product_id in self.product_variant_ids
+        return self.env["product.pricelist.item"].search(
+            [
+                ("pricelist_id", "in", pricelist.ids),
+                ("product_id.product_tmpl_id", "in", self.ids),
+            ]
         )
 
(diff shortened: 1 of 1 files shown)


website_sale_product_minimal_price/models/product_template.py
```diff
@@ -181,22 +212,50 @@ def _get_combination_info(
...
+                        target_currency=currency,
+                    )
+                )
+        for template in published_templates:
+            product, _add_qty, has_distinct_price = cheapest_info[template.id]
+            product_taxes = product.sudo().taxes_id._filter_taxes_by_company(
+                self.env.company
             )
-            product_price_info = template._get_additionnal_combination_info(
+            taxes = fiscal_position.map_tax(product_taxes)
```
pedrobaeza, line 245: Why adding taxes?

pilarvargas-tecnativa: By removing _get_additionnal_combination_info to avoid computing it product by product, we need to explicitly preserve the tax handling it performed internally to avoid changing the displayed price.

https://github.com/odoo/odoo/blob/2b5d75fbe7d7cb1c57ca10d8348c1bd2d319c4fa/addons/website_sale/models/product_template.py#L505-L521

PR #1317, opened 2026-09-09
[17.0][FIX] website_sale_hide_empty_category

pilarvargas-tecnativa:
- I've added a commit to make the test more resilient by activating the views in the corresponding website context
-  Backport of the performance improvement from #1226, adapted taking into account @ivantodorovich  review comments 

@Tecnativa

@pedrobaeza please review

### DIFF
--- a/website_sale_hide_empty_category/models/product_public_category.py
+++ b/website_sale_hide_empty_category/models/product_public_category.py
@@ -16,12 +16,24 @@ class ProductPublicCategory(models.Model):
     @api.depends("product_tmpl_ids", "child_id.has_product_recursive")
     @api.depends_context("website_id")
     def _compute_has_product_recursive(self):
-        for category in self:
-            website = self.env["website"].get_current_website()
-            website_domain = website.sale_product_domain()
-            has_products = bool(
-                category.product_tmpl_ids.filtered_domain(website_domain)
+        website = self.env["website"].get_current_website()
+        product_domain = website.sale_product_domain()
+        categories_with_products = self.search(
+            [
+                ("website_id", "in", [False, website.id]),
+                ("product_tmpl_ids", "any", product_domain),
+            ]
+        )
+        used_category_ids = set()
+        for category in categories_with_products:
+            used_category_ids.update(
+                int(category_id)
+                for category_id in category.parent_path.split("/")
+                if category_id
             )
-            category.has_product_recursive = has_products or any(
-                child.has_product_recursive for child in category.child_id
-            )
+        categories = self.search(website.website_domain()) | self
+        self.env.cache.update_raw(
+            categories,
+            self._fields["has_product_recursive"],
+            [category.id in used_category_ids for category in categories],
+        )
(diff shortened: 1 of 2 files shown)


ivantodorovich:
Hello 👋🏻 I don't see my recommendations applied here 🤔 
https://github.com/OCA/e-commerce/pull/1226#pullrequestreview-4429432934

pilarvargas-tecnativa:
> Hello 👋🏻 I don't see my recommendations applied here 🤔 [#1226 (review)](https://github.com/OCA/e-commerce/pull/1226#pullrequestreview-4429432934)

Hi @ivantodorovich! The first two recommendations are applied here: _read_group was replaced with a direct category search using any, and the website filtering was moved to the search domain.

I kept the parent_path propagation because the nested any suggestion only considers direct children and, as you mentioned, it was uncertain whether the nested domain would work.

Regarding update_raw(), I kept it intentionally because the compute may be triggered for a single category, while the cache is populated for all relevant categories at once, avoiding repeating the same search for each category. has_product_recursive is non-stored and context-dependent, so in this case the cache is being used specifically to avoid those repeated computations.

Is there any other recommendation from your review that you expected to be applied?

ivantodorovich:
>  The first two recommendations are applied here: _read_group was replaced with a direct category search using any, and the website filtering was moved to the search domain.

are you sure it's pushed? I don't see that in the code

pilarvargas-tecnativa:
> are you sure it's pushed? I don't see that in the code

You're right, the changes weren't properly included in the pushed commit. Thanks for noticing!

ivantodorovich (approved):
thanks!

website_sale_hide_empty_category/tests/test_ui.py
```diff
@@ -1,21 +1,16 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from odoo.tests import tagged
 from odoo.tests.common import HttpCase
 
+from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
 
+
+@tagged("post_install", "-at_install")
```
pedrobaeza, line 9: Why do you need post_install?

website_sale_hide_empty_category/models/product_public_category.py
```diff
@@ -16,12 +16,24 @@ class ProductPublicCategory(models.Model):
...
-                category.product_tmpl_ids.filtered_domain(website_domain)
-            )
-            category.has_product_recursive = has_products or any(
-                child.has_product_recursive for child in category.child_id
+        website = self.env["website"].get_current_website()
+        website_domain = website.sale_product_domain()
+        categories_with_products = self.search(
+            [
+                ("website_id", "in", [False, website.id]),
+                ("product_tmpl_ids", "any", website_domain),
```
pedrobaeza: Use a proper variable name:

```suggestion
        product_domain = website.sale_product_domain()
        categories_with_products = self.search(
            [
                ("website_id", "in", [False, website.id]),
                ("product_tmpl_ids", "any", product_domain),
```

website_sale_hide_empty_category/tests/test_ui.py
pilarvargas-tecnativa: To avoid the warning

```
2026-09-11 06:21:28,292 1 WARNING devel odoo.addons.website_sale_hide_empty_category.tests.test_ui.UICase: HttpCase test should be in post_install only

pedrobaeza (approved):
/ocabot merge patch

PR #1321, opened 2026-09-11
[18.0][IMP] website_sale_product_attribute_value_filter_existing: optimize attribute values lookup

pilarvargas-tecnativa:
@Tecnativa


pedrobaeza:
Can you please give reasons for this to be an optimization? Or at least, show traces of before/after with the improvement?

pilarvargas-tecnativa:
> Can you please give reasons for this to be an optimization? Or at least, show traces of before/after with the improvement?

We’ll have to wait until it’s rolled out to be able to make a proper comparison. Once I have it, I’ll complete the PR description.

pilarvargas-tecnativa:
It doesn’t bring about any real improvement

PR #1326, opened 2026-09-16
[19.0] [FIX] website_sale_product_brand: brand search bar layout

ivantodorovich:
Replace the fixed-width columns and the inline `margin-left:60%` hack with a plain Bootstrap input-group, so the search bar spans the page and stays usable on small screens.

Before:
Screenshot 2026-09-16 at 12 47 54 PM
image

After:
image
image


imlopes (approved):
LGTM

Ricardoalso (approved):
Thanks !

